PR #5, opened 2020-07-11
Fix #3 by removing use of __index metamethod and deep-extending config

### LINKED ISSUE #3: Configuration table is not handled correctly by init
(Sorry for the vague title.)

It seems that the config table that is not correctly merged with the default values. I am following the example setup in the readme, except that I want to pass a table to `packer.init()`. However, this prevents the package handling functionality (install, update, etc.) from working; nothing happens. Specifically, if I pass
```lua
{
  display = {
    open_cmd = 'new [packer]',
    working_sym = '⟳', -- The symbol for a plugin being installed/updated
    error_sym = '✗', -- The symbol for a plugin with an error in installation/updating
    done_sym = '✓', -- The symbol for a plugin which has completed installation/updating
    removed_sym = '-', -- The symbol for an unused plugin which was removed
    moved_sym = '→', -- The symbol for a plugin which was moved (e.g. from opt to start)
    header_sym = '━', -- The symbol for the header line in packer's display
  }
}
```
the new buffer is opened, but nothing happens. This is even the case if I paste the full table with default values from the readme. If I omit one of the symbols from the table, packer complains about
```
E5108: Error executing lua ...ite/pack/packer/start/packer.nvim/lua/packer/display.lua:299: attempt to concate
nate local 'working_sym' (a nil value)
```

wbthomason:
@clason: This fixes your example in #3 for me - can you confirm that it also fixes things on your end? Thanks!

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -2,52 +2,51 @@
 -- TODO: Fetch/manage LuaRocks dependencies
 -- TODO: Performance analysis/tuning
 -- TODO: Merge start plugins?
-
 -- WIP:
 -- TODO: Allow separate packages
-
-local a            = require('packer/async')
-local clean        = require('packer/clean')
-local compile      = require('packer/compile')
-local display      = require('packer/display')
-local handlers     = require('packer/handlers')
-local install      = require('packer/install')
-local log          = require('packer/log')
+local a = require('packer/async')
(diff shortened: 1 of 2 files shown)


clason:
Yes, that works, thank you very much!

wbthomason:
Great, thank you!

PR #10, opened 2020-07-12
Add Startup Example File

tjdevries:
Here's an example of how I was thinking you could "opt-in" to default packer style.

Example: https://github.com/tjdevries/config_manager/blob/master/xdg_config/nvim/lua/plugins.lua

Not sold on any of the names in particular, but I think it's better to have commands & function & boilerplate set up.


wbthomason:
Thanks! This would make it easier to get going with `packer`.

I am a little bit surprised that this works with `packer` managed as an `opt` plugin - specifically, I would expect https://github.com/tjdevries/config_manager/blob/63efa4124b5a39c33f6abf77bdc9549e639f360b/xdg_config/nvim/lua/plugins.lua#L3 to fail unless `packer` has been `packadd`'ed (or is a `start` plugin). Do you have any insight into why this doesn't fail?

Similarly, I'm not 100% sure how this is to be used - it seems like using the `startup` module requires that `packer` be loaded, but the commands defined in `startup` call `packadd packer.nvim`, and your example config (which looks great, btw! :D), implying that `packer` is **not** loaded. How have you been using this? Sorry if I'm just being dense and missing something here.

Also, could I request that you add a simplified example of usage to the README? Without that or documentation comments, it's a little confusing to see what e.g. the `f` argument to `startup` should do.

tjdevries:
Is there any reason you don't want to have `packer` in `start`? It doesn't have anything that gets sourced right away, so I think it's better to add packer as a start always (just to prevent weird stuff from happening).

You're right though, I just call `packadd` before I load all this stuff.

I can add documentation and clean up the general idea as well, just wanted to see if this was in line with what you're thinking.

BTW, started reading some luarocks stuff yesterday :) so after we get this sorted out, maybe we can start brainstorming on thyat.

wbthomason:
> Is there any reason you don't want to have packer in start?

I'd be happy to be disabused of this notion, but I've found (anecdotally) that loading Lua plugins from `start` imposes a nontrivial startup time cost. I should test this more thoroughly to get hard numbers to back up my subjective belief, though. That's the main motivation for pushing so hard to put `packer` in `opt` rather than `start`. If I'm wrong about this, then the whole "compile lazy-loaders" paradigm matters less (though I'll still keep it around as it's still going to be a bit faster than dynamic lazy-load logic at runtime, and it's already written).

> I can add documentation and clean up the general idea as well, just wanted to see if this was in line with what you're thinking.

Modulo figuring out if making `packer` a `start` plugin adds startup time cost, this seems great to me, thank you!

> BTW, started reading some luarocks stuff yesterday :) so after we get this sorted out, maybe we can start brainstorming on thyat.

Sounds good! I'm still busy with the conference I mentioned (https://roboticsconference.org/program/detailed/) through much of this week, but will have time starting late this week/early next week.

tjdevries:
Hmmm, when I just tried it and checking w/ startuptime I didn't see any difference between having it in and not having it in `opt`.

Moved between `opt`, `opt` + `packadd`, `opt` + `packadd` + `lua require('plugins')`, `start` and I got ~ the same time.

wbthomason:
Hmm, interesting. On my system, I see a cost difference with including any Lua from `~/.config/nvim/lua` in a startup-sourced file. For example, including the lines 

```vim
lua statusline = require('statusline')
lua vim.o.statusline = '%!v:lua.statusline()'
```
to `~/.config/nvim/plugin/statusline.vim`, where `~/.config/nvim/lua/statusline.lua` exists, causes a startup cost of ~11ms vs 0.33ms without those lines.

wbthomason:
(to clarify the above example a bit more, just using the first of those two lines costs ~10ms by itself)

tjdevries:
floating_manager

btw, added a floating window in my setup so I can be cool ;)

wbthomason:
Thanks, this looks great!

wbthomason:
Incidentally, I figured out the cause of the ~10ms delay - that module (indirectly, through `lsp-status`) calls `vim.lsp.protocol.make_client_capabilities()`, which takes a surprising amount of time (accounting for ~7ms of the 10; the rest seems to mostly be unavoidable from requiring a number of modules). So, not an inherent slowdown with requiring Lua.

PR #20, opened 2020-07-22
Use pcall() for open_fn and startup user_func.

### LINKED ISSUE #18: If `open_fn` fails, there is no error warning displayed
Had a bug in my `open_fn` and Packer just silently failed. Should probably use pcall or something and then grab the values out from there. 

If it doesn't work, fall back to default maybe?

wbthomason:
Also make formatting uniform and remove unused utils

Fixes #18


wbthomason:
@tjdevries Can you check that this does, indeed, fix #18 by trying to replicate the failure you noted?

tjdevries:
heh, it's hard to check for certain, it's probably fine though if you're using pcall. I fixed the problem on my end, (and just made it not error :laughing:). You can just try probably with specifying `open_fn = function() error('this should show up') end` or something.

tjdevries:
Also, are you running lua format over the whole plugin now? I can do that as well when I make PRs.

wbthomason:
It seems to work for me, so I'll go ahead and merge. And yep, I've been trying to be consistent about running the formatter for every file as I touch it (I should probably just add a commit hook...)

tjdevries:
Maybe we can run it over the entire repo when you add the hook? I have a few other lua format things I'd suggest if you wanted. I've been trying to tweak one recently to fit my style more (still WIP).

Considering at some point, if I really like it, adding the result to nvim core so that we don't have to nitpick people on style anymore there :laughing:

wbthomason:
Yeah, please do add any suggestions you have for lua-format style!

PR #22, opened 2020-07-23
Add documentation for functions

wbthomason:


wbthomason:
This is a first-pass effort to add basic vimdocs (right now, very similar to the README) and documentation of internal functions (currently just comments in the code for the non-exposed functions). It'll probably need extension/improvement later.

PR #24, opened 2020-07-24
[WIP] Lua Rocks

tjdevries:
You can install luarocks packages and they work :)

Stil WIP, just wanted to show you where I was at.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -180,6 +180,13 @@ packer.install = function(...)
       await(a.main)
       plugin_utils.update_helptags(install_paths)
       plugin_utils.update_rplugins()
+
+      if not vim.tbl_isempty(results.rocks) then
+        neorocks.install_rocks(results.rocks)
+      else
+        log.info("No rocks to install")
+      end
+
       local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
       display_win:final_results(results, delta)
     else
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -11,8 +11,10 @@ local display = require('packer.display')
 local handlers = require('packer.handlers')
 local install = require('packer.install')
 local log = require('packer.log')
+local neorocks = require('packer.neorocks')
 local plugin_types = require('packer.plugin_types')
 local plugin_utils = require('packer.plugin_utils')
+local result = require('packer.result')
 local update = require('packer.update')
 local util = require('packer.util')
 
@@ -339,5 +341,8 @@ packer.startup = function(spec)
 
   return packer
 end
+
+-- Always set up the paths on require.
+neorocks.setup_paths(true)
 
 return packer
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -182,13 +182,15 @@ packer.install = function(...)
       plugin_utils.update_rplugins()
 
       if not vim.tbl_isempty(results.rocks) then
-        neorocks.install_rocks(results.rocks)
+        await(neorocks.install_rocks(results.rocks))
       else
         log.info("No rocks to install")
       end
 
-      local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
-      display_win:final_results(results, delta)
+      vim.schedule_wrap(function()
+        local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
+        display_win:final_results(results, delta)
+      end)()
     else
       log.info('Nothing to install!')
     end
--- a/lua/packer/install.lua
+++ b/lua/packer/install.lua
@@ -24,6 +24,12 @@ local function install_plugin(plugin, display_win, results)
 
     results.installs[plugin_name] = r
     results.plugins[plugin_name] = plugin
+
+    if plugin.rocks then
+      for _, rock in ipairs(plugin.rocks) do
+        table.insert(results.rocks, rock)
+      end
+    end
   end)
 end
 
@@ -31,6 +37,7 @@ local function do_install(_, plugins, missing_plugins, results)
   results = results or {}
   results.installs = results.installs or {}
   results.plugins = results.plugins or {}
+  results.rocks = results.rocks or {}
   local display_win = nil
   local tasks = {}
   if #missing_plugins > 0 then
--- a/lua/packer/neorocks.lua
+++ b/lua/packer/neorocks.lua
@@ -14,28 +14,33 @@ local async = a.sync
 
 local neorocks = {}
 
---- Location of hererocks.py
-neorocks._hererocks_file = util.join_paths(vim.fn.stdpath('cache'), 'hererocks.py')
+---@return table Of the form: {
+---     lua: Lua Version (5.1, 5.2, etc.)
+---     jit: Jit Version (2.1.0-beta3, or nil)
+---     dir: Directory of hererocks installation
+--- }
+neorocks._lua_version = (function()
+  if jit then
+    return {
+      lua = string.gsub(_VERSION, "Lua ", ""),
+      jit = string.gsub(jit.version, "LuaJIT ", ""),
+      dir = string.gsub(jit.version, "LuaJIT ", "")
+    }
+  end
 
---- Determine the location where neorcks should install a particular version of lua
----
----@param lua_version table The lua version table
-neorocks._hererocks_install_location = function(lua_version)
-  return util.join_paths(vim.fn.stdpath('cache'), 'packer_hererocks', lua_version.dir)
-end
+  error("NEOROCKS: Unsupported Lua Versions", _VERSION)
+end)()
+
+neorocks._base_path = util.join_paths(vim.fn.stdpath('cache'), 'packer_hererocks')
+neorocks._hererocks_file = util.join_paths(vim.fn.stdpath('cache'), 'hererocks.py')
+neorocks._hererocks_install_location = util.join_paths(neorocks._base_path, neorocks._lua_version.dir)
+neorocks._is_setup = vim.fn.isdirectory(util.join_paths(neorocks._hererocks_install_location, "lib")) > 0
 
 --- Determine if the hererocks environment is usable for nvim
 ---
 ---@return boolean True if the hererocks environment is already set up
 neorocks.is_setup = function()
-  local lua_version = neorocks.get_lua_version()
-  local install_location = neorocks._hererocks_install_location(lua_version)
-
-  if vim.fn.isdirectory(vim.fn.fnamemodify(util.join_paths(install_location, "lib"), ":p")) > 0 then
-    return true
-  else
-    return false
-  end
+  return neorocks._is_setup
 end
 
 neorocks.get_hererocks = async(function()
@@ -77,29 +82,17 @@ neorocks.get_hererocks = async(function()
         }
       end
     )
-
-
-  -- TODO: Delete this, I don't think we need it anymore
-  -- -- Just make sure to wait til we can actually read the file.
-  -- -- Sometimes the job exists before we get a chacne to do so.
-  -- vim.wait(10000, function() ~= 0 end)
-  -- vim.fn.input("[Press enter to continue]")
-  -- print("All done....")
-  -- win_float.clear(run_buf)
 end)
 
-
 --- Handle the set up of luarocks
-neorocks.setup_hererocks = function(force)
-  local lua_version = neorocks.get_lua_version()
-  local install_location = neorocks._hererocks_install_location(lua_version)
-
+neorocks.setup_hererocks = async(function(force)
   if force == nil then
     force = false
   end
 
+  local r = result.ok(true)
   if not util.path_exists(neorocks._hererocks_file) then
-    neorocks.get_hererocks()
+    r:and_then(await, neorocks.get_hererocks)
   end
 
   if neorocks.is_setup() and not force then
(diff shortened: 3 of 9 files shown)


tjdevries:
You can add a luarocks dep to something just by doing `use { 'tjdevries/vim9jit', rocks = { 'lpeg' } }` and when you do call install, it will install the rock at the end of the installation.

Not sure if that was obvious in the PR @wbthomason

wbthomason:
Yup, thanks! Sorry for not having reviewed the changes yet; I'm going to do that today.

clason:
Just to make sure I understand this correctly: for luarocks to be available to packages, `packer` _must_ be `require`d first (as opposed to pure lua and vim packages, which -- once installed -- work even if packer isn't `packadd`ed or `require`d)?

wbthomason:
@clason Not sure if it's @tjdevries intent (I'm still going through the code), but my intent for the Luarocks feature is very much the latter - rocks should be available without needing packer to be loaded.

clason:
Makes sense, just curious. (Admittedly, I was also thinking of the start package availability issue, because if packer starts handling runtime features, one could think of doing something about that, e.g., with a `force` keyword on `start` packages. But separating is indeed cleaner.)

I guess `async` is also a candidate for upstreaming (although norcalli may have other views on coroutines...)

tjdevries:
Hmmm we could try do something where we tell luarocks where to install... and it could be somewhere that's automatically in the neovim runtime.

The problem is I think it's better to let luarocks manage all of the stuff about the packages.

Maybe we could add them to a `lua` folder somewhere in a `start` area....

tjdevries:
The only thing you need is `packer.neorocks.setup_paths()` to get the paths correct.

This could be done in `plugin` or something though (but then you'd have to put it somewhere else).

clason:
So every plugin would be responsible for setting up the luarocks path themselves?

tjdevries:
And what do you think should be upstreamed about the luarocks stuff? 

It could be possible to upstream it (we already build it on basically every platform I think...)

tjdevries:
No, no plugin would set anything up, they can just do `rocks = { 'lpeg' }` and then all they have to do is `require('lpeg')` and it will work.

Packer sets up the paths currently

clason:
I might be misunderstanding how this works (I probably am), but doesn't `neorocks.setup_paths` make sure that hererocks is accessible to neovim? That part.

tjdevries:
Sure, you could copy and paste the paths and add them to your package.path and package.cpath in lua.

tjdevries:
Oh, we could put it in the compiled stuff.

lua/packer.lua
```diff
@@ -178,8 +180,17 @@ packer.install = function(...)
       await(a.main)
       plugin_utils.update_helptags(install_paths)
       plugin_utils.update_rplugins()
-      local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
-      display_win:final_results(results, delta)
+
+      if not vim.tbl_isempty(results.rocks) then
+        await(neorocks.install_rocks(results.rocks))
```
wbthomason, line 185: Is there a reason (efficiency, etc.) to wait until after all the plugin installs to do the Luarocks installs? If not, I think I would prefer to add the Luarocks install step inside the installer (like the update hooks), for higher throughput. I could make this change after merging the initial PR, if you'd prefer.

As a nitpick here, we should probably handle rock installs before updating rplugins.

lua/packer.lua
```diff
@@ -178,8 +180,17 @@ packer.install = function(...)
...
+      else
+        log.info("No rocks to install")
```
wbthomason, line 187: Nitpick: I think the "no rocks" behavior should be silence, rather than logging. That said, totally makes sense for debugging WIP.

lua/packer.lua
```diff
@@ -178,8 +180,17 @@ packer.install = function(...)
...
+      end
+
+      vim.schedule_wrap(function()
```
wbthomason, line 190: Rather than placing this in a `schedule_wrap` call, let's move the `await(a.main)` line down after all the async code, to keep things cleaner. This will also entail moving the helptags and rplugins updates, since they too need to be on the main thread.

lua/packer/util.lua
```diff
@@ -43,6 +43,29 @@ util.join_paths = function(...)
   return table.concat({...}, separator)
 end
 
+util.absolute_path = function(p)
+  return vim.fn.fnamemodify(p, ":p")
+end
+
+util.path_exists = function(p)
```
wbthomason, line 50: Just curious - what's the rationale for this approach over e.g. `filereadable()`?

lua/packer/window.lua
```diff
@@ -0,0 +1,77 @@
...
+--- Create window that takes up certain percentags of the current screen.
+---
+--- Works regardless of current buffers, tabs, splits, etc.
+--@param col_range number | Table:
+--                  If number, then center the window taking up this percentage of the screen.
+--                  If table, first index should be start, second_index should be end
+--@param row_range number | Table:
+--                  If number, then center the window taking up this percentage of the screen.
+--                  If table, first index should be start, second_index should be end
+window.percentage_range_window = function(col_range, row_range, options)
```
wbthomason, line 22: This seems like it might be better suited to live in `packer.display`, but that's a minor point and easily fixed up after merge.

lua/packer/util.lua
```diff
@@ -94,4 +117,17 @@ util.deep_extend = function(policy, ...)
   return result
 end
 
+util.system_output = function(command)
```
wbthomason, line 120: Is this intended for running system commands that should not be run asynchronously?

lua/packer/install.lua
```diff
@@ -24,13 +24,20 @@ local function install_plugin(plugin, display_win, results)
 
     results.installs[plugin_name] = r
     results.plugins[plugin_name] = plugin
+
+    if plugin.rocks then
+      for _, rock in ipairs(plugin.rocks) do
+        table.insert(results.rocks, rock)
```
wbthomason, line 30: As mentioned in the comments on `packer.lua`, I think it would be better to handle the actual install here. I'm happy to write that up if you'd like.

lua/packer/neorocks.lua
```diff
@@ -0,0 +1,348 @@
...
+        return {
+          msg = "Failed to get hererocks",
+          data = err
+        }
+      end
+    ):map_ok(
+      function(ok)
+        return {
+          status = ok,
+          -- TODO: Figure out what we should be doing with the output
```
wbthomason, line 80: The output is usually collected on the `results` table (for installs and updates), for use in the operation status display. Here, we may not care about the output at all. We could probably just drop this entire `map_ok` call.

lua/packer/neorocks.lua
```diff
@@ -0,0 +1,348 @@
...
+    --     return t[k]
+    --   end
+    -- })
+
+    -- -- Install the package
+    -- local exec_luarocks = loadfile(util.join_paths(neorocks._hererocks_install_location, "bin", "luarocks"))
+    -- -- Redirect io.stdout & io.stderr
+    -- setfenv(exec_luarocks, new_env)('install', rocks[1])
+
+    -- TODO: I would really like to go back to the loadfile method and not shoot this out to a shell...
```
wbthomason, line 234: What's the loadfile method, out of curiosity?

lua/packer/neorocks.lua
```diff
@@ -0,0 +1,348 @@
...
+    --          I'm worried about trying to compile multiple of these at the same time.
+    -- TODO: move to io.popen, so we can pipe the output to the display
```
wbthomason, line 236: Using `io.popen` shouldn't be necessary if we run things through `jobs` - it can collect the output of a process and call a display update callback.

lua/packer/neorocks.lua
```diff
@@ -0,0 +1,348 @@
...
+    -- vim.schedule_wrap(function()
+      -- TODO: Figure out how we could queue up these lua rocks commands to do in job start
```
wbthomason, line 238: This is doable using https://github.com/wbthomason/packer.nvim/blob/a96f5f1ec20b758b27defdec433c664d57d48abe/lua/packer/async.lua#L60 with an `n` of 1 (for sequential execution).

lua/packer.lua
tjdevries: I don't think we should do it inside of each plugin because there is a chance:

1. they try to install to the same place
2. they try to install conflicting deps
3. they just end up getting messed up because luarocks wasn't meant to be concurrent.

So I have them running sequentially at the end right now because that's the only way I could think of solving these problems. Perhaps there is a way to fix it though.

lua/packer.lua
tjdevries: Added TODO

lua/packer.lua
tjdevries: hmmm I'm not sure exactly what you mean.

lua/packer/util.lua
tjdevries: I want to use some of these functions when we're not in `vim.schedule_wrapped` stuff. I was getting errors when I was trying to do that some of the time, so I figure I can just use the Lua way to do it.

lua/packer/window.lua
tjdevries: Yeah, just copied it in from another project of mine, so wanted to leave it separate for now. Can be merged into display whenever you want (I'll leave it to you to decide how)

lua/packer/util.lua
tjdevries: Yeah, I want to wait on the output of `luarocks list lpeg` and what not to check if somethign is already installed.

lua/packer/install.lua
tjdevries: I'll wait on your thoughts about the other section.

lua/packer/neorocks.lua
tjdevries: loadfile method would be we figure out how to call the luarocks stuff not from command line but actually from lua.... this didn't seem to work great.

lua/packer/neorocks.lua
tjdevries: This is more complicated than that -- I can explain after we chat about a few of the other things first.

lua/packer.lua
wbthomason: Hmm, fair points (particularly (3)). (2) seems like it could be an issue regardless of how we run luarocks - do you know how luarocks handles conflicting/multiple versions of packages in "normal" use? If it doesn't already handle that for us, we might need to install each plugin's rocks to a separate location. That adds some complication/increases disk space use by duplicating packages, but does make things like rock removal, etc. much easier as well. If we do need to use that approach, then I still think running the install inside each plugin is better; if we can avoid that approach, async-running the install after all the plugin installs should be fine.

lua/packer.lua
wbthomason: Ah - `await(a.main)` (on line 180) is a little hack to ensure that the code after that line until the next `await` call is running on the main event loop (which I sometimes incorrectly call the "main thread"). It's currently used to make sure that updating the helptags and remote plugins happens on the main event loop. Here, I'm suggesting that we could move the async code that currently runs *after* the helptags, etc. (which is the neorocks install call) to *before* the call to `await(a.main)`, which would mean that the code you have wrapped with `vim.schedule_wrap` here no longer needs to be wrapped (because there will no longer be any async code between it and the last `await(a.main)` call). That should make this cleaner and very slightly more efficient (no need to wait for the main event loop twice).

lua/packer/util.lua
wbthomason: Fair enough - I think, though, that using `vim.loop.fs_access` (https://github.com/luvit/luv/blob/master/docs.md#uvfs_accesspath-mode-callback) would be a bit better (should be faster than opening and closing the file, and is more semantically correct).

lua/packer/window.lua
wbthomason: Sounds good; I'll leave it where it is at least until everything else is working. I want to do some refactoring of the overall codebase soon anyway (I think a lot of the git code, in particular, could be cleaned up and deduplicated), so I might handle all that after we merge this PR.

lua/packer/install.lua
wbthomason: Yeah, per discussion on the other comment, I think we need to determine how conflicts will be handled before deciding where to do the actual install.

tjdevries:
(did you mean to push to this branch?) 

I'll work a bit more on this in the next week I think.

wbthomason:
Yup - sorry for the confusion. I was doing some cleanup of other issues and PRs, which created a small merge conflict for this PR. Just fixing that up!

lua/packer.lua
tjdevries: So I thinkg lua rocks itself already handles the fact that when you remove one package it cleans up the deps of the package (if no one else is using it). But I haven't followed up on that. We can always add that as an enhancement later (since I think our implementation may get a little more complicated over time :laughing: )

lua/packer.lua
tjdevries: Hm, I'm still not sure I 100% understand what you're thinking for where to place it and how tht works. I'm fine w/ either you adding a commit to the branch or we just do it after (if it's not a big deal)

lua/packer/neorocks.lua
tjdevries: Ah, I added half of this already but forgot to use $SHELL in the actual command. Yup, good call.

tjdevries:
I wouldn't be worried about the big comment block about switching back... that's primrily about trying to do luarocks from a module, which may not be possible.

tjdevries:
I can try and clean up the PR soon and we can merge v1 support of it? What do you think?

lua/packer.lua
wbthomason: Sure - let's get V1 ready except for this, then I can add a commit before merging.

wbthomason:
Sounds good, thank you! I'm also happy to help with this more directly; I haven't wanted to step on your toes.

tjdevries:
tbh, I don't know how much time I have to finish it up right now -- if you want to take it to the finish line, that'd be great (especially since a lot of the back and forth we'll be doing I think will just be making it fit better into some more areas of the existing code base)

One thing I was thinking as well is it would be great to expose the "install" commands as able to run easily from other plugins.

So if someone didn't want to use packer, but did want to install luarocks stuff easily, they could just do `require('packer.neorocks').install('lua-cjson')` for example.

tjdevries:
(Happy to reply to comments though as you're working through things or clear anything up in the issues we've chatted about above)

wbthomason:
Sounds good, thanks! I'll allocate some time for this later this week/early next week.

Just so we're on the same page, here's what I think needs to be done for v1:
1. Resolve current comments
2. Add update logic (should check for missing dependencies, removed dependencies, changed versions)
3. Add removal logic (should uninstall rocks when uninstalling - but not moving - plugins)
4. Make install, update, remove commands accessible without directly using `packer`.
5. Update docs to document new feature

Does that seem right to you?

tjdevries:
Yeah, sounds good. I haven't heard back from anyone on luarocks about using luarocks as a module, so it doesn't seem possible currently, so we can probably just go forward w/ the shelling stuff out to system for now (and potentially forever).

wbthomason:
I'm going to close this for now in favor of #104, which builds off of the more recent/stable Neorocks code in `plenary`. Thanks @tjdevries for this initial work and for the excellent starting point in `plenary`!

PR #27, opened 2020-07-31
Fix config of type function not being executed

gbrlsnchs:


wbthomason:
Shoot, nice catch! Thank you!

Issue #29, opened 2020-08-09
"if" in requires


p00f:
Currently I have
```lua
  use {
    'nvim-lua/completion-nvim',
    opt = true,
    requires = {{'hrsh7th/vim-vsnip', opt = true}, {'hrsh7th/vim-vsnip-integ', opt = true}, {'ncm2/float-preview.nvim', opt = true}, {'aca/completion-tabnine', opt = true}}
  }
```

I want to `require` ncm2/float-preview.nvim only if a particular global variable `g:xyz` is not set. How would I do that?

p00f:
Also, i don't have to mark `opt=true` manually if I have `ft={'xyz'}` right?

p00f:
I have this but it doesnt work (ncm2/float-previe.nvim gets added anyway)
```lua
reqcomp = {{'hrsh7th/vim-vsnip'}, {'hrsh7th/vim-vsnip-integ'}, {'aca/completion-tabnine'}}
if( not( vim.g.gonvim_running )) then
  table.insert(reqcomp, {'ncm2/float-preview.nvim'})
end

local packages = {
>other plugins here<
  {'nvim-lua/completion-nvim',
    requires = reqcomp
  },
>other plugins here<
}
return packages
```
(See [this](https://github.com/wbthomason/packer.nvim/issues/30#issuecomment-671187173) for my setup)

wbthomason:
Assuming that by "gets added anyway" you mean "is loaded", that's because you have `float-preview.nvim` specified as a `start` plugin, not an `opt` plugin (see the docs on the built-in packages feature in Neovim for the definitions of those kinds of plugins; basically, `start` plugins are always sourced at startup, whereas `opt` plugins must be manually loaded).

Because it uses native packages rather than the `vim-plug` method of appending to `runtimepath`, plugin specification with `packer` is a bit different than with `vim-plug`. In particular, naming a plugin means it gets installed; configuring loading happens by specifying something like `opt`, `setup`, `ft`, `cond`, etc. in its specification.

This is actually a great example use case for the `cond` configuration keyword. `cond` in general is used to define a test to determine if an `opt` plugin should be loaded. So, you could instead write the following:

```lua
use {
  'nvim-lua/completion-nvim',
  requires = {
    'hrsh7th/vim-vsnip', 'hrsh7th/vim-vsnip-integ', 'aca/completion-tabnine',
[...]

PR #31, opened 2020-08-10
Add plugin alias feature to close #28

### LINKED ISSUE #28: Custom name
How do I do the vim-plug equivalent of `Plug 'dracula/vim', { 'as': 'dracula' }` (ie install it in a folder named `dracula` or else it would be installed in a folder named `vim` because the repository is [https://github.com/dracula/***vim***](https://github.com/dracula/vim))

wbthomason:
Adds the `as` keyword to allow plugins to be installed under aliases. Closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,9 @@ return require('packer').startup(function()
 
   -- You can specify multiple plugins in a single call
   use {'tjdevries/colorbuddy.vim', {'nvim-treesitter/nvim-treesitter', opt = true}}
+
+  -- You can alias plugin names
+  use {'dracula/vim', as = 'dracula'}
 end)
 ```
 
@@ -188,6 +191,7 @@ use {
   'myusername/example',   -- The plugin location string
   -- The following keys are all optional
   disable = boolean,           -- Mark a plugin as inactive
+  as = string,                 -- Specifies an alias under which to install the plugin
   installer = function,        -- Specifies custom installer. See "custom installers" below.
   updater = function,          -- Specifies custom updater. See "custom installers" below.
   after = string or list,      -- Specifies plugins to load before this plugin.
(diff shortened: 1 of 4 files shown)


p00f:
Like this `{'dracula/vim', as = 'dracula'},`?

wbthomason:
Yep! That's a good point; I need to update the docs too.

p00f:
Ik i'm being lazy, but waht if i simply add `branch = feat-add-plugin-aliases` ?

wbthomason:
That should work; if it doesn't, the code for handling branch checkouts has bugs (which is possible; to my knowledge that code hasn't been used/tested much yet).

p00f:
Branch checkout code works, it uses the correct branch. Changed branch, updated packer, packercompile, deleted old `vim` folder, packerinstall. It cloned into `vim` again

p00f:
Can you try it?

wbthomason:
Yes, I've tried that exact sequence, and it installs to `dracula` for me. You shouldn't need the `PackerCompile` invocation in the middle, but that also shouldn't matter.

Sorry for all the hassle, but could you manually confirm that `packer` is on the right branch?

I may also just go ahead and merge this; I'm fairly confident from my own testing that it at least doesn't break things.

p00f:
Sure!

p00f:
Can you check if it appears in `PackerClean` though? After [this](https://github.com/wbthomason/packer.nvim/issues/29#issuecomment-671675946), `as` works but `dracula` appears in PlugClean

wbthomason:
Good catch, thank you! I've pushed bd1ccec which should fix this.

wbthomason:
I'll also note that if you ever *change* an alias, right now we don't have a good way to detect the change, so what happens is the plugin gets reinstalled under the new alias and the next `clean` operation you run will delete the copy under the old alias. This isn't ideal, so I'll try to figure out a better solution in the future.

Issue #43, opened 2020-09-05
Can't install plugins


TimUntersberger:
**init.vim**
```vimscript
packadd packer.nvim 

execute 'luafile ' . stdpath('config') . '/lua/plugins.lua'

command! PackerInstall lua require('plugins').install()
command! PackerUpdate lua require('plugins').update()
command! PackerSync lua require('plugins').sync()
command! PackerClean lua require('plugins').clean()
command! PackerCompile lua require('plugins').compile(
```

**lua/plugins.lua**
```lua
return require('packer').startup(function()
    use {'wbthomason/packer.nvim', opt = true}
    use 'morhetz/gruvbox'
end)
```

When I run `:PackerCompile` followed by `:PackerInstall` it fails to install both plugins.

I am using `NVIM v0.5.0-659-g7b1b271f4` on `Windows 10`.

Can I somehow provide you with some logs?

wbthomason:
Fascinating. Do you have `packer` cloned to somewhere on your `packpath`? What does happen when you run those commands?

TimUntersberger:
I tried to see what happens exactly when i run `:PackerInstall`, but going from running the command to failed installation is so fast that I can't really see what is happening. I do notice that it is cloning the repo. 

I cloned `packer` into `%LOCALAPPDATA%\nvim-data\site\pack\packer\opt\packer.nvim`.

After looking through the folders I found that `packer` is now downloaded twice (even though one created by `packer` is empty).

![image](https://user-images.githubusercontent.com/32014449/92320112-475ab200-f01f-11ea-8550-44f7cbd66227.png)

Nothing besides my manually created folders exist in `%LOCALAPPDATA%\nvim-data\site\pack`.

In `C:\Users\timun\AppData\Local\nvim-data\site\pack\packer\start` a folder called `morhetz` (for gruvbox) does exist, but it is also empty.

My assumption is that something goes wrong when `packer` tries to clone the repos. This is 90% a configuration error on my part, but because I don't know how I can see logs or errors I can't really debug this on my own.

TimUntersberger:
Another thing I don't really understand is, why do you need to do

```lua
execute 'luafile ' . stdpath('config') . '/lua/plugins.lua'
```

and not just

```lua
require('plugins')
```

TimUntersberger:
I think I found the problem.

![image](https://user-images.githubusercontent.com/32014449/92321340-6fe7a980-f029-11ea-8a64-30791979ee9c.png)

[...]

PR #45, opened 2020-09-06
fix: windows related bugs

TimUntersberger:
Fixed the problems mentioned in #43 .

# Fixes

### in lua numbers are always true

> When used as control expression, the only false values in Lua are false and nil , everything else is evaluated as true value

This means that the following if statement is always true, because `vim.fn.executable` returns a number

```lua
vim.fn.executable('ln')
```

### `mklink` requires admin privileges

In Windows 10 you need to have developer mode enabled to use local plugins. This should probably be mentioned in the README.

I don't know where to put this, so could you please do this?

### `mklink` signature

`mklink` has the `from` and `to` swapped.

### powershell instead of cmd

It looks like your job lib is using powershell and not cmd. This means that you have to use `cmd /C mklink ...` instead of `mklink ...`

### mklink doesn't automatically create directory symlinks

Don't think I need to explain this. You just forgot the `/d` flag of `mklink`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,14 @@ Neovim.
 - Support for `git` tags, branches, revisions, submodules
 - Support for local plugins
 
+## Requirements
+- You need to be running Neovim v0.5.0+; `packer` makes use of extmarks and other newly-added Neovim
+  features.
+- Your version of Neovim must be compiled with LuaJIT support; `packer` relies on this to detect
+  whether you are running Windows or a Unix-like OS (for path separators)
+- If you are on Windows 10, you need developer mode enabled in order to use local plugins (`packer`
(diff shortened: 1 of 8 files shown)


wbthomason (approved):
These are very helpful, thank you!

wbthomason:
I think the correct addressing to your remaining comment about changing the separator in `get_plugin_full_name` may be valid but is better addressed by a refactoring of the (currently messy and overloaded) way I use plugin full vs short names. I've made #47 to track that.

If the current version of this PR isn't any more broken for you, and actually fixes #43 on Windows, I'll go ahead and merge it.

Thank you again for the help!

TimUntersberger:
I am now using this PR and I haven't noticed any problems yet.

lgtm

wbthomason:
Great, thank you again!

PR #59, opened 2020-09-29
Added so the buffer shows info when open

Kirikmelet:
Before when I called PackerSync or PackerUpdate, I would get an error in display.lua line 388; which reads 'Buffer is not "modifiable"'.
packer_error
So to fix it, I added  `api.nvim_buf_set_option(buf, 'modifiable', true)` in `make_header` above the function to get width.

### DIFF
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -383,6 +383,7 @@ end
 
 --- Utility to make the initial display buffer header
 local function make_header(disp)
+  api.nvim_buf_set_option(buf, 'modifiable', true)
   local width = api.nvim_win_get_width(0)
   local pad_width = math.floor((width - string.len(config.title)) / 2.0)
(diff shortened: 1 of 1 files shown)


wbthomason:
Hmm, could you give me more information on your setup? The display window has been working without error for me (and others) for months now, so (unless there's been a breaking change on Neovim HEAD), I'm not sure why you would've seen this issue.

Thanks for the PR, also!

Kirikmelet:
Sorry, I found the issue I was having, turns out an autogroup I was using to save folds caused the buffer to be set nomodifiable.

Should I close this pull request or should I leave it open?

wbthomason:
Ah, I'm glad you got things working. Let's close this PR, but if you'd like to make another PR adding a "Troubleshooting" section to the README with this information, that'd be very welcome! No worries if you'd rather not, though.

Thanks!

Kirikmelet:
Ok

PR #74, opened 2020-10-27
Fix: tidy display of installation errors. Fixes #73

### LINKED ISSUE #73: Packer update display doesn't cater for broken updates well.
I have a package with local changes, that I know will fail during updates. I'm happy to have the error message in the window.

If subsequent package updates are successful however, their status messages get intermingled with the output from the failed update. 

packer_update

sunjon:
removes the `vim.inspect()` wrapping the installation error message and splits the message by newline for correct display


wbthomason (approved):
Looks good, thank you!

Issue #42, opened 2020-09-05
it hangs when misspel github package name


bfredl:
I did `use 'vim-cnof-live/pres.vim'` and it freezed up. I would expect no freeze and an async error message a few secs later when `git` gives up.

wbthomason:
How long did you wait? Clones should time out after the value of `config.git.clone_timeout` (...which I now realize I never added to the docs, whoops), which defaults to 60s (based on the timeout value used in `plug.vim`). The freeze is bizarre, though - the entire editor froze up?

wbthomason:
I also cannot reproduce this on my machines....exactly what you describe happens for me....

bfredl:
ok, I will try again today..

bfredl:
yes it hangs on `username for https://github.com/xx/yy` message (the entire neovim UI)

bfredl:
it unhangs after 60s yes, but neovim should not freeze during this minute.

wbthomason:
Agreed; I'll look into why it's freezing. Could you post your .gitconfig, since I can't seem to replicate the behavior on my system?

wbthomason:
Hmm. I'm shooting in the dark with this theory as I still cannot reproduce the bug, but I wonder if this would be caused by not passing a `stdin` fd to the `luv.spawn` call used to start `git`? Since you mention that it hangs on a prompt for username...

bfredl:
@wbthomason I would guess you have to override the stdin explicitly. We should make our `vim.loop` documentation clearer about this.

wbthomason:
I pushed a commit just now adding this. It doesn't break anything on my end, but I have still never been able to repro the issue here (or in #40 or #26, which I think are duplicates of this), so I'm not sure if it fixes the issue.

wbthomason:
Fixed by #91

Issue #40, opened 2020-08-30
wrong github package name leads to UI freeze


bfredl:
I misspelled an `use 'githubuser/pakcage'` and the UI locks up with `Username for 'https://github.com':  1 / 5 plugins` message.
It should give an error message instead.

wbthomason:
Thanks for the report! I'm having trouble reproducing this - would you mind telling me the specific typo you made? For instance, testing with
```lua
use 'junegunn/vim-peekabool'
```
fails to install as expected, without freezing the UI.

I'm wondering if your typo accidentally stumbled upon a private repo name, which could make Git prompt for authentication?

Regardless, I should add some timeout/job stopping logic for installs (though probably not build stages; plugins requiring compilation sometimes take a while to finish).

wbthomason:
I'm wondering if this is related to #26 - I'm still not able to reproduce either issue, but I've pushed a timeout feature in fd3a94f which should hopefully help. Could you please give it a try and let me know if it resolves the issue?

I still need to add documentation for the associated new config option (`config.git.clone_timeout`), but the default timeout is 60s (taken from `plug.vim`'s default timeout value); you might want to set that to be shorter for testing.

nanotee:
Leaving this here just in case it turns out to be useful: Git has a `GIT_TERMINAL_PROMPT` environment variable that deactivates the password prompt (which you probably don't want when using Git non-interactively)

https://serverfault.com/a/665959

Issue #26, opened 2020-07-26
Install freezes if not a valid git repository


tjdevries:
If you try and install an invalid repo using packer, the entire Neovim freezes and you can't do anything about it.

I was able to reproduce with `use 'asdfasdfasdf/asdfasdfasdf'`

(This is on WSL, I can check somewhere else later as well)

clason:
I just tested on macOS (using the table version of `startup`), and got a proper error message (no freeze)

wbthomason:
Hmm, yeah, I cannot reproduce this on Arch Linux. Maybe related to #25/WSL-specific?

Shougo:
It maybe git command freezes

tjdevries:
(I ended up just instaling ubuntu on my laptop because I was having some troubles with WSL... so it may be awhile before I get around to testing this :laughing: ) Not a high priority for me and user error in some ways anyway ;)

wbthomason:
Presumed fixed by #91

Issue #63, opened 2020-10-12
Some edge cases leading to cryptic errors


gwerbin:
Hi, thanks for the great plugin manager!

I found a few rough edges where the user "falls through" to Packer internals, leading to cryptic errors. Hopefully this report is helpful in future development.

### System details

`uname -orsv` output: `Linux 5.8.0-2-amd64 #1 SMP Debian 5.8.10-1 (2020-09-19) GNU/Linux`

Packer commit: `237a134d5144f9f99df1a2f5bb29950c3d0ed795`

`nvim --version` output:
<details>
NVIM v0.5.0-746-gf7cc3ae0b
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: /usr/bin/gcc-5 -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/travis/build/neovim/bot-ci/build/neovim/build/config -I/home/travis/build/neovim/bot-ci/build/neovim/src -I/home/travis/build/neovim/bot-ci/build/neovim/.deps/usr/include -I/usr/include -I/home/travis/build/neovim/bot-ci/build/neovim/build/src/nvim/auto -I/home/travis/build/neovim/bot-ci/build/neovim/build/include
Compiled by travis@travis-job-14258239-027b-4892-8d10-88b6f65b3f23

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "
/home/travis/build/neovim/bot-ci/build/neovim/build/nvim.AppDir/usr/share/nvim
"

Run :checkhealth for more info
</details>

### Problem 1: quitting Packer too fast?

When I run `nvim +PackerCompile +PackerUpdate +qall`, I get a bunch of the following error repeated over and over:
```
[...]

PR #91, opened 2020-11-19
Fix git prompt leading to UI freeze

nanotee:
I have managed to reproduce issue #40 (and #42 / #26, possibly?), this PR implements a fix using the `GIT_TERMINAL_PROMPT` environment variable.

I would really appreciate it if people who encounter the issue could test the PR to see if it fixes the problem for them as well

### DIFF
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -123,7 +123,13 @@ git.setup = function(plugin)
       stderr = jobs.logging_callback(output.err.stderr, output.data.stderr, nil, disp, plugin_name)
     }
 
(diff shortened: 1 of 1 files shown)


wbthomason:
@nanotee Thank you! I've been meaning to get to this, but hadn't made the time.

I think @bfredl was the primary user experiencing this issue. Also @gwerbin in #63. If either of you all (or anyone else who can reproduce the issue) could verify that this PR fixes the problem for you, I'll happily merge this.

bfredl:
@wbthomason yes it fixes it :)

wbthomason:
Great, thank you! (and thanks @nanotee for the PR!)

PR #101, opened 2020-11-25
Fix #100: copy existing environment variable values

### LINKED ISSUE #100: PackerInstall fails
After 06caefb116998aa60fab37e54137942da4de4755 / #91 `:PackerInstall` fails. This is on a network that requires a proxy and it looks like the existing environment variable `https_proxy` is no longer used. Adding that variable to the env config in `lua/packer/plugin_types/git.lua` fixes the issue. It seems like it would make sense to keep the old behavior of using variables that are already set rather than needing to add this additional configuration.

wbthomason:
Copy the entire environment before assigning `GIT_TERMINAL_PROMPT`

### DIFF
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -123,11 +123,13 @@ git.setup = function(plugin)
       stderr = jobs.logging_callback(output.err.stderr, output.data.stderr, nil, disp, plugin_name)
     }
 
+    local job_env = vim.loop.os_environ()
+    job_env['GIT_TERMINAL_PROMPT'] = 0
     local installer_opts = {
       capture_output = callbacks,
       timeout = config.clone_timeout,
       options = {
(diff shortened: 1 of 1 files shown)


ranebrown:
I'm getting this error `[packer] Error in coroutine: ...k/packer/opt/packer.nvim/lua/packer/plugin_types/git.lua:126: attempt to call field 'os_environ' (a nil value)`

wbthomason:
Hmm, ok. Could you please post your Neovim version? I cannot reproduce that failure on my end right now.

ranebrown:
NVIM v0.5.0-7c4f34966
Build type: Release
LuaJIT 2.0.5

wbthomason:
Thanks @nanotee and @ranebrown for help testing! Please let me know if this does not fix the issue (as I've still been unable to reproduce #26, etc.)!

ranebrown:
Seems to be working. Thanks!

Issue #87, opened 2020-11-15
Behaviour of requires seems slightly unintuitive


rickysaurav:
When plugin A requires plugin B, I expect Plugin B to be available before Plugin A starts loading. Packer adds Plugin B to "after" of A i.e. plugin B is loaded after plugin A is loaded, which is slightly unintuitive in terms of what I would expect from a dependency. 
This causes issues when the dependency of an optional plugin is not available in "start" packages.
Reproduction: 
```lua 
vim.cmd [[packadd packer.nvim]]

local packer = require("packer")

packer.startup(
    function(use)
        -- Packer can manage itself as an optional plugin
        use {"wbthomason/packer.nvim", opt = true}
        use {
            "nvim-lua/telescope.nvim",
            requires = {{"nvim-lua/popup.nvim", opt = true}, {"nvim-lua/plenary.nvim", opt = true}},
            cmd = {"Telescope"},
        }
    end
)
```
Calling "Telescope find_files" results in the following error.
```
Error detected while processing function <SNR>8_load:
line    1:
E5108: Error executing lua [string ":lua"]:111: Vim(let):E5108: Error executing lua ...ite/pack/packer/opt/telescope.nvim/lua/telescope/log.lua:2: module 'plenary.log' not found:
        no field package.preload['plenary.log']
        no file './plenary/log.lua'
        no file '/usr/share/luajit-2.0.5/plenary/log.lua'
        no file '/usr/local/share/lua/5.1/plenary/log.lua'
        no file '/usr/local/share/lua/5.1/plenary/log/init.lua'
        no file '/usr/share/lua/5.1/plenary/log.lua'
        no file '/usr/share/lua/5.1/plenary/log/init.lua'
        no file './plenary/log.so'
        no file '/usr/local/lib/lua/5.1/plenary/log.so'
        no file '/usr/lib/lua/5.1/plenary/log.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
        no file './plenary.so'
        no file '/usr/local/lib/lua/5.1/plenary.so'
        no file '/usr/lib/lua/5.1/plenary.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
```

wbthomason:
`requires` is meant to specify dependencies, but not loading order. For loading order, there's the `after` key.

[...]

PR #104, opened 2020-11-27
Luarocks, take 2

wbthomason:
This is still very work-in-progress, adapting @tjdevries lovely https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/neorocks/init.lua to work with `packer`'s job and display system (and adding plugin-specific logic).

### DIFF
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -231,6 +231,7 @@ local function clean_packages(plugins, disp)
       for _, package in ipairs(installed_packages) do to_remove[package.name] = package end
       for _, plugin in pairs(plugins) do
         if plugin.rocks then
+          if type(plugin.rocks) == 'string' then plugin.rocks = {plugin.rocks} end
           for _, rock in ipairs(plugin.rocks) do
             if type(rock) == 'table' then
               if to_remove[rock[1]] and to_remove[rock[1]].version == rock[2] then
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -319,7 +319,7 @@ local function handle_command(cmd, ...)
   local packages = {...}
   if cmd == 'install' then
     task = install_packages(packages)
-  elseif cmd == 'clean' then
+  elseif cmd == 'remove' then
     task = uninstall_packages(packages)
   else
     log.warning('Unrecognized command!')
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -89,14 +89,20 @@ end
 local function generate_path_setup_code()
   local package_path_str = vim.inspect(package_paths)
   local install_cpath = util.join_paths(hererocks_install_dir, 'lib', 'lua', lua_version.lua)
-  local install_cpath_pattern = fmt('%s?.so', install_cpath)
-  return [[if not string.find(package.path, ]] .. package_path_str .. [[, 1, true) then
-    package.path = package.path .. ';' .. ]] .. package_path_str .. [[
+  local install_cpath_pattern = fmt('"%s?.so"', install_cpath)
+  return [[
+  local package_path_str = ]] .. package_path_str .. [[
+
+  local install_cpath_pattern = ]] .. install_cpath_pattern .. [[
+
+  if not string.find(package.path, package_path_str, 1, true) then
+    package.path = package.path .. ';' .. package_path_str
   end
 
-  if not string.find(package.cpath, ]] .. install_cpath_pattern .. [[, 1, true) then
-    package.cpath = package.cpath .. ';' .. ]] .. install_cpath_pattern .. [[
-  end]]
+  if not string.find(package.cpath, install_cpath_pattern, 1, true) then
+    package.cpath = package.cpath .. ';' .. install_cpath_pattern
+  end
+]]
 end
 
 local function activate_hererocks_cmd(install_path)
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -253,23 +253,24 @@ end
 
 local function ensure_packages(plugins, disp)
   return async(function()
-    local r = result.ok()
-    if not hererocks_is_setup() then r = r:and_then(await, hererocks_installer(disp)) end
-    r = r:and_then(await, luarocks_list(disp))
-    r = r:map_ok(function(installed_packages)
-      local to_install = {}
-      for _, plugin in pairs(plugins) do
-        if plugin.rocks then
-          if type(plugin.rocks) == 'string' then plugin.rocks = {plugin.rocks} end
-          for _, rock in ipairs(plugin.rocks) do
-            if type(rock) == 'table' then
-              to_install[rock[1]] = rock
-            else
-              to_install[rock] = true
-            end
+    local to_install = {}
+    for _, plugin in pairs(plugins) do
+      if plugin.rocks then
+        if type(plugin.rocks) == 'string' then plugin.rocks = {plugin.rocks} end
+        for _, rock in ipairs(plugin.rocks) do
+          if type(rock) == 'table' then
+            to_install[rock[1]] = rock
+          else
+            to_install[rock] = true
           end
         end
       end
+    end
+    local r = result.ok()
+    if #to_install == 0 then return r end
+    if not hererocks_is_setup() then r = r:and_then(await, hererocks_installer(disp)) end
+    r = r:and_then(await, luarocks_list(disp))
+    r = r:map_ok(function(installed_packages)
 
       for _, package in ipairs(installed_packages) do
         local spec = to_install[package.name]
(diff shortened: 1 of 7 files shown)


tjdevries:
Will it be possible to use this without using packer as your package manager?

tjdevries:
`PackerLuarocks install busted` or just `NeovimLuaRocks` or `Luarocks` or something and then pass the correct things to the correct places.

wbthomason:
This should be ready for testing. Basic operations seem to work:

- [x] Specifying Lua packages via the `rocks` key on a plugin spec
- [x] Installing missing rocks
- [x] Cleaning unused rocks
- [x] Installing/setting up `hererocks` and `luarocks`

Still remaining:
- [ ] Add option to check for a file specifying rocks in a plugin's directory (what format should this file be?)
- [ ] Test installing different versions of a rock
- [ ] (maybe) add support for semantic versioning in determining rocks to install?
- [ ] Add commands like @tjdevries requests to use this outside of `packer` as a plugin manager (the necessary functions exist already)
- [ ] Cleanup/improve integration with display (e.g. to display cleaned rocks)

Also, for myself, this has brought to light a few areas for improvement in `packer`'s code (probably not part of this PR):
- The jobs API is clunky (specifically with making and manipulating output tables)
- The results API is clunky (should modify objects in-place)
- There's too much duplicated code in the main management operations; this should be simplified

wbthomason:
If anyone has a chance to try this and validate that things work for them and not just me, that would be helpful!

wbthomason:
Also, if anyone is more familiar with Luarocks: is it proper to expect plugins to provide a `rockspec` listing the rocks they require? Or is another format (e.g. a TOML file listing requirements) more appropriate?

tjdevries:
@wbthomason I'd ask in the luarocks gitter. They usually answer questions there.

lua/packer/luarocks.lua
```diff
@@ -0,0 +1,305 @@
...
+
+local cache_path = vim.fn.stdpath('cache')
+local rocks_path = util.join_paths(cache_path, 'packer_hererocks')
+local hererocks_file = util.join_paths(rocks_path, 'hererocks.py')
+local hererocks_install_dir = util.join_paths(rocks_path, lua_version.dir)
+local _hererocks_setup_done = false
+local function hererocks_is_setup()
+  if _hererocks_setup_done then return true end
+  await(a.main)
+  _hererocks_setup_done = vim.fn.isdirectory(util.join_paths(hererocks_install_dir, 'lib')) > 0
```
tjdevries, line 27: May want to use vim.loop.fs_stat here. In general I'd say you want to avoid as much `vim.fn` here in case people call it during weird situations. But probably works for now though :)

lua/packer/luarocks.lua
```diff
@@ -0,0 +1,305 @@
...
+      stderr = jobs.logging_callback(output.err.stderr, output.data.stderr)
+    }
+
+    local opts = {capture_output = callbacks}
+    local r = await(jobs.run(hererocks_cmd, opts)):map_err(
+                function(err)
+        return {msg = 'Error installing hererocks', data = err, output = output}
+      end)
+
+    local luarocks_cmd = 'python ' .. hererocks_file .. ' --verbose -j ' .. lua_version.jit
```
tjdevries, line 59: may want to be able to configure `python` command somehow, in case it's not in path or need to specify certain version, etc.

lua/packer/luarocks.lua
```diff
@@ -0,0 +1,305 @@
...
+                           .. ' -r latest ' .. hererocks_install_dir
+    r = r:and_then(await, jobs.run(luarocks_cmd, opts)):map_err(
+          function(err) return {msg = 'Error installing luarocks', data = err, output = output} end)
+    return r
+  end)
+end
+
+local function package_patterns(dir) return fmt('%s?.lua;%s&/init.lua', dir, dir) end
```
tjdevries, line 67: btw, I think some of these path things may be messed up, I'd just check that this works for all the cases. I noticed I had some bugs in my path code in plenary.

tjdevries: (i.e. just check the package.path when you're done and make sure it looks right for these cases.

lua/packer/luarocks.lua
```diff
@@ -0,0 +1,305 @@
...
+  if shell == 'fish' then
+    activate_file = 'activate.fish'
+  elseif shell == 'csh' then
+    activate_file = 'activate.csh'
+  end
+
+  return fmt('source %s', util.join_paths(install_path, 'bin', activate_file))
+end
+
+local function run_luarocks(args, disp)
```
tjdevries, line 132: I had dreams of importing luarocks here and running `require('luarocks.cmd').install(...)` or similar, but never panned out. Probably works better for you to use it via shell anyway tho so that it's async

tjdevries:
If you add an example to the readme how you expect it to work in this PR, I will give it a try.

lua/packer/luarocks.lua
wbthomason: Fair enough, the less overhead the better.

lua/packer/luarocks.lua
wbthomason: Noted. This seems to be working for me so far, but I'll take a closer look before merging. Does the current version of `neorocks` in `plenary` have the correct (as far as you know) path patterns?

lua/packer/luarocks.lua
wbthomason: Yeah, so did I. Unfortunately, the LuaRocks folks say there's no official way to do that, and I haven't found an unofficial one (though I haven't exactly looked hard; like you say, being able to run it async is critical)

lua/packer/luarocks.lua
wbthomason: Heh. Just after I wrote that, I found a bug in one of the paths.

wbthomason:
A couple of current known issues:
- [ ] Failure is not reported on the final results display
- [ ] You can only have a single version of a given rock installed at once
- [ ] Initial installation of LuaRocks does not chain correctly into installing the requisite packages; you have to run `PackerUpdate` or similar again

All but maybe the second of these needs to be fixed before merging.

wbthomason:
@tjdevries (or any other brave souls) the README in this PR now includes a simple example of how to specify rocks for a plugin.

nanotee:
I've tested this briefly with `lua-cjson`, `fun` and `penlight`. All three seemed to work perfectly, very exciting stuff! :)
I did run into a small bug:

```lua

packer.startup(function(use)
    use 'wbthomason/packer.nvim'
    use {'repo/plugin', rocks = 'lua-cjson'}
end)
```
```
[packer] Error in coroutine: ...te/pack/packer/start/packer.nvim/lua/packer/luarocks.lua:234: bad argument #1 to 'ipairs' (table expected, got string)
```

Looks like `clean_packages()` doesn't check the type of `plugin.rocks` before feeding it to `ipairs()`:

https://github.com/wbthomason/packer.nvim/blob/3e0eab82854cbfe91046f9af947509e7e8ace1b8/lua/packer/luarocks.lua#L233-L234

wbthomason:
Ok! I think this is ready to merge (at long last; sorry for how little time I had to move this along).

It currently supports:
- Specifying packages, optionally with versions, via the new `rocks` key
- Installing missing packages and removing unused packages
- Installing (via `hererocks`) `Luarocks` for users
- Updating `package.path` to make installed rocks discoverable
- The `PackerRocks` command, which can install or remove packages separate from use of `packer`
- Displaying the results of installing/removing rocks in the final results views of `packer.install`, `packer.update`, and `packer.sync`

It does **not** support (with context):
- Installing multiple versions of a package. I'm not sure how to handle this with `luarocks` in general, and it seems as though - because there is only one Lua environment for all of Neovim - having multiple package versions could cause name conflicts. If I'm wrong about this/if someone has ideas about how to handle it, I'd be happy to hear them!
- Semantic versioning of packages/anything other than equality constraints on versions. Again, this is just me not being sure how/if `luarocks` handles more sophisticated version constraints on the command line.
- Installing `rocks` from a file, rather than the manual specification. I still intend to add this, but I'm trying to decide on the right format. I don't want to reinvent any wheels, but `rockspec` feels like a relatively poor fit because it would require some parsing for `packer` to detect which rocks are still required and which need removal. I'm leaning toward something like a `requirements.lua` file standard, both for specifying rocks a plugin depends on and other plugins the plugin depends on. Either way, this will be another PR. Thoughts and suggestions are welcome!
- Updating rocks. It would probably be helpful to add a flag to check for updates (which I think is just running `luarocks install` again, potentially redundantly)
- Handling conflicts from Luarocks, e.g. conflicting dependency requirements. This should come in a later PR
- Providing more detailed error information in the results display. This should come in a later PR.
- Probably a billion things I haven't realized are missing yet

I'm going to leave this for a day or two, in case any brave souls have a chance to test it. Assuming no significant bugs are discovered in that time, I'll go ahead and merge it in.

Next steps will be some general cleanup work for `packer`, then resolving the issues around `require` semantics per #87 and friends. File-based rocks specification should come as a part of this; the other issues will probably be handled afterward.

As always, PRs, issues, and comments are welcome!

Issue #106, opened 2020-12-02
nvim --headless +PackerCompile +PackerUpdate +qa generate errors


cempassi:
Hi,

I'm trying to use packer to run tests on plugins in Docker, but I get some errors when I run this command:
`nvim --headless +PackerCompile +PackerUpdate +qa`

There should be a way do check if the buffer update can happend?

wbthomason:
What errors do you get?

cempassi:
`Error executing vim.schedule lua callback: .../site/pack/packer/opt/packer.nvim/lua/packer/display.lua:130: Expected 4 arguments%`

I'm sorry I shoud have stated with that.

I think the error comes from the way packer updates the buffer to report progress?

wbthomason:
Ah, I see. I'll admit that `packer` isn't really designed to support headless operation, but this would be good to add. It's going to take some modification to `display.lua` to validate that the display window actually opened.

This is related to #63

cempassi:
I can try to implement and submit a PR for this specific case? I'm really a beginner in Lua but that doesn't seem to hard to just validate if the buffer/window is available?

wbthomason:
You'd think not - unfortunately, it seems like some of the existing checks for this in `display.lua` (search for `is_valid`) aren't cutting it, so I need to find a new solution. If you can find something that works and send a PR, that'd be great!

wbthomason:
Heh, as soon as I wrote that I figured out a better solution. Please try #107 if you get a chance - if that fixes the issue for you too, I'll merge it in.

Also, just as a note, you probably want `nvim --headless +PackerUpdate +PackerCompile +qa` or just `nvim --headless +PackerSync +qa` - it seems better (though not critical) to update before compiling loaders.

cempassi:
I tested the PR on my end, fixes my issue. Thanks!

fsouza:
@wbthomason this is awesome, thank you very much! :D One last thing for perfect headless support would be a non-interactive mode (for example, when deleting a plugin, the `OK to remove? [y/N]` prompt will make ). Is this something you'd take a PR for?

wbthomason:
Yes, I'd happily merge that in! That was requested in another issue (I forget where) and has been on my backlog for a while now.

PR #109, opened 2020-12-04
display: add a non_interactive config parameter

fsouza:
Following up on #106: the idea is to be able to run PackerSync and have
it cleanup stale plugins without asking for input (imagine the user is
in --headless mode).

Originally, I had a no_input parameter, but based on feedback I updated
the code to have a config parameter that modifies the behavior of the
display module.

### DIFF
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -35,6 +35,11 @@ local keymaps = {
 
 --- Utility function to prompt a user with a question in a floating window
 local function prompt_user(headline, body, callback)
+  if config.non_interactive then
+    callback(true)
+    return
+  end
+
   local buf = api.nvim_create_buf(false, true)
   local longest_line = 0
   for _, line in ipairs(body) do
@@ -81,7 +86,7 @@ local display = {}
 local display_mt = {
   --- Check if we have a valid display window
   valid_display = function(self)
-    return not in_headless and self and api.nvim_buf_is_valid(self.buf) and api.nvim_win_is_valid(self.win)
+    return self and self.interactive and api.nvim_buf_is_valid(self.buf) and api.nvim_win_is_valid(self.win)
   end,
   --- Update the text of the display buffer
   set_lines = function(self, start_idx, end_idx, lines)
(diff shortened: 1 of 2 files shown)


wbthomason:
Thanks! This looks good.

I think we can cover most use cases with this option, so I'm happy to merge unless there's more you think should be added. I may eventually want to expand `no_input` to a `no_display` type of option, which would imply `no_input` but also prevent the display window for update and install operations from displaying. This would be to serve both users who run headless operations as well as users who prefer not to have a display window (there's been at least one issue asking about this).

I think that's probably best in another PR, though. If you agree, I'll go ahead and merge. Thanks again!

fsouza:
@wbthomason ohh that's a good call, I can turn this into a `no_display` option later today if you prefer. I can do that before we merge, just to make sure we don't introduce an option then remove it. What do you think?

wbthomason:
If you're willing, that'd be great before we merge. Note that `no_display` is probably easiest implemented as part of the `display` config, since many functions expect a display object. So, I think it should be simplest to make `no_display` control the `display` module to return a display object that skips all operations.

This could probably be done easily by adding a check to avoid opening the window/buffer in the `display.open` function, then adding a check in the `valid_display` method.

If you don't get to this today, no worries. Thanks again!

fsouza:
Sorry @wbthomason didn't get to it yesterday, but will do today.

fsouza:
This should be good for review now @wbthomason!

I changed my config to use packer.nvim from my fork and was able to test PackerSync, PackerInstall, PackerClean and co. Let me know if you have any feedback or questions!

lua/packer/display.lua
```diff
@@ -81,7 +86,7 @@ local display = {}
 local display_mt = {
   --- Check if we have a valid display window
   valid_display = function(self)
-    return not in_headless and self and api.nvim_buf_is_valid(self.buf) and api.nvim_win_is_valid(self.win)
+    return self.interactive and self.win and api.nvim_win_is_valid(self.win)
```
wbthomason, line 89: Out of curiosity, why does this drop the check that `self.buf` is valid?

wbthomason (commented):
This looks generally good, thanks! My only question before merging is why the check that the current display buffer remains valid was dropped? We're _probably_ ok without checking that, but I'd prefer to err on the side of caution (I don't think the check has significant performance impacts, and it is possible that some other plugin could interfere with the display buffer).

lua/packer/display.lua
fsouza: It was an oversight, sorry about that. Let me fix it

fsouza: Ok pushed a fix! :D

wbthomason: Great, thank you!

wbthomason (approved):

PR #110, opened 2020-12-05
Test for Neovim 0.5 on packer_compiled

Iron-E:
When using older versions of Neovim alongside `packer`, `has('nvim')` isn't quite restrictive enough. We need to know for sure if we have 0.5.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -8,7 +8,7 @@ local config = nil
 local function cfg(_config) config = _config end
 
 local feature_guard = [[
-if !has('nvim')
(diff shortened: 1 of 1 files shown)


wbthomason:
Thanks, this is a good addition. I added a quick tweak to also print a warning message on an invalid version. If you're OK with that, I'll go ahead and merge.

wbthomason (approved):

Iron-E:
That looks like a good warning to issue. I agree!

wbthomason:
Great, thanks again!

PR #112, opened 2020-12-06
Restore cursor position after floating window is closed

### LINKED ISSUE #111: Restore cursor position in floating mode
Packer is awesome and thanks for making it. :heart: 

So, I enabled the floating mode, and now whenever I run `PackerInstall` or `PackerUpdate` the cursor always moves to the first buffer in the splits which is NERDTree.

What I want is `packer` to restore the cursor position to the last buffer/cursor position after the floating window is closed.

numToStr:
Fixes #111

### DIFF
--- a/lua/packer/util.lua
+++ b/lua/packer/util.lua
@@ -97,6 +97,7 @@ end
 
 -- Credit to @crs for the original function
 util.float = function()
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Looks good, thank you!

PR #119, opened 2020-12-10
init: make config.disable_commands work also with require'packer'.init()

bfredl:
currently `config.disable_commands` only work with `require'packer'.startup()` which seems like an arbitrary limitation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,8 @@ Neovim.
 7. [Status](#status)
 
 ## Notices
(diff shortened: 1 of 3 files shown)


wbthomason:
I'm OK with this, but I'll note that the original reason for restricting `disable_commands` to work with `startup` was the belief that most users who opted not to use `startup` did so because they needed more control over how and when `packer` was loaded, meaning that they would want to define their own commands (for instance, this is true of me).

This should be fine to change as long as we update the docs and add a notice (which I've done).

PR #121, opened 2020-12-13
Use native API calls rather than vim.{fn,cmd}

katsika:


wbthomason (approved):
Looks good, thank you!

PR #123, opened 2020-12-14
Fix #103: Check for a non-empty table before using math.max(unpack(...))

### LINKED ISSUE #103: Math.max changes in LuaJIT
So it appears that LuaJIT made some breaking changes to `math.max` and `math.min` (possibly in https://github.com/LuaJIT/LuaJIT/commit/03208c8162af9cc01ca76ee1676ca79e5abe9b60?) which lead to an
```
Error executing vim.schedule lua callback: .../packer.nvim/lua/packer/plugin_utils.lua:48: bad argument #1 to 'max' (number expected, got no value)
```
on `:PackerUpdate`.

Unfortunately, this change is pulled in when bumping LuaJIT to a commit required to get it building on ARM64 (specifically, Apple M1), see https://github.com/neovim/neovim/pull/12624. (You can see the problem in action with the latest [snapshot of VimR](https://github.com/qvacua/vimr/releases), which is built with that PR (even on Intel).) Not sure what the best course of action is here; maybe do nothing until that PR (or a similar one) is merged? Or just implement your own `max`/`min` in this line for the sake of compatibility?

wbthomason:
Handle changes to `math.max` (I think...) in new LuaJIT without breaking old versions by checking for an empty table before using `math.max(unpack(...))`

### DIFF
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -45,8 +45,14 @@ plugin_utils.helptags_stale = function(dir)
   vim.list_extend(txts, vim.fn.glob(util.join_paths(dir, '*.[a-z][a-z]x'), true, true))
   local tags = vim.fn.glob(util.join_paths(dir, 'tags'), true, true)
   vim.list_extend(tags, vim.fn.glob(util.join_paths(dir, 'tags-[a-z][a-z]'), true, true))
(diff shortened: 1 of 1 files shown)


clason:
Yep, seems to work fine! (Meaning no Lua errors, at least; can't test whether helptags updates work.)

wbthomason:
Great, thanks. I think this is a pretty safe change, so I'll go ahead and merge.

Issue #75, opened 2020-10-29
Add floating packer display


clason:
Echoing the following comment on #73, it'd be lovely to have this built-in so I could just set `open_fn = require'packer.util'.float`!

@sunjon

> Totally unrelated question @sunjon , but how did you manage to have a packer open in a popup. Thanks

the options have a parameter `open_fn` to specify a function that will create a window for Packers content:

```lua
local opts = {
  display = {
    open_fn = "call vimrc#floating_window()",
  }
}
```

with a function based on https://github.com/neovim/neovim/issues/9718#issuecomment-559573308

_Originally posted by @sunjon in https://github.com/wbthomason/packer.nvim/issues/73#issuecomment-717569491_

wbthomason:
@sunjon Would you be willing to share the floating window function you appear to use? It looks nice, and (if you're willing) I'd be happy to steal it for the use @clason suggests.

clason:
crs pasted the following Lua snippet in Gitter chat when I asked about this: https://dpaste.org/zghD
```lua
return {
  floatingCentered = function()
    local columns = vim.o.columns
    local rows = vim.o.lines
    local width = math.min(columns - 4, math.max(80, columns - 20))
    local height= math.min(rows - 4, math.max(20, rows - 10))
    local top = ((rows - height) / 2) - 1
    local left = (columns - width) / 2
    local opts = {
      relative = 'editor',
      row = top,
      col = left,
      width = width,
      height = height,
      style = 'minimal'
    }
​
    local buf = vim.api.nvim_create_buf(false, true)
    local win = vim.api.nvim_open_win(buf, true, opts)
    vim.api.nvim_win_set_option(win, 'winhl', 'Normal:NormalFloat')
    opts.row = opts.row + 1
    opts.height = opts.height - 2
    opts.col = opts.col + 2
    opts.width = opts.width - 4
    vim.api.nvim_open_win(vim.api.nvim_create_buf(false, true), true, opts)
    vim.cmd([[autocmd! BufWipeout <buffer> execute 'bw ']] .. buf)
  end
}
```

tjdevries:
This: `vim.cmd([[autocmd! BufWipeout <buffer> execute 'bw ']] .. buf)` can be `vim.api.nvim_buf_delete` now :)

wbthomason:
@clason I've added this. Please let me know if it's not working as expected!

[...]

PR #127, opened 2020-12-19
remove `bg_buf` and `bg_win` from floating window

numToStr:
As you can see in the screenshot, the current floating window looks broken. So, I changed the highlights, dimensions, and added borders to make it look nice. Need feedback on how it looks on different themes.

Before:
image

After:
image

Thoughts? @wbthomason

### DIFF
--- a/lua/packer/util.lua
+++ b/lua/packer/util.lua
@@ -99,19 +99,20 @@ end
 util.float = function()
   local last_win = vim.api.nvim_get_current_win()
   local last_pos = vim.api.nvim_win_get_cursor(last_win)
-  local columns, lines = vim.o.columns, vim.o.lines
-  local win_width = math.ceil(columns * 0.8)
-  local win_height = math.ceil(lines * 0.8 - 4)
-  local col = math.ceil((columns - win_width) * 0.5)
-  local row = math.ceil((lines - win_height) * 0.5 - 1)
+  local columns = vim.o.columns
+  local lines = vim.o.lines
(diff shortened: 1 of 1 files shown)


katsika:
It's not broken on my colorscheme. And not everyone likes borders. 

Would be nice if for example plenary provided configurable floating windows and Packer reused that code rather than adding window management to the plug-in which should have nothing to do with it.

wbthomason:
@numToStr Thanks for the PR! This is good work.

I'd be happy to merge the bounds fixes, but I would like to keep the `open_fn` as minimal as possible - per the original request that led to it being added (#75), this is meant to be a minimal "it just works" floating display. I think I would actually prefer to remove the `bg_buf` entirely, rather than adding a border. What do you think?

@katsika: It would indeed be nice if somewhere provided configurable floating windows; unfortunately, unless we want to depend on every `packer` user having `plenary` (which I do not want to depend on), this would need to come from Neovim core. For the time being, I'm just trying to keep the floating window code simple and minimal.

numToStr:
> I think I would actually prefer to remove the `bg_buf` entirely, rather than adding a border. What do you think?

Yeah, this makes perfect sense to me. Would you mind if I the remove the `bg_buf` and update this PR?

wbthomason:
Sounds great, thank you!

numToStr:
@wbthomason I removed the `bg_win` and `bg_bug`. Now its ready to review.

wbthomason (approved):
Looks good, thank you!

PR #136, opened 2020-12-23
Make results window keybindings configurable

### LINKED ISSUE #34: Make results window keybindings configurable

nanotee:
Closes #34

This PR makes the keybindings of the display window configurable via `config.display.keybindings`

A few notes/questions:
- There is no validation process for the keybindings, users can input invalid mappings
- Should there be an option to disable the keybindings altogether?
- I can delete `keymap_display_order` if you think it makes the the code harder to read for no real benefit

### DIFF
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -350,7 +350,9 @@ q                    close the display window
 <CR>                 toggle information about a particular plugin
 r                    revert an update
 
-They can be configured by changing the value of `config.display.keybindings`.
+They can be configured by changing the value of `config.display.keybindings`
+(see |packer-configuration|). Setting it to `false` will disable all keybindings.
+Setting any of its keys to `false` will disable the corresponding keybinding.
 
 ==============================================================================
 API                                            *packer-api*
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -220,6 +220,11 @@ default values: >
       moved_sym = '→', -- The symbol for a plugin which was moved (e.g. from opt to start)
       header_sym = '━', -- The symbol for the header line in packer's display
       show_all_info = true, -- Should packer show all update details automatically?
+      keybindings = { -- Keybindings for the display window
+        quit = 'q',
+        toggle_info = '<CR>',
+        prompt_revert = 'r',
+      }
     }
   }
 
@@ -344,6 +349,8 @@ Once an operation completes, the results are shown in the display window.
 q                    close the display window
 <CR>                 toggle information about a particular plugin
 r                    revert an update
+
+They can be configured by changing the value of `config.display.keybindings`.
 
 ==============================================================================
 API                                            *packer-api*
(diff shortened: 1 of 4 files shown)


doc/packer.txt
```diff
@@ -345,6 +350,8 @@ q                    close the display window
 <CR>                 toggle information about a particular plugin
 r                    revert an update
 
+They can be configured by changing the value of `config.display.keybindings`.
```
wbthomason, line 353: Perhaps we should describe what values this config variable can take? Presumably either a table like in the example or (maybe) the value `false` to disable all keybindings?

lua/packer/display.lua
```diff
@@ -232,15 +246,10 @@ local display_mt = {
...
     if #raw_lines == 0 then table.insert(raw_lines, ' Everything already up to date!') end
 
-    local keymap_definitions = {
-      { lhs = keymaps[1][2], action = 'quit' },
-      { lhs = keymaps[2][2], action = 'show more info' },
-      { lhs = keymaps[3][2], action = 'revert an update' },
-    }
     table.insert(raw_lines, '')
-    for _, keymap in ipairs(keymap_definitions) do
+    for _, keymap in ipairs(keymap_display_order) do
```
wbthomason, line 250: If we add the ability to disable keybindings, then this needs to check that.

lua/packer/display.lua
```diff
@@ -409,6 +418,7 @@ end
 
 display.cfg = function(_config)
   config = _config.display
+  for name, lhs in pairs(config.keybindings) do keymaps[name].lhs = lhs end
```
wbthomason, line 421: Alternatively, the check for disabled keybinds could happen here.

lua/packer/display.lua
```diff
@@ -425,7 +435,7 @@ end
 --- Initialize options, settings, and keymaps for display windows
 local function setup_window(disp)
   api.nvim_buf_set_option(disp.buf, 'filetype', 'packer')
-  for _, m in ipairs(keymaps) do api.nvim_buf_set_keymap(disp.buf, m[1], m[2], m[3], m[4]) end
+  for _, m in pairs(keymaps) do api.nvim_buf_set_keymap(disp.buf, 'n', m.lhs, m.rhs, {nowait = true, silent = true}) end
```
wbthomason, line 438: Again, disable check needed here.

wbthomason (commented):
Thank you! This looks good overall. Regarding your questions/my comments:
1. I'm pretty OK allowing users to input invalid mappings. If there are some classes of invalid mapping that are easy to validate (e.g. using the same key for multiple purposes), then maybe we could add checks for those?
2. I think this is a good idea. We could add two changes:
    1. If `config.display.keybindings == false` or `nil`, then all keybindings are disabled
    2. If any value in `config.display.keybindings` is `false` or `nil`, then the corresponding keybinding is disabled.
What do you think? Is that adding too much complication?
3. I'm fine with `keymap_display_order`; I don't have a strong opinion on maintaining a consistent display order, but it doesn't complicate the code significantly.

Thanks again!

nanotee:
Thanks for the detailed comments!
I have implemented the checks needed for `false` to disable the keybindings (I think `nil` is equivalent to deleting a key so it gets overridden when user configuration gets merged with the default config)

wbthomason (approved):
Looks great, thanks!

PR #142, opened 2020-12-28
Use aliased name as plugin name if present

### LINKED ISSUE #140: Plugin aliased with `as` not distinguished from same plugin without `as`
In the following example, `as` does not make `packer` see the two `foo` as different plugins.

```lua
use {'upstream/foo', as='upstream-foo'}
use 'origin/foo'
```

[Evidence for this](https://github.com/wbthomason/packer.nvim/blob/master/lua/packer.lua#L130-#L143) was discovered by @akinsho on #118.

> I've had a look through the code and can see how it's happening now. Both the local and remote plugin get registered as `foo` since the path isn't used to make the plugin's name unique e.g. `my-username/foo`. As for using the `as` key this is actually not checked till after the plugin name check so the above error fails with `foo` has already been added.

Since the problem seems to be that the names are treated incorrectly (rather than changes to the state of the plugin spec not being detected) I created a separate issue.

Iron-E:
This is in reference to #140. Before, the `as` name was not used preferentially to the repo name provided in the specification. This will make the `as` alias be used instead, if present.

Closes #140

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -154,7 +154,7 @@ manage = function(plugin)
   end
 
   -- Handle aliases
-  plugin.short_name = plugin.as or name
+  plugin.short_name = name
   plugin.name = path
   plugin.path = path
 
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Looks good, thank you!

PR #143, opened 2020-12-28
Detect plugin.type changes

### LINKED ISSUE #118: No change detected when altering plugin URI from upstream repo to local path
Hello,

Thanks again for your continued work on this project. I've been enjoying using it lately!

There seems to be an issue when changing between local projects and upstream URLs. For example:

1. Specify an upstream plugin.
	```lua
	return require('packer').startup{function(use)
		use {'rust-lang/rust.vim', opt=true}
	end}
	```
2. Install the upstream plugin.
	```sh
	nvim +PackerCompile +PackerSync
	```
3. Re-specify the plugin to use a local path.
	```lua
	return require('packer').startup{function(use)
		use {fn.glob('~/Programming/ACTIVE/rust.vim'), opt=true}
	end}
	```
4. Install the local plugin.
	```sh
	nvim +PackerCompile +PackerSync
	```

The second `PackerSync` will report "already clean" and won't replace the currently downloaded plugin with a symlink to the directory specified.

Iron-E:
This PR adds necessary checks in order to determine whether or not a plugin type has changed during the most recent compile, and will install the plugin accordingly.

Todo:

* [ ] Alter logic of `:PackerClean` so that properly detected missing plugins have a directory to install to.
  * Currently, this PR can make `packer` detect that a plugin's type has changed, but then it will fail to reinstall because the directory will not have been cleaned. That is my next goal.

Topics left to discussion:

* [ ] Better way to get `packer` compiled plugin table than global function?

Closes #118

### DIFF
--- a/lua/packer/clean.lua
+++ b/lua/packer/clean.lua
@@ -9,6 +9,9 @@ local async = a.sync
 
 local config
 
+local PLUGIN_OPTIONAL_LIST = 1
+local PLUGIN_START_LIST = 2
+
 local function is_dirty(plugin, typ)
   return plugin.disable or (plugin.opt and typ == 2) or (not plugin.opt and typ == 1)
 end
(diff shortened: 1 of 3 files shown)


Iron-E:
I might try to develop a better method for checking the type of a plugin folder than to look at the last compile. Playing around with it makes me see that this approach works most of the time but not all of the time, maybe using `libuv` to actually analyze the type of directory it is rather than what the specification was last time.

This would also make it sensitive to changes that occur from outside `packer`, and we could remove the global function.

I'll play with it some more.

PR #144, opened 2020-12-29
Check dir type for changes

### LINKED ISSUE #118: No change detected when altering plugin URI from upstream repo to local path

Iron-E:
This is a followup to #143. It uses a more consistent method for checking the type of a directory than attempting to embed it into the `packer_compile`.

The following changes have been made:

* `plugin_utils.find_missing_plugins` now uses two metrics to determine whether or not a plugin is "missing":
    1. The plugin is not installed at all (directory is missing).
    2. The plugin is installed, but the `guess_dir_type` does not match `plugin.type`.
* `PackerClean` has been re-implemented in terms of `find_missing_plugins`. This is because some missing plugins are also plugins which need to be cleaned before a reinstall (i.e. moving from local to upstream and vice versa).
    * See the commit message for 51bf0962d450eebe7d123a41279810ee1724406c. Essentially it just uses more logic from other parts of the plugin.
* More strings and numbers have been converted to variables in case the values change in the future.

Closes #118.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -178,7 +178,7 @@ manage = function(plugin)
                                         plugin.short_name)
 
   if not plugin.type then plugin_utils.guess_type(plugin) end
-  if plugin.type ~= 'custom' then plugin_types[plugin.type].setup(plugin) end
+  if plugin.type ~= plugin_utils.custom_plugin then plugin_types[plugin.type].setup(plugin) end
   for k, v in pairs(plugin) do if handlers[k] then handlers[k](plugins, plugin, v) end end
   plugins[plugin.short_name] = plugin
 
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -178,7 +178,7 @@ manage = function(plugin)
                                         plugin.short_name)
 
   if not plugin.type then plugin_utils.guess_type(plugin) end
-  if plugin.type ~= plugin_utils.custom_plugin then plugin_types[plugin.type].setup(plugin) end
+  if plugin.type ~= plugin_utils.custom_plugin_type then plugin_types[plugin.type].setup(plugin) end
   for k, v in pairs(plugin) do if handlers[k] then handlers[k](plugins, plugin, v) end end
   plugins[plugin.short_name] = plugin
 
--- a/lua/packer/clean.lua
+++ b/lua/packer/clean.lua
@@ -49,14 +49,14 @@ local clean_plugins = function(_, plugins, results)
       local plugin_missing = missing_plugins[plugin_config.short_name] and vim.loop.fs_stat(path)
 
       if plugin_missing or is_dirty(plugin_config, plugin_source) then
-        dirty_plugins[#dirty_plugins+1] = path
+        table.insert(dirty_plugins, path)
       end
     end
 
     -- Any path which was not set to `nil` above will be set to dirty here
     local function mark_remaining_as_dirty(plugin_list)
       for path, _ in pairs(plugin_list) do
-        dirty_plugins[#dirty_plugins+1] = path
+        table.insert(dirty_plugins, path)
       end
     end
 
--- a/lua/packer/clean.lua
+++ b/lua/packer/clean.lua
@@ -19,61 +19,65 @@ end
 -- Find and remove any plugins not currently configured for use
 local clean_plugins = function(_, plugins, results)
   return async(function()
+    local dirty_plugins = {}
     results = results or {}
     results.removals = results.removals or {}
 
-    local function map_install_folder(install_paths)
-      local install_folders = {}
-
-      for install_path, _ in pairs(install_paths) do
-        local split_path = vim.split(install_path, '/', true)
-        install_folders[split_path[#split_path]] = true
-      end
-
-      return install_folders
-    end
-
-    local function map_names_to_paths(plugin_names)
-      local paths = {}
-
-      for _, plugin_name in ipairs(plugin_names) do
-        paths[#paths+1] = plugins[plugin_name].install_path
-      end
-
-      return paths
-    end
-
     local opt_plugins, start_plugins = plugin_utils.list_installed_plugins()
-    local dirty_plugins = plugin_utils.find_missing_plugins(plugins, opt_plugins, start_plugins)
 
-    opt_plugins = map_install_folder(opt_plugins)
-    start_plugins = map_install_folder(start_plugins)
+    local missing_plugins = plugin_utils.find_missing_plugins(plugins, opt_plugins, start_plugins)
+    -- turn the list into a hashset-like structure
+    for idx, plugin_name in ipairs(missing_plugins) do
+      missing_plugins[plugin_name] = true
+      missing_plugins[idx] = nil
+    end
 
+	 -- test for dirty / 'missing' plugins
     for _, plugin_config in pairs(plugins) do
-      local plugin_name = plugin_config.short_name
-      local plugin_source = (opt_plugins[plugin_name] and PLUGIN_OPTIONAL_LIST) or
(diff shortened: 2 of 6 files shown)


lua/packer/clean.lua
```diff
@@ -9,44 +9,71 @@ local async = a.sync
...
+
+    opt_plugins = map_install_folder(opt_plugins)
+    start_plugins = map_install_folder(start_plugins)
+
+    for _, plugin_config in pairs(plugins) do
+      local plugin_name = plugin_config.short_name
+      local plugin_source = (opt_plugins[plugin_name] and PLUGIN_OPTIONAL_LIST) or
+        (start_plugins[plugin_name] and PLUGIN_START_LIST)
+
+      if is_dirty(plugin_config, plugin_source) then
```
Iron-E, line 57: I have to add `if not plugin_source` here as well.

lua/packer/clean.lua
```diff
@@ -9,44 +9,71 @@ local async = a.sync
...
+
+    local opt_plugins, start_plugins = plugin_utils.list_installed_plugins()
+    local dirty_plugins = plugin_utils.find_missing_plugins(plugins, opt_plugins, start_plugins)
+
+    opt_plugins = map_install_folder(opt_plugins)
+    start_plugins = map_install_folder(start_plugins)
+
+    for _, plugin_config in pairs(plugins) do
+      local plugin_name = plugin_config.short_name
+      local plugin_source = (opt_plugins[plugin_name] and PLUGIN_OPTIONAL_LIST) or
```
Iron-E, line 54: I'm going to add something here that removes detected plugins from the list, so that anything left over must not have been in the plugin spec and should be removed.

lua/packer.lua
```diff
@@ -178,7 +178,7 @@ manage = function(plugin)
                                         plugin.short_name)
 
   if not plugin.type then plugin_utils.guess_type(plugin) end
-  if plugin.type ~= 'custom' then plugin_types[plugin.type].setup(plugin) end
+  if plugin.type ~= plugin_utils.custom_plugin then plugin_types[plugin.type].setup(plugin) end
```
wbthomason, line 181: Nitpick: I'd request that `custom_plugin` be named `custom_plugin_type`.

lua/packer/clean.lua
```diff
@@ -9,44 +9,75 @@ local async = a.sync
...
+      elseif start_plugins[path] then
+        plugin_source = PLUGIN_START_LIST
+        start_plugins[path] = nil
+      end
+
+		-- We don't want to report paths which don't exist for removal; that will confuse people
+      local plugin_missing = missing_plugins[plugin_config.short_name] and vim.loop.fs_stat(path)
+
+      if plugin_missing or is_dirty(plugin_config, plugin_source) then
+        dirty_plugins[#dirty_plugins+1] = path
```
wbthomason, line 52: Nitpick: Here and at line 59, is there a reason this isn't `table.insert(dirty_plugins, path)`?

lua/packer/display.lua
```diff
@@ -221,7 +222,7 @@ local display_mt = {
         local actual_update = true
         local failed_update = false
         if result.ok then
-          if plugin.type ~= 'git' or plugin.revs[1] == plugin.revs[2] then
+          if plugin.type ~= plugin_utils.git_plugin or plugin.revs[1] == plugin.revs[2] then
```
wbthomason, line 225: Nitpick: Would request that `git_plugin` be called `git_plugin_type`.

lua/packer/plugin_utils.lua
```diff
@@ -8,20 +8,41 @@ local config = nil
 local plugin_utils = {}
 plugin_utils.cfg = function(_config) config = _config end
 
+plugin_utils.custom_plugin = 'custom'
+plugin_utils.local_plugin = 'local'
```
wbthomason, line 12: Nitpick: would request that `local_plugin` be called `local_plugin_type`.

wbthomason (changes requested):
Thanks! This looks like a significant improvement. I've left a few requests for minor changes (mostly just variable naming), but will be happy to merge.

We will eventually need to work this into `packer.install` and `packer.update` too, since not everyone runs `packer.sync` to ensure that the `clean` operation happens first (and - correct me if I'm wrong - the clean operation **must** run first for directory type changes to be picked up). Perhaps this is better in a different PR?

....to be honest, I'm now starting to wonder if it's better to just remove the `PackerInstall` and `PackerUpdate` commands entirely in lieu of `PackerSync` as the one-stop shop. What do you think? This change would not go in this PR, but thinking about the necessary changes to `install` and `update` has me feeling that there's some redundancy which could be removed there.

lua/packer/clean.lua
Iron-E: I usually try to use the insertion syntax where possible, because it's just one less resolution for Lua to go through (see [here](https://stackoverflow.com/questions/38345582/why-is-tt-1-e-faster-than-table-insertt-e)). I did notice that `table.insert` is preferred by a lot of the Neovim community, though, and I don't mind changing this— it was just my instinct.

Iron-E:
I made the requested changes on the `xxx_plugin` variables. I see though that there is a `plugin_types` file— should these variables go there or stay in `plugin_utils`?

lua/packer/clean.lua
wbthomason: Oh, cool. I didn't know that - intuitively, I thought that the insertion syntax would trigger metamethods, whereas `table.insert` would not, so I actually thought `table.insert` would be faster. Good to learn, thanks! I'm fine with using the insertion syntax (or not if you don't feel like reverting the change to `table.insert`).

cloggier:
Only downside in removing those commands is that when wanting to install new plugin or clean old one Sync would also run update.

Iron-E:
Maybe `:PackerInstall` should always clean before installation. The user could choose to either clean if a change is made to the plugin type, but if the user doesn't want to (or if false positives occur) they can elect not to.

Then `:PackerSync` would just be `:PackerInstall` followed by `:PackerUpdate`, and the cleaning would happen during the installation routine.

wbthomason:
> :PackerInstall also sees that a directory change is present but won't actually clean the existing path before attempting to install. This of course means that the :PackerInstall routine will yield failures when directory type changes are detected, since a cleaning is needed in that case.

Right - sorry, I wasn't very clear in my initial response; this is what I meant w.r.t. working the changes into `install`. This will probably also be an issue for `update` for the same reason.

wbthomason (approved):
This looks great, thank you! I think we should eventually refactor the `X_plugin_type` variables to `plugin_types`, but for this PR, I'm content for them to live in `plugin_utils`. I'm good to merge if you are. Thanks again!

lua/packer/clean.lua
Iron-E: I'll leave it to match the style in the rest of the project, though we can create another performance PR if we wanted to use that syntax for appending.

Iron-E:
Looks good to me.

Glad I could help with this, thanks for reviewing!

PR #145, opened 2020-12-29
Fix jump on operator lazy-loading

### LINKED ISSUE #115: Weird jump on operator lazy-loading
Hi, first, thanks for the amazing plugin! Second, I noticed that if I try to lazy-load any operator plugin, like vim-surround, vim-commentary, replacewithregister, the moment packer loads it, cursor jumps to the beginning of the line.
Peek 2020-12-08 21-04

nanotee:
Fixes #115

Just a small additional check for the value of `vim.v.count`, otherwise it's always truthy

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -115,7 +115,7 @@ _packer_load = function(names, cause)
     end
 
     if cause.prefix then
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Ah, nice catch! Thank you!

PR #146, opened 2021-01-03
Made packer work without luajit

shadmansaleh:
- Added a secondary os detection when jit is not available
  based on package.config

### DIFF
--- a/lua/packer/util.lua
+++ b/lua/packer/util.lua
@@ -31,7 +31,14 @@ util.nonempty_or = function(opt, alt)
   end
 end
 
-util.is_windows = jit.os == 'Windows'
+util.is_windows = (function()
+  if jit ~= nil then
+    return jit.os == 'Windows'
+  else
+    -- detect os based on configuaration flags
+    return package.config:sub(1,1) == '\\'
+  end
+end)()
 
 util.get_separator = function()
   if util.is_windows then return '\\' end
--- a/lua/packer/util.lua
+++ b/lua/packer/util.lua
@@ -31,14 +31,7 @@ util.nonempty_or = function(opt, alt)
   end
 end
 
-util.is_windows = (function()
-  if jit ~= nil then
-    return jit.os == 'Windows'
-  else
-    -- detect os based on configuaration flags
-    return package.config:sub(1,1) == '\\'
-  end
-end)()
+util.is_windows = package.config:sub(1,1) == '\\'
 
 util.get_separator = function()
   if util.is_windows then return '\\' end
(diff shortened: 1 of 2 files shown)


lua/packer/util.lua
```diff
@@ -31,7 +31,14 @@ util.nonempty_or = function(opt, alt)
   end
 end
 
-util.is_windows = jit.os == 'Windows'
+util.is_windows = (function()
```
wbthomason, line 34: Nitpick: does this need to be a function? Is there a reason to not to do this check at the top level?

wbthomason (commented):
Thanks! I have one nitpick about the style, but this looks good otherwise.

lua/packer/util.lua
shadmansaleh: > Is there a reason to not to do this check at the top level?
Sorry I couldn't understand what you ment by top lavel .
We could just use
`util.is_windows = package.config:sub(1,1) == '\\'`
Do you want me to change it to that ?

wbthomason:
Hmmm. Not sure why, but I'm not seeing my additional commit here. What I was suggesting was to use the following snippet:

```lua
if jit ~= nil then
  util.is_windows = jit.os == 'Windows'
else
  util.is_windows = package.config:sub(1,1) == '\\'
end
```

wbthomason (approved):
I'm good to merge this if you are, @shadmansaleh

shadmansaleh:
I've removed LuaJit requirement from README.md

Actually I was using neovim-ppa on ubuntu for a while. For some reason it's compiled with lua5.1 . Also neovim in termux is compiled with lua5.3 . those are two I know of.

I donn't think there is any reason to depend on luajit for one line of code.

I've tested package.config:sub(1,1) == '\\' this to work on luajit,lua5.1,lua5.2,lua5.3 .
So i'm pretty confident it wonn't cause any problems. So you can merge it whenever you want to.

shadmansaleh:
> I'm not certain that I don't use other LuaJIT features elsewhere in packer

I've used packer on neovim compiled with lua5.1 
with this branch for a few days . Nothing seems to break installing, loading, uninstalling all seems to work without any errors . So I donn't think theres anything that depends on luajit. atlest not on master.

wbthomason:
Good to know, thanks. I reserve the right to depend on LuaJIT properties or functionality in the future (in particular, I have some ideas around using the `ffi` module), but I'm happy to merge this for now to make things as broadly compatible as they can be. Thanks again!

wbthomason (approved):

Issue #131, opened 2020-12-21
Help wanted - Solution approach for lazy load via function call


weilbith:
Hey :wave: 
I currently have the need to load a package lazy when a function or `<Plug>` gets called. This should work similar to the lazy loading via commands. So I started to experiment and try if my solution approach works. So I tried it for a specific example but kept it generic enough to later make it a new feature for this plugin here. Since I have no clue how `<Plug>` does work in Vim, I'm focusing on the function call. 

So here I add a plugin to packer and make it just optional:

```lua
require('packer').use({ 'preservim/nerdcommenter', opt = true })
```

And then I have a `plugin/comment_mappings.vim` file with the following content:

```vim
function s:load(package, function, arguments) abort
  execute 'delfunction ' . a:function
  packadd a:package
  call call(function(a:function), a:arguments)
endfunction

function g:NERDComment(...) abort
  call s:load('vim-nerdcommenter', 'NERDComment', a:000)
endfunction

nnoremap <silent> <leader>cl <cmd>call NERDComment('n', 'Comment')<CR>
```

Note that I used `s:load` as this is the naming in this plugin. The `g:NERDComment` function would be what the compiler module needs to generate later. The `package` and `function` arguments will change as you maintain such information in a global dict, if I'm not mistaken.

Well, after all this approach doesn't work. When I `call s:load` from `g:NERDComment`, the latter function is still "active". This means the `delfunction` command fails with the error: `E131: Cannot delete function NERDComment: It is in use`.
Makes sense...

So I'm asking you and everyone who is watching the repository: do you have any other ideas how to implement such a feature? The idea is to later have a packer plugin attribute called `functions`, which works the same as the already existing `cmd` attribute.

For my currently "productive" NeoVim setup I still use `dein.nvim` which has this feature of lazy loading via functions. I should eventually take a look there how Shougo is doing it.

PS: I know that functions can also make use of ranges as commands do with the `range` keyword. I haven't used this feature yet and also haven't thought about how to support it here. To be 100% save we should think about that issue as well. Just for notice.

weilbith:
Would an approach using Lua jobs work? So that we start a job from the "fake function" with a delay, which makes this function be finished when the job function deletes it?

weilbith:
I tried to use Vim timer, but it throws the same error:

```vim
function g:NERDComment(...) abort
  let load_call = call(function('s:load'), ['vim-nerdcommenter', 'NERDComment', a:000])
[...]

PR #151, opened 2021-01-07
Feature/lazy load via undef function

weilbith:
Fixes: #131 

I usually prefer to have full names in code. But since `function` (although not its plural) is a reserved word in Lua, I decided to make an exception and name it `func` (and `funcs` respectively).

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,7 @@ A table given to `use` can take two forms:
    and may additionally have a number of optional keyword elements, shown below:
 ```lua
 use {
-  'myusername/example',   -- The plugin location string
+  'myusername/example',        -- The plugin location string
   -- The following keys are all optional
   disable = boolean,           -- Mark a plugin as inactive
   as = string,                 -- Specifies an alias under which to install the plugin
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -275,13 +275,14 @@ use {
...
   run = string or function,    -- Post-update/install hook. See "update/install hooks".
-  requires = string or list -- Specifies plugin dependencies. See "dependencies".
+  requires = string or list    -- Specifies plugin dependencies. See "dependencies".
   config = string or function, -- Specifies code to run after this plugin is loaded.
   -- The following keys all imply lazy-loading
   cmd = string or list,        -- Specifies commands which load this plugin.
   ft = string or list,         -- Specifies filetypes which load this plugin.
   keys = string or list,       -- Specifies maps which load this plugin. See "Keybindings".
   event = string or list,      -- Specifies autocommand events which load this plugin.
+  func = string or list        -- Specifies functions which load this plugin.
```
wbthomason, line 285: Nitpick: What if we call this keyword `fn` instead? It's slightly more concise, and common to some programming languages (e.g. Rust). This latter property might be true of `func` too, but I'm not sure.

wbthomason (changes requested):
Without having tested it myself yet, this looks great, thanks!

I did have one nitpicky request: how do you feel about calling the key `fn` rather than `func`? I don't have a strong preference here, but slightly favor `fn` for its conciseness and (relative) familiarity from other programming language syntaxes.

I also have a question that's likely one for the core devs: how does `FuncUndefined` interact with Lua functions? I assume it does not trigger for them, since this would require a fairly significant amount of hooking in to Lua's call resolution mechanism (I think), but this would be good to know (and document).

README.md
weilbith: Sure I can change that. I have no hard feelings about it as I stated in the PR summary.
Asap I'm sitting at a PC again I'll change it.

weilbith:
Haha. Despite testing it manually, there are no unit/integration tests yet, aren't there? :thinking:

weilbith:
@wbthomason if you are searching for an example to test this, I have it setup for NERDCommenter and vim-repeat. :shrug:

README.md
weilbith: Done :heavy_check_mark:

wbthomason:
> Haha. Despite testing it manually, there are no unit/integration tests yet, aren't there?

Yeaahhh... testing is an admitted weak point of `packer` :grimacing:

wbthomason (approved):
Looks good, thank you!

PR #153, opened 2021-01-08
Fix escaping of keys with <Plug> for lazy loading

weilbith:
This is still not working 100%. While the created mapping in the compiled script does not throw an error anymore and successfully loads the packages, the feedkeys do not have an affect. The next time the keys are typed they take effect. I think must be related to [this here](https://github.com/wbthomason/packer.nvim/blob/f62f2fc1fb1802a217afdbac6a487e541019f658/lua/packer/compile.lua#L133). According to [the docs](https://neovim.io/doc/user/eval.html#feedkeys()) this requires to use double quotes to make the special keys working. But I don't know how to achieve that. The [nvim_feedkeys](https://neovim.io/doc/user/api.html#nvim_feedkeys()) API function allows to specify a parameter to "trigger" the auto-espace of special chars. But it still does not work for me. Does someone has an idea?

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -128,9 +128,10 @@ _packer_load = function(names, cause)
       vim.fn.feedkeys(prefix, 'n')
     end
 
-    -- NOTE: I'm not sure if the below substitution is correct; it might correspond to the literal
-    -- characters \<Plug> rather than the special <Plug> key.
-    vim.fn.feedkeys(string.gsub(string.gsub(cause.keys, '^<Plug>', '\\<Plug>') .. extra, '<[cC][rR]>', '\r'))
+    local formatted_plug_key = string.format('%c%c%c', 0x80, 253, 83)
+    local keys = string.gsub(cause.keys, '^<Plug>', formatted_plug_key) .. extra
+    local escaped_keys = string.gsub(keys, '<[cC][rR]>', '\r')
+    vim.fn.feedkeys(escaped_keys)
   elseif cause.event then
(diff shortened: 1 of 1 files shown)


wbthomason:
Just to confirm, is the problem that you're not sure how to get double quotes in the parameter to `feedkeys` or that even with double quotes in the call to `feedkeys`/the `escape_csi` parameter set when using `nvim_feedkeys`, you get the behavior that the keys don't trigger on the first press?

Also, we should probably use `nvim_feedkeys` rather than `vim.fn.feedkeys`, if possible - I'm trying to gradually convert to using API functions where possible throughout the codebase.

weilbith:
> Just to confirm, is the problem that you're not sure how to get double quotes in the parameter to feedkeys or that even with double quotes in the call to feedkeys/the escape_csi parameter set when using nvim_feedkeys, you get the behavior that the keys don't trigger on the first press?

Both? ^^
I don't know for sure how to add the double quotes. I tried to surround the parameter with something like `'"' .. <all-the-sub-parts> .. '"'`. But that did not workout.

> Just to confirm, is the problem that you're not sure how to get double quotes in the parameter to feedkeys or that even with double quotes in the call to feedkeys/the escape_csi parameter set when using nvim_feedkeys, you get the behavior that the keys don't trigger on the first press?

The issue with that function was that the named parameter is the third. This requires us to define the second as well (if I'm not mistaken), which is the mode. Can we safely use the currently active mode for that? :thinking:

weilbith:
Alright. I found our how to do it. It is working now.

wbthomason (approved):
Nice solve! Looks good to me. Thanks again!

PR #157, opened 2021-01-12
ref: clean before :PackerInstall/Update

Iron-E:
This is a followup to #144. We discussed that `:PackerInstall` and `:PackerUpdate` needed to clean before peforming their operations, since directory changes may have occurred. In order to prevent an increase to the complexity of `:PackerSync`, this PR implements changes to the high-level Vim commands `:PackerInstall` and `:PackerUpdate` in order to perform the clean.

* This means that `packer.install()` and `packer.update()` do not clean automatically, but their respective commands do.

I'm happy to discuss this solution as well as potential alternatives.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -246,9 +246,9 @@ packer.install = function(...)
   end
 
   async(function()
-    vim.cmd 'PackerClean'
     local start_time = vim.fn.reltime()
     local results = {}
+    await(clean(plugins, results))
     local tasks, display_win = install(plugins, install_plugins, results)
     if next(tasks) then
       table.insert(tasks, 1, function() return not display.status.running end)
@@ -280,9 +280,9 @@ end
 packer.update = function(...)
   local update_plugins = args_or_all(...)
   async(function()
-    vim.cmd 'PackerClean'
     local start_time = vim.fn.reltime()
     local results = {}
+    await(clean(plugins, results))
     local missing_plugins, installed_plugins = util.partition(
                                                  plugin_utils.find_missing_plugins(plugins),
                                                  update_plugins)
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -100,8 +100,8 @@ packer.init = function(user_config)
   plugin_utils.ensure_dirs()
 
   if not config.disable_commands then
-    vim.cmd [[command! PackerInstall  lua require('packer').install()]]
-    vim.cmd [[command! PackerUpdate   lua require('packer').update()]]
+    vim.cmd [[command! PackerInstall  lua local packer = require 'packer'; packer.clean(); packer.install()]]
+    vim.cmd [[command! PackerUpdate   lua local packer = require 'packer'; packer.clean(); packer.update()]]
     vim.cmd [[command! PackerSync     lua require('packer').sync()]]
     vim.cmd [[command! PackerClean    lua require('packer').clean()]]
     vim.cmd [[command! PackerCompile  lua require('packer').compile()]]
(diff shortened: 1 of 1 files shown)


wbthomason (commented):
Thanks for this! I think I would prefer to add this call within `packer.install()` and `packer.update()` - not everyone uses the default commands, and I think it's cleaner to have all the logic in one place. What do you think?

Iron-E:
You're definitely right. My main concern was just repeating a clean twice for every `:PackerSync`, maybe `packer.install()` or `packer.update()` should accept a `should_clean` parameter so that if `:PackerSync` is run it can tell `packer.update` that it doesn't need to worry about cleaning, since a `packer.install` with cleaning enabled was already run.

Iron-E:
Sorry for the delay. Been busy with another project and some irl stuff, but I am working on this. As soon as I test this latest round of commits I can push it.

Can't get to it today, but definitely this week.

wbthomason:
Thanks, and absolutely no rush. I'm a bit slammed with real work right now myself.

lua/packer.lua
```diff
@@ -246,6 +246,7 @@ packer.install = function(...)
   end
 
   async(function()
+    vim.cmd 'PackerClean'
```
Iron-E, line 249: This is a problem line. I've tried a lot of things and ultimately this is the only solution I could come across.

Here's what I _think_ is happening: whenever I replace `vim.cmd` with a direct call to `packer.clean()`, it uses an `async()` method. This means that the clean will occur after the install function tries to install the plugin, meaning that the directory has not been cleaned and the install always fails. I believe `vim.cmd` syncs the commands here.

I have tried a few things to get it to sync up without using a vim command:

1. `a.wait(packer.clean())`, though this causes only the clean to be performed and the install instructions thereafter are dropped. I'm not sure why this happens, although I'm guessing this has something to do with the Lua implementation of `async` not working exactly as one would expect async to work (since it is built on top of loosely scheduled coroutines, from what I can tell).
2. I have tried moving the `packer.clean()` call from out of this `async(function()` scope, but there's no difference to be made there.
3. I accidentally tried `async(packer.clean())`, and interestingly it gave the same result as `await(packer.clean())`

Maybe I'm missing something with the async implementation? Or maybe I'm barking up the wrong tree altogether. I feel like there's something obvious here but I can't quite see it…

Iron-E (commented):
I have it working but there's a hiccup— right now it depends on a call to `vim.cmd` because of some strange syncing that's happening. I've left a comment on the relevant lines and what I've tried to make it work, as well as what I think might be happening.

If you could take a brief look and see what the problem might be, I'd be appreciative. However, if you're perfectly happy with running `vim.cmd` in this case we can just forget about it and merge anyway (since that seems to be working just fine).

lua/packer.lua
wbthomason: Ah, yes - sorry, the async framing has a bit of a learning curve (though I assert it makes things easier to write in net).

What you probably want here is either `await(clean(plugins, results))` or changing `packer.clean` to return an `async` function and using `await(packer.clean())`. The issue right now is that `packer.clean()` simply creates and immediately runs an `async` function (necessary because the `clean` function uses async operations and you cannot use `coroutine.yield` across a C function call), and returns nothing.

Iron-E:
Thanks for the help! Using the code you suggested worked like a charm, and it even shows a summary at the end now with what was cleaned.

wbthomason:
Great, thank you so much for this!

Iron-E:
Team effort ;)

PR #158, opened 2021-01-12
Reload required config/setup modules before packer compile

### LINKED ISSUE #152: Packer compile does not pick up require-d config changes
When using packer with dependency configs specified via `require`
e.g. `use {"myplugin", config = require("path.settings")}`, updating the file being required isn't picked up till packer compile is re-run multiple times.

Understandably the changes won't be seen till `packer_compiled.vim` is sourced but the actual file doesn't reflect the newly added change till it is rerun. So the ideal flow would be
1. make change in "path.settings"
2. Run `PackerCompile`
3. Restart neovim (ideally not even this since it's disruptive but 1 problem at a time 😄 )

I've narrowed down the issue to require's behaviour i.e. it caching the result of the first require so subsequent calls to compile don't see that the file is different. So the way to get this to work is currently.
1. Make a settings change
2. Restart neovim or if you know to do this set the settings package to be unloaded and re require it
3. Start neovim run packer compile
4. (if you want the setting to actually be active) restart neovim (or potentially re-source the compiled vim file)

Ideally I wonder if configs supplied like this (a pattern I've seen in quite a few dotfiles) can either work this way i.e. I make a change run a command and packer picks up that it needs to recompile with these changes, seems like in this case packer will have to unload a bunch of config modules for the user 🤔 😓 . Not sure how easy that would be. Or a built in mechanism specifically for specifying config options for plugins that require reams of it (that might be tidier kept elsewhere) that would take a path and manage requiring it and reloading it every time compile is run so every run picks up the changes.

e.g.
```lua
use {'myplugin', config_path = "myplugin.settings"}
-- In packer somewhere
require(plugin.config_path)

-- During compile
if config_path then
  packages[plugins.config_path].loaded = nil
  config = require(plugins.config_path)
  config_to_dump_as_str = config
end
```

This is all only thought halfway through, just a sketch of a possible solution, alternatively maybe the solution would be as simple as something like `dofile` but trying that causes packer to error. I'm not super familiar with how that function differs from `require` except that I'd assumed it wouldn't cache values but would run the code and return the value

akinsho:
This PR will fix #152 by automagically reloading any required configs in the packer startup call when compile is called.

@wbthomason this largely works as discussed except for one big caveat which is that the file where the user's plugins are specified needs to be reloaded (using luafile) before the change is picked up. From what I can tell the `precompile` (can change any names or relocate any functions) function correctly reloads the module but `compile` doesn't seem to actually rerun once packer is started up already i.e. it doesn't seem able to pick up any changes itself.

### Demo
You can't see my key presses unfortunately (don't have any good software for that), but I run packer compile then switch window

packer-reload

~I tried reloading the module where the user called `packer.startup` using `debug.getinfo()` to find the module name but that doesn't work and also seems brittle. Open to any suggestions~

## TODO
- [x] reload `setup` as well

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -381,7 +381,7 @@ end
 --- return a function, then cross reference them with all
 --- the plugin configs and setups and if there are any matches
 --- reload the user module.
-local function precompile(plugs)
+local function refresh_configs(plugs)
   local reverse_index = {}
   for k, v in pairs(package.loaded) do
     if type(v) == "function" then
@@ -400,7 +400,7 @@ end
 -- Takes an optional argument of a path to which to output the resulting compiled code
 packer.compile = function(output_path)
   output_path = output_path or config.compile_path
-  precompile(plugins)
+  refresh_configs(plugins)
   -- NOTE: we copy the plugins table so the in memory value is not mutated during compilation
   local compiled_loader = compile(vim.deepcopy(plugins))
   output_path = vim.fn.expand(output_path)
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -369,10 +369,29 @@ packer.sync = function(...)
   end)()
 end
 
+local function precompile(plugs)
+  local reverse_index = {}
+  for k, v in pairs(package.loaded) do
+    if type(v) == "function" then
+        reverse_index[v] = k
+    end
+  end
+  -- TODO also reload setup as well as config
+  for _, plugin in pairs(plugs) do
+    local value = plugin.config
+    local package_name = reverse_index[value]
+    if package_name then
+      package.loaded[package_name] = nil
+      require(package_name)
+    end
+  end
+end
+
 --- Update the compiled lazy-loader code
 -- Takes an optional argument of a path to which to output the resulting compiled code
 packer.compile = function(output_path)
   output_path = output_path or config.compile_path
+  precompile(plugins)
   local compiled_loader = compile(plugins)
   output_path = vim.fn.expand(output_path)
   vim.fn.mkdir(vim.fn.fnamemodify(output_path, ":h"), 'p')
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -369,6 +369,18 @@ packer.sync = function(...)
   end)()
 end
 
+
+local function reload_module(name)
+  if name then
+    package.loaded[name] = nil
+    return require(name)
+  end
+end
+
+--- Search through all the loaded packages for those that
+--- return a function, then cross reference them with all
+--- the plugin configs and setups and if there are any matches
+--- reload the user module.
 local function precompile(plugs)
   local reverse_index = {}
   for k, v in pairs(package.loaded) do
@@ -376,14 +388,11 @@ local function precompile(plugs)
         reverse_index[v] = k
     end
   end
-  -- TODO also reload setup as well as config
   for _, plugin in pairs(plugs) do
-    local value = plugin.config
-    local package_name = reverse_index[value]
-    if package_name then
-      package.loaded[package_name] = nil
-      require(package_name)
-    end
+    local cfg = reload_module(reverse_index[plugin.config])
+    local setup = reload_module(reverse_index[plugin.setup])
+    if cfg then plugin.config = cfg end
+    if setup then plugin.setup = setup end
   end
 end
 
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -401,7 +401,8 @@ end
 packer.compile = function(output_path)
   output_path = output_path or config.compile_path
   precompile(plugins)
-  local compiled_loader = compile(plugins)
+  -- NOTE: we copy the plugins table so the in memory value is not mutated during compilation
+  local compiled_loader = compile(vim.deepcopy(plugins))
   output_path = vim.fn.expand(output_path)
   vim.fn.mkdir(vim.fn.fnamemodify(output_path, ":h"), 'p')
   local output_file = io.open(output_path, 'w')


akinsho:
Ok some debugging and I forgot to update the `plugins` variable which is used in `compile` although this took things down from needing to reload twice to needing to reload `plugins.lua` once still for this to take effect.

Still trying to figure out why I have to reload that file at all 🤔 presumably the `require'packer'.startup` is a cached call to require so I wouldn't expect that to be changing anything

akinsho:
So it seems that the value of `plugins[x].config` goes from `<function-ref>` to `plugins[x].config = <StringifiedFunction>` after the first compile, so essentially the edit only works the first time since on subsequent runs the function reference will be being compared to a stringified function. The stringification seems to be for use in saving the table output to the compiled file since the function reference cant be output into the file

akinsho:
any ideas @wbthomason ?

akinsho:
Actually I managed to find a solution. So it seems that the plugins table was being mutated every time compile was called, the `config` values would be stringified so it could be output into the vim file.

As far as I can see we don't actually need the in memory table to change in this way just the output needs to have a particular format. So rather than giving compile the actual table, I deep copy it and pass that to compile.

Now when you make a change in a config file and run packer compile the generated output is automatically changed.

It might be nice to also source the `packer_compiled.vim` as well after compile is run but I'm not sure how that will behave and maybe don't want to muddy the waters in this PR 🤷🏿

wbthomason:
Sorry for my slow response! I'm fine with this workaround for now - the "proper" fix is to adjust how the `compile` module builds its output (it should be using a separate table anyway), but that's something I can easily add after we merge this initial PR.

The problem with sourcing the compiled file immediately is that it won't have the correct state for which plugins are loaded. Normally, this isn't a problem - `packadd`-ing twice is a nullop - but if it also runs stateful `config` or `setup` code twice, then we could have problems.

Perhaps the solution isn't to source the entire `packer_compiled.vim`, but only the lines corresponding to plugins still unloaded at compile time? We have the information necessary to do this, but it still presents problems: we may *want* to reload plugins if we've changed their config. If `config` functions are idempotent, this isn't an issue, but I'm not sure we can make that assumption.

Anyway, that's a lot of muttering and rumination to say: I need to keep thinking about how to do this, but it's probably best for another PR.

lua/packer.lua
```diff
@@ -369,11 +369,40 @@ packer.sync = function(...)
...
+  end
+end
+
+--- Search through all the loaded packages for those that
+--- return a function, then cross reference them with all
+--- the plugin configs and setups and if there are any matches
+--- reload the user module.
+local function precompile(plugs)
+  local reverse_index = {}
+  for k, v in pairs(package.loaded) do
```
wbthomason, line 386: We might be able to replace this loop with `vim.tbl_add_reverse_lookup`. Not a big deal either way - this loop is certainly more specialized to our needs.

lua/packer.lua
```diff
@@ -369,11 +369,40 @@ packer.sync = function(...)
...
+    package.loaded[name] = nil
+    return require(name)
+  end
+end
+
+--- Search through all the loaded packages for those that
+--- return a function, then cross reference them with all
+--- the plugin configs and setups and if there are any matches
+--- reload the user module.
+local function precompile(plugs)
```
wbthomason, line 384: Super nit-picky: I'd prefer we call this something else, like `refresh_configs` or similar. `precompile` is a bit vague.

wbthomason (changes requested):
Looks good - just a couple small tweaks to consider. Thank you!

lua/packer.lua
akinsho: happy to change it 👍🏿 I just went with something vaguely applicable

akinsho:
@wbthomason regarding sourcing the file I ended up adding a source call locally earlier today and it seemed to behave fine i.e. I didn't notice any unusual behaviour and my plugins just updated straightaway (where they were designed to) but doesn't guarantee that it won't break for others.

IMO the main use case for needing it reloaded is say I tweak my config mid session which is almost always because something is running already and I want to change it. The equivalent I'd do in vimscript is tweak the value and reload my vimrc and hope the plugin responds

Re. idempotent configs and setups, based on the configs and setup examples I just assumed they would be, I can't imagine what someone would be doing in there that'd depend on state since you aren't allowed upvalues so there cant be any local variable outside the function, of course you could require another module that does mutate something but that module wouldn't be being reloaded so 🤷🏿 .

lua/packer.lua
akinsho: Tbh I couldn't
 a) tell if it would mutate the table in place, and given it's the `package.loaded` table I didn't want to risk it
 b) it works by adding the reverse look up to the existing table (copy or not is unclear), seems like it'd double the size of this table which isn't a huge deal since we aren't iterating through it, but makes it harder to make sense of 🤷🏿‍♂️ 

TLDR: it just felt like it'd only be a slight improvement/if that over just a loop that does specifically whats needed.

wbthomason: Totally reasonable; I'm good with this tradeoff.

wbthomason (approved):
Looks good. I'm happy to either merge this or wait and add live-reload functionality. If we do the latter, I think we should also add a config variable to enable/disable, in case some people's configs are broken by live reloading.

Thank you for all your work on this!

akinsho:
Thanks @wbthomason I'm good with this bit being merged and I'll make another PR to add the (configurable) live reloading 👍🏿

wbthomason:
Sounds good to me. Thanks again!

PR #159, opened 2021-01-13
Fix typos for lazy loading via functions

weilbith:
I'm very sorry, but during the requested name changing of the variables from `func` to `fn` I forgot an occurrence (because it was a plain string) and another I adopted wrongly. 

Furthermore while using this new feature I discovered that autoload functions appear not to work. The `FuncUndefined` event does not get emitted for these. I hope that it is only a matter of escaping the `#` character and not a completely special treatment of autoload functions. I'll open a new PR for such a feature extension (if you wanna call it like that).

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -279,7 +279,7 @@ local function make_loaders(_, plugins)
 
           for _, fn in ipairs(plugin.fn) do
             fns[fn] = fns[fn] or {}
-            table.insert(fns[fns], quote_name)
+            table.insert(fns[fn], quote_name)
           end
         end
       end
@@ -482,7 +482,7 @@ then
   vim.list_extend(result, keymap_defs)
   table.insert(result, '')
 
-  -- The filetype, event and func autocommands
+  -- The filetype, event and function autocommands
   table.insert(result, 'augroup packer_load_aucmds\n  au!')
   table.insert(result, '  " Filetype lazy-loads')
   vim.list_extend(result, ft_aucmds)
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Good catch - sorry, I should've noticed this when reviewing before.

wbthomason:
re: autoload functions: I'm not certain, but http://vimdoc.sourceforge.net/htmldoc/eval.html#autoload-functions seems to imply that `FuncUndefined` and `autoload` may be mutually exclusive.

....Actually, upon some very basic testing, it seems that `FuncUndefined` still gets triggered for functions of the pattern `foo#bar` for me.

I defined `au FuncUndefined * echom expand('<afile>')` then ran `:call foo#bar()`. As expected, I get `foo#bar` printed to the message line.

Do you have a test case for where `FuncUndefined` *doesn't* get triggered?

weilbith:
Your approach to experiment is good. I tested it with an action function. So this is strengthening my guess that it is probably the `#` which is causing the trouble. Try it with `autocmd FuncUndefined foo#bar echom "foobar" `. This does not work for me. Actually I also though that eventually prefix matching is working. So you could use something like `plugin_name#` to lazy load. Saves unnecessary long/specific lists of function names.

wbthomason:
Interesting - the example you give *does* work for me (i.e. I see `foobar` printed). What platform are you on?

weilbith:
I should try my own examples I write practically. I wrote this from my phone as I remember I tried it with an exact example of a plugin on my PC. Now copy pasting my example it works here as well. So I must have made a different mistake in my previous experiments. :thinking: 
But at least the prefix matching does not work for me. My test (which I actually tried :stuck_out_tongue_winking_eye: ): 
```vim
:autocmd FuncUndefined bar# echom "undefined bar prefix function"
:call bar#what#ever()
:messages
```
This prints nothing for me except the error message by Vim.

wbthomason:
Yep, I was going to suggest using the asterisk. Seems like things are working?

Issue #161, opened 2021-01-14
attempt to index a boolean value


F4IL3D:
Hi Wil,

for 2-3 days i have been unable to load the packer module.
In order to rule out an error in my config, I carried out the following steps:

nvim -u NORC --noplugin
:packadd packer.nvim
:lua require('packer').startup(function() use {'wbthomason/packer.nvim'} end)

This then leads to the following error:

E5108: Error executing lua [string ":lua"]:1: attempt to index a boolean value

![image](https://user-images.githubusercontent.com/48415231/104608115-e6ed1880-5681-11eb-8229-e4ac75f2a8a8.png)

Even stranger, `:lua print(vim.inspect(require('packer')))` returns _true_ 

NVIM Version: v0.5.0-dev+1023-gea8756f85
Packer Version: latest (today)
OS: Arch Linux (WSL2)

akinsho:
@F4IL3D I've become pretty familiar with this error you probably have something invalid in your `packer_compiled`. For me when I was first setting up it was that I had an `upvalue`(using a local variable inside a packer config function) in my `packer_compiled.vim` as a quick test if you delete this file and re-open nvim does that work?
1. If it does then try removing all `setup`,`config` etc. from your plugin declarations and re-run packer compile and see if the error happens again.
2. If it doesn't it means one of `functions` in one fo the plugin configs is using a local variable, which isn't allowed.

I'd been wondering if it's possible to capture this error and embellish it with some sort of hint but it's quite generic, I wonder if all of the `packer_compiled.vim` script can be placed in a `try/catch/finally` block and all exceptions are caught and somehow massaged into something more coherent e.g. `[packer] Your packer compiled file is likely invalid please inspect you plugins`. Thoughts @wbthomason (also sorry for sticking my nose in 😆 )

wbthomason:
@akinsho: That's a reasonable theory, but I would think that @F4IL3D's use of `--noplugin` and `-u NORC` would prevent `packer_compiled.vim` from being sourced, right? I separately like the idea of providing better error messages from `packer_compiled`; I should be able to throw something together for that now. (and no worries for "sticking your nose in"; I'm always happy to have debugging help :smile:)

@F4IL3D: In case I'm wrong about `packer_compiled.vim` not getting sourced because of your flags, could you please try deleting or moving that file and see if the error persists? Alternatively, there's no chance you have a `packer.lua` anywhere that could be getting found first, right?

F4IL3D:
[...]

PR #162, opened 2021-01-14
Wrap packer_load with pcall

wbthomason:
As suggested by @akinsho in #161, wrap `packer_load` in `pcall` to report errors with a little more context.

@akinsho: Is this what you had in mind? I suppose we might also want to wrap the entire script in a vimscript `try...catch` too.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -17,7 +17,7 @@ endif
 
 local vim_loader = [[
 function! s:load(names, cause) abort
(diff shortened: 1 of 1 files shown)


akinsho:
@wbthomason yep this is exactly what I had in mind 👍🏿

wbthomason:
Cool, this seems to work from my limited testing, so I'll go ahead and merge it in. Thanks for the suggestion!

Issue #56, opened 2020-09-26
Disabled plugins get picked up by PackerClean


nanotee:
I sometimes want to disable plugins along with their associated configuration while still keeping them installed and up to date. I assumed that was the goal of the `disable` option but the `:PackerClean` command asks me if I want to remove plugins that are disabled. Am I misunderstanding the purpose of `disable`?

wbthomason:
Ah, I should clarify this in the docs. Yes, `disable` is meant as a mechanism for temporarily removing a plugin - it is by design that `PackerClean` removes disabled plugins. For keeping a plugin around and up to date, but not loading it, it is intended to use `opt = true` (manually setting the plugin as `opt`).

I can see how this would be confusing, though. Perhaps I should make `disable` effectively just set `opt = true` and add some other keyword (though I don't know what would be a good choice for this) for the current `disable` functionality.

clason:
Note that you can't rely on `opt=true` disabling the plugin, since there might be a `packadd` somewhere else in the config...

The safest way to handle that would be to move disabled plugins to another `pack/**/disabled` directory.

wbthomason:
That's true - my intent was that if a user manually marks `opt = true`, then they are responsible for loading or not loading it manually, as they prefer. Perhaps that's also a bad assumption.

clason:
That is not an unreasonable assumption. (In my case, it's because they're lua plugins so I need to manually load them anyway to configure them from init.vim.)

Regarding `disable = true`, what does this differently from just commenting out the `use` line in the config? (I guess that is the source of the confusion -- people would expect it to do something different since commenting out is easier than adding the keyword?)

wbthomason:
That's fair. The difference is mostly in a capability I don't think I've ever explicitly described: `disable` enables conditional installation. Consider the following:

Imagine you use both Windows and Linux machines, and there's some plugin `foobar/WindowsVim` that you only like to use on Windows. You could then write:
```lua
use { 'foobar/WindowsVim', disable = test_for_windows() }
```
and use your configuration across both of your machines without needing to edit it, while still only having `WindowsVim` be installed on Windows.

There are, of course, other ways to do this - you could install but not load `WindowsVim` by using the `cond` keyword, or you could just wrap the `use` call in a conditional, e.g.
```lua
if test_for_windows() then
  use 'foobar/WindowsVim'
end
```
but I find `disable` to be more concise and a little nicer.

wbthomason:
[...]

PR #163, opened 2021-01-14
Only try and remove disabled plugins that are installed

akinsho:
This relates a bit to #56. I read through that issue and so I'm hoping I didn't misunderstand how this is intended to work. @wbthomason mentions in that issue that `disable` should work similarly to a conditional i.e. if a plugin is disabled it should be removed and the declaration ignored generally as though it is in an if statement.

### Problem
If plugins are disabled they are picked up for cleaning *which is correct* as far as I understand. They are cleaned i.e. removed from the system. *Now for the problem*, __subsequent calls to PackerClean still try to remove these plugins__ even though they are no longer installed.

### Solution
`PackerClean` should not offer to remove things that do not exist on the system.

If you look carefully at this screen shot you will see that packer is offering to clean plugins that do not exist inside the directory shown in the terminal

packer-disabled-clean

Since I'm not 100% certain I'm still not missing something, lemme know if maybe this is still the desired behaviour

### DIFF
--- a/lua/packer/clean.lua
+++ b/lua/packer/clean.lua
@@ -13,7 +13,7 @@ local PLUGIN_OPTIONAL_LIST = 1
 local PLUGIN_START_LIST = 2
 
 local function is_dirty(plugin, typ)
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Looks great, thank you!

PR #164, opened 2021-01-14
Auto reload packer compiled on compile

akinsho:
This is a follow up to #158, now that required configs and setups are reloaded when packer compile is run, this pr adds a config-gated way to also automatically re-source the `packer_compiled.vim`. The config key I chose is `auto_reload_compiled` and defaults to `true` but I'm happy to change either.

I defaulted to `true` since I think/hope the vast majority of users have simple enough configs and they'd automatically get the benefit of things reloading without having to discover this option first and this could be an opt out for people with complex configs.

NOTE: this doesn't work with plugins that seem to be optional or lazy loaded, I haven't dug into the full compiled output but I'm guessing those plugins setups etc are skipped till they are loaded which is expected but also means that when they load they don't update

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ default configuration values (and structure of the configuration table) are:
   opt_default = false, -- Default to using opt (as opposed to start) plugins
   transitive_opt = true, -- Make dependencies of opt plugins also opt by default
   transitive_disable = true, -- Automatically disable dependencies of disabled plugins
+  auto_reload_compiled = true, -- Automatically reload the compiled file after creating it.
   git = {
     cmd = 'git', -- The base command for git operations
     subcommands = { -- Format strings for git subcommands
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -33,6 +33,7 @@ local config_defaults = {
   opt_default = false,
   transitive_opt = true,
   transitive_disable = true,
+  auto_reload_compiled = true,
   git = {
     cmd = 'git',
     subcommands = {
@@ -408,6 +409,9 @@ packer.compile = function(output_path)
   local output_file = io.open(output_path, 'w')
   output_file:write(compiled_loader)
   output_file:close()
+  if config.auto_reload_compiled then
+    vim.cmd("source "..output_path)
+  end
   log.info('Finished compiling lazy-loaders!')
 end
 


wbthomason (commented):
Thanks! Yes, the reason that this will not reload any lazy-loaded plugins that are already loaded is that this effectively re-makes the lazy-loaders, and does not check which have already run.

If, as suggested in #167, we make the `plugins` table in `packer_compiled` global, then we could add to this logic a check for lazy-loaded plugins with `loaded = true` in `plugins`, and re-run their setup/config functions.

What do you think?

akinsho:
@wbthomason Sorry for the late reply, been a bit busy. I guess this can wait till a global plugins table is implemented, although I don't think I follow regarding checking if plugins are lazy loaded. The logic to reload `setup` and `config` functions works regardless of lazy loading.

I was more pointing out that if you sourced the compiled output it seemed like whatever the changes to the lazy loaded plugins config (though picked up in the new compiled output) weren't applied. I think I recall seeing an early return for unloaded plugins in the generated vimscript.

I need to refresh my memory when I have a bit more time since I actually can't remember exactly what the behaviour was and I'm not sure the comment I left here is that clear in hindsight

wbthomason:
@akinsho No worries, I've also been rather busy :sweat_smile: 

The "check for lazy-loaded" plugins feature I was referring to is a way to ensure that changes to lazy-loaded plugins actually get applied. I think this doesn't currently happen because we do not check which lazy-load triggers have already run. This means that e.g. a plugin which loads on a given filetype won't reload (and therefore have the updated config applied) until that filetype event triggers again.

What I was proposing was that, once the global `packer_plugins` table exists, we can iterate through that table and check which plugins are (1) lazy-loaded and (2) marked as already loaded. Those are the plugins which we need to re-setup/config immediately, rather than waiting for their corresponding lazy-load events to trigger again.

wbthomason:
Am I forgetting something, or is this ready for merge?

akinsho:
This should be fine as is. I think the issue with the lazy loaded plugins I'd have to look into again since tbh I don't actually recall exactly what the issue was, but I think it can probably be investigated and resolved separately.

wbthomason:
Sounds good - I'll merge this and we can add the improved lazy-reloading in another PR.

For that, we shouldn't need to run any vimscript - before we source the new compiled file, we can look through the `packer_plugins` table to find any plugins which need config reloading (i.e. any plugins which are already loaded). We can also set these same plugins to `loaded = true` in the new `packer_plugins` after sourcing the new compiled file, to avoid lazy-loads getting run again.

PR #166, opened 2021-01-14
Fix #165: escape spaces in the compile_path and package_root

### LINKED ISSUE #165: PackerInstall fails because git clone has to many arguments
I trying to setup Packer for my nvim config, Nvim is seeing packer and I'm able to run `PackerInstall`.

The clone operations of the plugins will result in a git clone failure with `Fatal: Too many arguments`

I think it fails because the resulting clone directory has a space in it, the user on my machine has a space in the user name.

Thanks in advance!

wbthomason:
Account for `package_root` and `compile_path` settings which may contain a space. If not escaped, this causes an error with `git` and other commands, per #165.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -86,7 +86,8 @@ packer.init = function(user_config)
 
   packer.reset()
 
-  config.package_root = vim.fn.fnamemodify(config.package_root, ':p')
+  config.package_root = vim.fn.escape(vim.fn.fnamemodify(config.package_root, ':p'), ' ')
+  config.compile_path = vim.fn.escape(config.compile_path, ' ')
   local _
   config.package_root, _ = string.gsub(config.package_root, util.get_separator() .. '$', '', 1)
   config.pack_dir = util.join_paths(config.package_root, config.plugin_package)


wbthomason:
@koenverberg: Could you please test this branch and see if it fixes the issue for you? Thanks!

koenverburg:
@wbthomason I set my local instance of packer to this branch and opening up nvim will start with a message

`E5113: Error while calling lua chunk: Vim:E739: Cannot create directory C:\Users\Koen: operation not permitted`

wbthomason:
Drat, ok. What is your full `package_root` path?

wbthomason:
@koenverburg: Sorry, but if you have a chance to give this PR another try, I realized that `shellescape` is probably what I want more than `escape`. Hopefully this actually fixes the issue? If not, please post your `packer` config so that I can do some testing.

koenverburg:
My `package_root` is `C:\Users\Koen Verburg\AppData\Local\nvim-data\site\pack`.

I created a branch on my dotfiles for this, here is the [packer config](https://github.com/koenverburg/dotfiles/pull/4/files#diff-86d2bfd8811f4c409ddb0c00c1205a6cea0b2260bff9a51cc89ed2b673bd38a9R1)

Thank you for the rapid response and let me know if there is something wrong with the config.

wbthomason:
Thanks, I'll take a look in the morning my time. Just to confirm, the latest commit on this PR still causes the same problem for you?

koenverburg:
No problem, The last commit resulted in the same message when starting up nvim

wbthomason:
Hmm. I don't think it would cause this issue, but I am a bit confused by this line: `local install_path = vim.fn.stdpath('data') .. '/site/pack/packer/opt/packer.nvim'`

It seems you're on Windows - could the forward slashes in the path (rather than the usual backslashes) be causing trouble? Apologies if this is a dumb question; I don't typically use Windows.

koenverburg:
that's to locate the `nvim-data` folder in which all the plugins will be downloaded.
haha, no worries, I replaced to forward slashes with double backslashes but it did not make a difference. I tried it with Paq-nvim and that works for now.

I also found that the download paths with `C:\Users\first\ lastname\AppData\etc.` required permission on windows for some reason and `C:\Users\first lastname\AppData\etc.` did not.

Another thing I noticed was while setting up the packer bootstrap, I had to put quotes around the clone directory because of the space in my username. The result of the clone would be `git clone <repo> "<directory>"`

I also tried that in the `subcommand.install` string but that did not work.

koenverburg:
@wbthomason The issue is harder than I thought, from what I can see in the code the git clone is spawned in the current shell nvim is running, correct me if I'm wrong.


In Powershell, you can escape a space with ` and that will just work.
I can successfully run the following git clone commands in PowerShell

```PowerShell
git clone https://github.com/glepnir/zephyr-nvim "C:\Users\Koen Verburg\AppData\Local\nvim-data\site\pack\packer\start\zephyr-nvim" --depth 1 --no-single-branch --progress

git clone https://github.com/glepnir/zephyr-nvim C:\Users\Koen` Verburg\AppData\Local\nvim-data\site\pack\packer\start\zephyr-nvim --depth 1 --no-single-branch --progress
```

wbthomason:
Yeah, I admit that I haven't been able to make any progress on this either. One easy workaround should be just manually specifying your install dir (rather than using `stdpath`) and escaping the space yourself, but that should not, in general, be necessary.

Stupid question: if, in `packer.lua:packer.init`, you add `config.package_root = '"' .. config.package_root .. '"'` after the `string.gsub` line, does that fix the problem? I suspect not, but it may be worth a shot.

PR #171, opened 2021-01-21
Expose packer plugins table

### LINKED ISSUE #167: Check if plugin is loaded
Is there a way to check if an arbitrary plugin is loaded?

```lua
if packer.is_loaded("neoclide/coc.nvim") then
  print("coc.nvim is available!")
end
```

---

`vim-plug` provides a global `g:plugs` which can be used to determine if a plugin is loaded. For instance:

```vim
function! PlugLoaded(name)
    return (
        \ has_key(g:plugs, a:name) &&
        \ isdirectory(g:plugs[a:name].dir) &&
        \ stridx(&runtimepath, trim(g:plugs[a:name].dir, "/")) >= 0)
endfunction

if PlugLoaded("coc.nvim")
    echo "coc.nvim is available!"
endif
```

akinsho:
This PR will fix #167 by exposing the packer plugins table as a `lua` global as `packer_plugins`. I chose to prefix it since `plugins` as a global seemed way too likely to clash with something.

A user can now do
```lua
if packer_plugins['coc.nvim'] and packer_plugins['coc.nvim'].loaded then
 print('Coc is Loaded!')
end
```

@wbthomason was this what you had in mind? 

Changes:
* Rename `plugins` -> `packer_plugins` I explictly used `_G` in the definition so it's clearer that this is intentional.
* If a plugin is `not opt` I set it's loaded value to true

I tested this locally and nothing *seemed* to be broken but would appreciate a second pair of eyes
Also this seems like it should be clearly documented somewhere but I don't know where. `vim-plug` has a wiki for these sort of tips and tricks, could also be in the readme 🤷🏿 ?

I'd suggest hiding whitespace changes so the change of indentation doesn't cause a bunch of noise during review

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,19 @@ use {
 }
 ```
 
+#### Checking plugin statuses
+You can check whether or not a particular plugin is installed with `packer` as well as if that plugin is loaded.
+To do this you can check for the plugin's name in the `packer_plugins` global table.
+Plugins in this table are saved using only the last section of their names
+e.g. `tpope/vim-fugitive` if installed will be under the key `vim-fugitive`.
+
+```lua
+if packer_plugins["vim-fugitive"] and packer_plugins["vim-fugitive"].loaded then
+print("Vim fugitive is loaded")
+-- other custom logic
+end
+```
+
 #### Luarocks support
 
 You may specify that a plugin requires one or more Luarocks packages using the `rocks` key. This key
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -251,6 +251,18 @@ CONFIGURING PLUGINS                            *packer-plugin-configuration*
 `setup` key described in |packer.use()|) or after they are loaded (the
 `config` key described in |packer.use()|).
 
+PLUGIN STATUSES                                 *packer-plugin-status*
+You can check whether or not a particular plugin is installed with `packer` as
+well as if that plugin is loaded. To do this you can check for the plugin's
+name in the `packer_plugins` global table. Plugins in this table are saved
+using only the last section of their names e.g. `tpope/vim-fugitive` if
+installed will be under the key `vim-fugitive`.
+>
+  if packer_plugins["vim-fugitive"] and packer_plugins["vim-fugitive"].loaded then
+  print("Vim fugitive is loaded")
+  -- other custom logic
+  end
+
 CUSTOM INSTALLERS                              *packer-custom-installers*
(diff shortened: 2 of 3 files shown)


wbthomason (approved):
This looks good to me, thanks! We should add something about this feature to the README/vimdocs; it can either be part of this PR, or I can write something up after we merge this.

Thanks again!

akinsho:
Sounds good @wbthomason, I'm gonna be a bit busy the next couple of days so won't be able to add the docs on my end till I'm free. Happy to have it merged and come back to add the docs later or if you wanted to/had the time I'm fine with that also.

wbthomason:
Sounds good; thanks. I'll either push some docs to this branch in the next day or two or just merge and add docs later.

akinsho:
@wbthomason had a bit of time so added the docs to the README and help doc in 👍🏿

wbthomason:
Wonderful, thank you! Takes a task off of my plate for today.

Thanks again for all your work on this (and your other contributions!)

wbthomason (approved):

olimorris:
No matter how I try and access `packer_plugins` I get the following error:

```
E5113: Error while calling lua chunk: ...e/nvim/site/pack/packer/start/packer.nvim/lua/packer.lua:476: /Users/Oli/.config/nvim/lua/plugins/lsp.lua:6: attempt to index global
'packer_plugins' (a nil value)
```

The file in question is called within the packer startup function as follows:

```lua
use {
        'neovim/nvim-lspconfig',
	event = {'BufRead *'},
	setup = require('plugins.lsp').setup(),
        config = require('plugins.lsp').config()
    }
```

and line 6 is:

```lua
if packer_plugins["nvim-lspconfig"] then
```

wbthomason:
Just to check, does your `startup` function run *after* `packer_compiled` is sourced? If not, then the `packer_plugins` table won't exist at the time that `startup` runs.

olimorris:
Great point. I was trialling setting up Neovim on a new MacBook. Basically, with a populated packer startup and zero compilation, it was freaking out. I've gone back to using a physical folder check for the plugin before loading it's config and setup files

vignesh0025:
How should we detect if packer_complied is sourced ?

wbthomason:
@vignesh0025 You should be able to check if the `_G.packer_plugins` table exists.

Arrow-x:
`packer_plugins` doesn't exist before `init.lua` it seems, ie when I try to use it in my `init.lua` it doesn't exist

Issue #174, opened 2021-01-25
Better examples for post-install/update hooks


matu3ba:
I am quite confused, where the problem of converting
```vim
Plug 'nvim-treesitter/nvim-treesitter', {'do': ':TSUpdate'}
Plug 'junegunn/fzf', { 'do': { -> fzf#install() } }
Plug 'glacambre/firenvim', { 'do': { _ -> firenvim#install(69) } }
```
is.

I tried
```lua
use { 'glacambre/firenvim', run = "vim.fn['fzf#install()']"}
use { 'glacambre/firenvim', run = "vim.fn['firenvim#install'](0)"}
use { 'glacambre/firenvim', run = 'vim.fn["firenvim#install"](0)'}
...
```

However the installer fails without further report on what is wrong.

Can you shed some light on this? I am feeling that I miss something trivially simple.

wbthomason:
So, from the docs: 
> If `run` is a string, then either:
> 1. If the first character of `run` is ":", it is treated as a Neovim command and executed.
> 2. Otherwise, `run` is treated as a shell command and run in the installation directory of the
   plugin via `$SHELL -c 'cd <plugin dir> && <run>'`.

`run` can also be a function - which is actually what you're doing in the `Plug` example, as `{ -> ...}` is VimL for a lambda.

The most idiomatic translation for your example would probably be:

```lua
use { 'glacambre/firenvim', run = function() vim.fn['firenvim#install'](0) end}
```

matu3ba:
Thanks alot. Got all working with
```lua
use { 'nvim-treesitter/nvim-treesitter', run = ':TSUpdate' }
use { 'junegunn/fzf', run = function() vim.fn['fzf#install']() end }
use { 'glacambre/firenvim', run = function() vim.fn['firenvim#install'](0) end }
```

Can add the firenvim example to the README ?

wbthomason:
I'd be happy to merge a PR adding examples!

wbthomason:
Closing per #178. Thanks again!

PR #178, opened 2021-01-25
Add different examples of calling vimscript and lua to README (#174)

matu3ba:


wbthomason (approved):
This is great, thank you!

PR #183, opened 2021-01-28
Add plugin update diff preview

### LINKED ISSUE #78: Feature request: show hunk diff
Hi Will, please add ability to see diff of each commit from update result window. `vim-plug` has that as well as `:PlugDiff` which can show results of last `:PlugUpdate` command once again.

Great plugin, thanks a lot!

akinsho:
This PR adds the ability to view a diff of plugin changes when they are updated. It is similar to the `PlugDiff` command from `vim-plug`. So a user can see what bits of the code were changed. Fixes #78 

packer_diff

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -63,7 +63,7 @@ local config_defaults = {
     header_lines = 2,
     title = 'packer.nvim',
     show_all_info = true,
-    keybindings = {quit = 'q', toggle_info = '<CR>', prompt_revert = 'r'}
+    keybindings = {quit = 'q', toggle_info = '<CR>', diff = 'd', prompt_revert = 'r'}
   },
   luarocks = {python_cmd = 'python'}
 }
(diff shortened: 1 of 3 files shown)


wbthomason (commented):
This looks super slick, thanks! I do have one question - what do you think about adding a binding to the preview window to close the preview (or is this already something the preview window provides)? It seems like it would be convenient to bind e.g. `q` to close a diff, particularly as this jumps the user's cursor to the preview window (if I'm reading the code correctly).

akinsho:
@wbthomason that sounds good tbh I have an autocommand in my config for a whole host of window to map `q` to quit so I occasionally forget that that behaviour isn't default 😅 . I'm assuming you mean a binding within the preview window itself right?

wbthomason:
Yes, that is what I mean, sorry.

akinsho:
done 🚀

wbthomason (approved):
Looks good, thank you!

PR #186, opened 2021-01-29
Diff command should show commit for the given line

akinsho:
This PR fixes a bug/enhances the changes in #183. Currently the `diff` command will show the first revision in the plugin's revisions which doesn't actually map to the changes shown for the update. This PR makes this functionality work exactly like the `PlugDiff` command in `vim-plug` i.e. hitting `d` on a given line show the diff *for that commit* as well as the commit message in full

packer_diff


wbthomason (approved):
Looks good, thanks!

PR #188, opened 2021-01-30
Remove unncessary string escapes

rickyson96:
Currently, the string passed to `string.format` is escaped using `vim.fn.escape`.
Then the `string.format` will escape the resulting string too.
The result will be something like this.
`echom "[packer.nvim] [WARN  02:29:49] packer.lua:156: Plugin \\"telescope.nvim\\" is used twice!"`

This fixed it so that vim won't cause errors when printing logs.

### DIFF
--- a/lua/packer/log.lua
+++ b/lua/packer/log.lua
@@ -88,7 +88,7 @@ log.new = function(config, standalone)
 
     local split_console = vim.split(console_string, "\n")
     for _, v in ipairs(split_console) do
-      vim.cmd(string.format([[echom "[%s] %s"]], config.plugin, vim.fn.escape(v, '"')))
+      vim.cmd(string.format([[echom "[%s] %s"]], config.plugin, v))
     end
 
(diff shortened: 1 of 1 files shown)


wbthomason:
Hmm. I'm not convinced that this escape is unnecessary - I think the proper fix is to ensure that we don't pass escaped quotes to `log` function calls. If a raw string passed to a `log` function contains unescaped quotes and we remove this call to `escape`, then we'll get the same error.

wbthomason:
Thanks for this PR/letting me know about the problem! I think b1f7170 may fix this problem for you - could you confirm? Thanks!

rickyson96:
I see.
I didn't know that the `log` function is used for other logging as well. My bad!

I've tested b1f7170, it works! 

Thanks! 👍

wbthomason:
No worries, and thank you again for the PR! I hadn't realized that this bug existed. Thanks!

PR #194, opened 2021-02-02
Various fixes related to keymap lazy-loading

nanotee:
The changes introduced in e1aa188c2bf61c9a791df7fc3ccd3e59b946f3ce caused a few regressions:

- The `prefix` key for lazy-loading keymaps still used the syntax for Vim dictionaries
- Since the code now uses Lua strings, the `"\<"` and `"\>"` escape sequences are invalid. `<` characters have to be replaced with `<lt>`

Let me know if you have comments/concerns!


wbthomason (approved):
Shoot, thought I'd caught all the Vim syntax -> Lua syntax issues. Thanks!

PR #205, opened 2021-02-09
Update luarocks.lua

### LINKED ISSUE #203: Can't Use in RaspberryPi (aarch64 Architecture)
In aarch64 neovim is use Lua5.1 NOT LuaJIT
So it has error in it.

mirachan010:
to set if jit = nill for don't use LuaJIT
Fixes #203


wbthomason:
Thanks for this! Unfortunately, I think this fix will cause issues later, when we try to concatenate `lua_version.jit`, etc. 

I think the proper fix will be changing the Luarocks initialization to optionally not use LuaJIT (or else returning early with a warning that LuaJIT is required for `packer`'s Luarocks functionality)

PR #207, opened 2021-02-10
Return value in try_loadstring

gbrlsnchs:


wbthomason (approved):
Thanks!

gbrlsnchs:
@wbthomason Just a heads-up, I quickly fixed and rebased this PR because the value I should return was the second one returned by `pcall` and not the first. You merged it correctly but you probably only saw the first version, so I just wanted to let you know.

wbthomason:
Thanks! I actually did see the second version before merging; I was going to patch that issue myself after seeing just the first version, but you beat me to it.

gbrlsnchs:
Haha all good then! :smile_cat:

PR #210, opened 2021-02-11
Add arguments to luarocks installation

akinsho:
This PR will hopefully eventually add argument keys to `use_rocks` to be passed on to luarocks during installation.
There are currently a few bugs/kinks to work out/ a potentially breaking change for any early adopters.

Currently the API for specifying luarocks allows specifying the version of a rock as
```lua
use_rocks {"cjson", "2.1.0"}
user_rocks {"cjson", "penlight"}
```
I might have misunderstood the intended patterns for usage but the above 2 patterns are in conflict short of using a regex or something to detect that one of the two patterns includes a version number.

### Example of such an issue (attempts to install `2.1.0`)
packer-version-bug

@wbthomason would appreciate some input on how this API should work, in this PR I've tested moving the version specification to an explicit key, since it makes things clearer and allows listing strings to work since it doesn't have to consider that a list of strings might contain a version

#### TODO
- [x] Fix incorrectly re-installing already installed plugins
- [x] Check for potential breakages in other code paths

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,7 +347,8 @@ end
 You may specify that a plugin requires one or more Luarocks packages using the `rocks` key. This key
 takes either a string specifying the name of a package (e.g. `rocks=lpeg`), or a list specifying one or more packages.
 Entries in the list may either be strings, a list of strings or a table --- the latter case is used to specify arguments such as the
-particular version of a package, e.g.
+particular version of a package.
+Supported lua rocks keys are: `server`, `only_server`, `only-sources`
 ```lua
 rocks = {'lpeg', {'lua-cjson', version = '2.1.0'}}
 use_rocks {'lua-cjson', 'lua-resty-http'}
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ return require('packer').startup(function()
   -- Plugins can also depend on rocks from luarocks.org:
   use {
     'my/supercoolplugin',
-    rocks = {'lpeg', {'lua-cjson', '2.1.0'}}
+    rocks = {'lpeg', {'lua-cjson', version = '2.1.0'}}
   }
 
   -- You can specify rocks in isolation
@@ -346,8 +346,13 @@ end
 
 You may specify that a plugin requires one or more Luarocks packages using the `rocks` key. This key
 takes either a string specifying the name of a package (e.g. `rocks=lpeg`), or a list specifying one or more packages.
-Entries in the list may either be strings or lists --- the latter case is used to specify a
-particular version of a package, e.g. `rocks = {'lpeg', {'lua-cjson', '2.1.0'}}`.
+Entries in the list may either be strings, a list of strings or a table --- the latter case is used to specify arguments such as the
+particular version of a package, e.g.
+```lua
+rocks = {'lpeg', {'lua-cjson', version = '2.1.0'}}
+use_rocks {'lua-cjson', 'lua-resty-http'}
+use_rocks {'luaformatter', server = 'https://luarocks.org/dev'}
+```
 
 Currently, `packer` only supports equality constraints on package versions.
 
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -399,7 +399,11 @@ local function ensure_packages(rocks, results, disp)
         local spec = to_install[package.name]
         if spec then
           if type(spec) == 'table' then
-            if spec.version == package.version then to_install[package.name] = nil end
+            -- if the package is on the system and the spec has no version
+            -- or it has a version and that is the version on the system do not install it again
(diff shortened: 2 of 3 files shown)


akinsho:
will fix #210

wbthomason:
Thanks! I'll take a look at the changes when the PR is ready (or if there are specific things you have questions about). 

I'm good with making rock versions explicit via a key; the original specification format is unambiguous if used correctly, but using it correctly may require an unintuitive nesting of tables. Using a key simplifies usage.

akinsho:
@wbthomason I *think* this is done now I've tested it with the common ways I'm aware of for interacting with rocks i.e. using `use_rocks` to install and remove and that works. I've also updated other code paths so they should work as well but would appreciate if you could give this a thorough look over + test when you get the time as I'm pretty unfamiliar with all the places where luarocks might be used

README.md
```diff
@@ -346,8 +346,13 @@ end
 
 You may specify that a plugin requires one or more Luarocks packages using the `rocks` key. This key
 takes either a string specifying the name of a package (e.g. `rocks=lpeg`), or a list specifying one or more packages.
-Entries in the list may either be strings or lists --- the latter case is used to specify a
-particular version of a package, e.g. `rocks = {'lpeg', {'lua-cjson', '2.1.0'}}`.
+Entries in the list may either be strings, a list of strings or a table --- the latter case is used to specify arguments such as the
```
wbthomason, line 349: I haven't yet looked to see if this PR allows all valid `luarocks` arguments to be specified, but, if it does not, then we ought to state which arguments are allowed here.

lua/packer/luarocks.lua
```diff
@@ -163,11 +163,33 @@ local function run_luarocks(args, disp, operation_name)
   end)
 end
 
+local luarocks_keys = {
```
wbthomason, line 166: Since this looks like we're only supporting these luarocks options, we should add mention of this to the docs.

lua/packer/luarocks.lua
```diff
@@ -300,8 +327,12 @@ local function clean_packages(rocks, results, disp)
       local to_remove = {}
       for _, package in ipairs(installed_packages) do to_remove[package.name] = package end
       for _, rock in pairs(rocks) do
-        if type(rock) == 'table' then
-          if to_remove[rock[1]] and to_remove[rock[1]].version == rock[2] then
+        if vim.tbl_islist(rock) then
```
wbthomason, line 330: I'm probably missing something obvious, but I'm a bit confused by the `tbl_islist` cases here and in `ensure_packages`. Shouldn't the `rocks` table contain as values only rocks specified as strings or as non-list tables (i.e. with version or server keys)? When do we encounter the list case?

wbthomason (changes requested):
Thanks, this is looking good. I have a few small comments - mostly, I think we need to be explicit about which `luarocks` arguments are/aren't supported, and I have a question about cases when rock specifications are lists (rather than strings or non-list tables).

lua/packer/luarocks.lua
akinsho: @wbthomason the it appeared that before a user could specify rocks using `use_rocks {'cjson', 'penlight', 'luaformatter}` so those are the cases that the `tbl_islist`. I might have misunderstood, if that isn't a required format since lists of tables aren't supported then I can just have it support a table or a string.

README.md
akinsho: Will do 👍🏿 , I only included the arguments that made sense to me to be able to set in packer, some seemed like they could mess things up e.g. `--tree` or `--local`. The full list is [here](https://github.com/luarocks/luarocks/wiki/luarocks).

wbthomason: Yup, supporting a limited set seems wise to me - I just want to make sure we document it. Thanks!

wbthomason (commented):
The docs change looks good, thanks!

lua/packer/luarocks.lua
wbthomason: You're right that this format is allowed, but - I think - by the time we reach `clean_packages` or `ensure_packages` - the `rocks` table had been converted to a non-list table. In particular, looking at `packer.lua:117` and `packer.lua:193`, I'm pretty sure we always perform this conversion before any call to a function in the `luarocks` module, so this case should be redundant. I could be forgetting something, though!

akinsho: @wbthomason you're right just tried without those checks and plugins are still added and removed fine and as you say things are correctly flattened. I think my thinking at the time was related to other code paths other than `use_rocks` but I guess those specify packages as strings

akinsho:
Removed the unnecessary conditionals 👍🏿

wbthomason (approved):
Looks good to me, thanks!

PR #215, opened 2021-02-14
Add a step to manually source files in after/plugin if they exist

### LINKED ISSUE #214: nvim-compe in opt dir, not work
https://user-images.githubusercontent.com/19209958/107872044-48531200-6ee2-11eb-8d46-e72ba6bced29.mp4

see this video.

I use packadd to use it ,but is not work

and [hrsh7th comment](https://github.com/hrsh7th/nvim-compe/issues/181#issuecomment-778744459)

we should deal with this situation

or I do something can lazy it and worked?

wbthomason:
Fixes #214.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -75,6 +75,13 @@ local function detect_after_plugin(name, plugin_path)
   return nil
 end
 
+local source_dirs = {'ftdetect', 'ftplugin', 'after/ftdetect', 'after/ftplugin'}
+local function detect_bufread(plugin_path)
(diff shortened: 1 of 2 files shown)


wbthomason:
Did some testing myself with lazy-loading `compe` in particular, and this branch seems to work.

Basically, we can check at compile-time (rather than load time) for the presence of files in `after/plugin`, then source those files when the plugin gets loaded.

Based on my testing, I'm going to go ahead and merge this, but @hamsterBiscuit please also confirm that this works for you.

PR #222, opened 2021-02-18
fix: string.fmt -> string.format

elianiva:
seems like a mistake from search and replace


wbthomason (approved):
Very much so; thanks for the catch!

Issue #187, opened 2021-01-29
Add ability to specify args to luarocks


wbthomason:
As mentioned in #184, some rocks require further args for luarocks, e.g. `server`. We should add the ability to specify these args (or some subset of the possible args) to a rock's specification, and have `packer` forward these to `luarocks`.

wbthomason:
Closed by #210.

wbthomason:
Note from Gitter discussion with @clason: we may need to expand the set of allowed arguments, given that some packages (e.g. openssl) need certain environment variables to be set, etc.

akinsho:
@wbthomason I *briefly* looked at this and the key @clason mentioned isn't actually a `luarocks` key so I'm not sure how we would represent these? the actual keys are formatted as `--key=value` in this case we'd need to add `OPENSSL_DIR=<value>` but to not append the key syntax to it we'd need to be able to flag these kind of keys differently. Any ideas? some sort of prefix? or a heuristic based on the casing?

clason:
@akinsho Maybe treat this as a separate thing (`env`?) to pass to `luarocks`?

akinsho:
Something like?
```lua
use_rocks {"openssl", env = {OPENSSL_DIR = "path/to/dir"}}
```

clason:
Yes, perfect!

PR #230, opened 2021-02-22
Allow env key and more luarocks arguments

akinsho:
@wbthomason, this resolves the issue raised in gitter and discussed in #187. It allows specifying an `env` key for `luarocks` so you can pass arguments such as `OPENSSL_DIR` for the `openssl` rock, as an example.
```lua
 use_rocks {"openssl", env = {OPENSSL_DIR = "<value>"}}
```
@clason could you try this branch out and see if it does what you need.

### DIFF
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -176,16 +176,16 @@ local function format_luarocks_args(package)
   if type(package) ~= 'table' then return '' end
   local args = {}
   for key, value in pairs(package) do
-    if type(key) ~= "string" or not is_valid_luarock_key(key) then goto continue end
-    local luarock_key = luarocks_keys[key] and luarocks_keys[key] or key
-    if luarock_key and type(value) == 'string' then
-      table.insert(args, string.format('--%s=%s', key, value))
-    elseif key == "env" and type(value) == "table" then
-      for name, env_value in pairs(value) do
-        table.insert(args, string.format('%s=%s', name, env_value))
+    if type(key) == "string" and is_valid_luarock_key(key) then
+      local luarock_key = luarocks_keys[key] and luarocks_keys[key] or key
+      if luarock_key and type(value) == 'string' then
+        table.insert(args, string.format('--%s=%s', key, value))
+      elseif key == "env" and type(value) == "table" then
+        for name, env_value in pairs(value) do
+          table.insert(args, string.format('%s=%s', name, env_value))
+        end
       end
     end
-    ::continue::
   end
   return ' ' ..table.concat(args, ' ')
 end
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -166,17 +166,26 @@ end
 local luarocks_keys = {
   only_server = 'only-server',
   only_source = 'only-sources',
-  server = 'server'
 }
+
+local function is_valid_luarock_key(key)
+  return not (key == 'tree' or key == 'local')
+end
 
 local function format_luarocks_args(package)
   if type(package) ~= 'table' then return '' end
   local args = {}
   for key, value in pairs(package) do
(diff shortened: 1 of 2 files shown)


clason:
Hmm, no, sorry, that doesn't seem to work for me. I'm using
```
{'XX/YY.nvim', rocks = {'lua-http-parser', {'openssl', env = {OPENSSL_DIR='/usr/local/Cellar/openssl@1.1/1.1.1j'}}}}
```
`lua-http-parser` installs correctly, but I still get
```
Error: Could not find header file for OPENSSL
  No file openssl/evp.h in /usr/local/include
  No file openssl/evp.h in /usr/include
  No file openssl/evp.h in /include
You may have to install OPENSSL in your system and/or pass OPENSSL_DIR or OPENSSL_INCDIR to the luarocks command.
Example: luarocks install openssl OPENSSL_DIR=/usr/local
```

lua/packer/luarocks.lua
```diff
@@ -166,17 +166,26 @@ end
...
+local function is_valid_luarock_key(key)
+  return not (key == 'tree' or key == 'local')
+end
+
 local function format_luarocks_args(package)
   if type(package) ~= 'table' then return '' end
   local args = {}
   for key, value in pairs(package) do
-    local luarock_key = luarocks_keys[key]
+    if type(key) ~= "string" or not is_valid_luarock_key(key) then goto continue end
```
wbthomason, line 179: Lua has labels and `goto`?? I've never seen this in the language reference.

As a small nitpick, I'd generally prefer if we inverted this conditional to avoid the `goto`, e.g.

```lua
if type(key) == 'string' and is_valid_luarock_key(key) then
....do the things....
end
```
but it's good to learn that `goto` is an option.

wbthomason (changes requested):
Thanks for this! It looks great - I just have one small style nitpick, with the use of `goto` rather than wrapping in a conditional.

wbthomason:
Also, sorry for how long it took me to get to taking a look.

lua/packer/luarocks.lua
akinsho: @wbthomason I *think* based on a commit I saw from TJ in Telescope earlier that maybe `goto`s aren't specifically lua 5.1 syntax and maybe shouldn't be used if a plugin is assuming 5.1 so maybe a good idea to remove it anyway

lua/packer/luarocks.lua
```diff
@@ -176,16 +176,16 @@ local function format_luarocks_args(package)
...
   local args = {}
   for key, value in pairs(package) do
-    if type(key) ~= "string" or not is_valid_luarock_key(key) then goto continue end
-    local luarock_key = luarocks_keys[key] and luarocks_keys[key] or key
-    if luarock_key and type(value) == 'string' then
-      table.insert(args, string.format('--%s=%s', key, value))
-    elseif key == "env" and type(value) == "table" then
-      for name, env_value in pairs(value) do
-        table.insert(args, string.format('%s=%s', name, env_value))
+    if type(key) == "string" or is_valid_luarock_key(key) then
```
wbthomason, line 179: I think this should be an `and` rather than an `or`, right? We want to process keys which are both a string (ensuring that `is_valid_luarock_key` will behave as expected) and for which `is_valid_luarock_key` returns `true`.

wbthomason (changes requested):
Sorry, one last nitpick - I think the `or` in the inverted conditional needs to be an `and`.

lua/packer/luarocks.lua
akinsho: Sorry that's right, pushed the change in a bit of a rush last night

wbthomason: No worries and thanks again!

wbthomason (approved):
Looks great, thank you very much!

PR #236, opened 2021-02-26
auto add * in event

glepnir:
no need to write '*' in event keyword.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ return require('packer').startup(function()
   use {'tpope/vim-dispatch', opt = true, cmd = {'Dispatch', 'Make', 'Focus', 'Start'}}
 
   -- Load on an autocommand event
-  use {'andymass/vim-matchup', event = 'VimEnter *'}
+  use {'andymass/vim-matchup', event = 'VimEnter'}
 
   -- Load on a combination of conditions: specific filetypes or commands
   -- Also run code after load (see the "config" key)
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -182,9 +182,12 @@ local function make_loaders(_, plugins)
       if plugin.event then
         loaders[name].only_sequence = false
         loaders[name].only_setup = false
-        if type(plugin.event) == 'string' then plugin.event = {plugin.event} end
+        if type(plugin.event) == 'string' then plugin.event = {plugin.event .. ' *'} end
 
         for _, event in ipairs(plugin.event) do
+	  if event:sub(#event,-1) ~= '*' then
+            event = event ..' *'
+	  end
           events[event] = events[event] or {}
           table.insert(events[event], quote_name)
         end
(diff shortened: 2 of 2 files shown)


akinsho:
I wonder if this change doesn't make things much less versatile? I mean I always just use `Event *` but I guess people could otherwise do really custom things like `BufRead .bashrc` or `SessionLoadPost custom_thing` etc. and this would take that ability away entirely 🤷🏿

glepnir:
this behavior like dein `on_event` .

akinsho:
@glepnir its similar but I'm just pointing out that this change trades off customisability, which I guess in some cases might be a lot to lose to avoid having to type in ` *`.

Imagine I had some `zsh` plugin that I only used for editing my `.zshrc` which might be something I do a lot and I wanted to only load that plugin just for `.zshrc` currently I can do `BufRead .zshrc` and this would only load that plugin if I was ever to enter my zshrc. But with this change I would have to always load it on `BufRead`. I could also try `ft = {'zsh'}, event = {'BufRead'}` but that still wouldn't be exactly equivalent.

Anyway it's ultimately up to @wbthomason when he's free, think he's quite busy atm.

glepnir:
I don't think it makes sense. ft is more useful in this situation. dein has been around for a long time. The `on_event` keyword has not been changed to prove that this requirement is not practical

akinsho:
This is just my 2 cents  as I noted this is one for wbthomason ultimately (butting out now) but I don't think it working just like `dein` should be the objective 🤷🏿‍♂️ especially if there's an opportunity to provide some extra value. To be clear it's not just about `filetype` you can do quite a lot of stuff
```lua
use {"someone/plugin", event = {"BufRead *$ENVIRONMENT_VAR/*subdir*/*.txt*"}}
```
To say only load a particular plugin when you are in a specific directory, there are a lot of possibilities.

glepnir:
you can use `cond` ? I think this also can do it.

akinsho:
I think `cond` is only run once when you enter neovim, whereas an autocommand can get triggered at any point during the running vim session, so it wouldn't do the same thing

glepnir:
It should not be just hypothesis. Specific examples can illustrate the problem. Instead of some/plugin. I don't know what plugin only works for txt?

nanotee:
Just as a heads up, I have the exact use case that @akinsho described:

```lua
use {
    'lervag/wiki.vim',
    event = 'BufReadPre ~/Documents/wiki/*.md'
}
```

I'm happy to find another solution if this PR makes `packer` easier to use, but I do like the flexibility of the current API :)

glepnir:
Now both work.

```lua
'BufReadPre','BufReadPre *','BufReadPre ~/Documents/wiki/*.md'
```

wbthomason (approved):
Thanks all for the contribution and conversation! While I agree with @akinsho and @nanotee that removing the ability to specify an autocommand pattern would be bad, I'm happy with this change to allow both patterns and elided *. Thanks again!

PR #238, opened 2021-02-28
Fix compilation when there are multple rtp settings

jdelkins:
When there are multiple plugins defined with `rtp`, the compiler was concatenating the path strings directly. This is a simple fix to add the string concatenation operator.


wbthomason:
Thanks for the fix! I missed this when I moved this step from happening in Vimscript to using Lua.

wbthomason (approved):

PR #246, opened 2021-03-03
Enable config merging

runiq:
If a plugin is `use`d multiple times, the configs are merged to the best
of our abilities:

- disable: Warn & ignore on disabling a previously enabled plugin
- branch, tag, commit, lock: Must all be the same, else warning
- opt: Lots of song and dance, see large comment in `merge_specs` for
  details. The gist of it is, opt = true < nil < false.
- config, setup: Additional lines/functions are appended.
- installer, updater: Only original value is used. This is required for
  `make_logging_callback` to work (since it updates the original table
  with results).
- run, rtp: Keep old key if it's set, otherwise use new one.
- cmd, cond, event, ft, keys: Append (dedup is done in `compile`).
- require, after: Handled in `manage`/`compile`.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -13,6 +13,7 @@ local config_defaults = {
   snapshot_path = join_paths(stdpath 'cache', 'packer.nvim'),
   package_root = join_paths(stdpath 'data', 'site', 'pack'),
   compile_path = join_paths(stdpath 'config', 'plugin', 'packer_compiled.lua'),
+  merge_duplicates = false,
   plugin_package = 'packer',
   max_jobs = nil,
   auto_clean = true,
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
   end
 end
 
+-- Merges new into old; returns merged spec.
+local function merge_specs(old, new)
+  local name = new.short_name
+
+  -- disable: Warn & ignore on disabling a previously enabled plugin
+  if old.disable then
+    if not new.disable then log.warn('Trying to disable previously non-disabled mod "' .. name .. '", ignoring') end
+  else
+    old.disable = new.disable
+  end
+
+  -- branch, tag, commit, lock: Must all be the same, else warning
+  for _, key in ipairs{'branch', 'commit', 'lock', 'tag'} do
+    if new[key] ~= old[key] then
+      log.warn('Plugin "' .. name .. '" wants to overwrite "' .. key .. '" setting. Keeping original value')
+    end
+  end
+
+  -- opt
+  -- - Conditions and opt_default have already been handled in manage()
+  -- - Rule of thumb: true < nil < false. Set the lowest necessary--if nil suffices, don't set false
+  -- - For manual_opt:
+  --   - If result is start, remove mo. Else, if omo or nmo, keep mo.
+  --   - There is no situation where opt = falsy but manual_opt = true
+  --
+  -- old new result (1 = true, - = nil, 0 = false, s = start, o = opt)
+  -- ---+---+------
+  --   -   -     -s
+  --   1   -     -s
+  --   -   1     -s
+  --   0   0     0s
+  --   1   0     0s
+  --   0   1     0s
+  --   -   0     0s
+  --   0   -     0s
+  --   1   1     1o
+  local old_opt = old.opt
+  local new_opt = new.opt
+  if old_opt == false or new_opt == false then
+    old.opt = false
+    old.manual_opt = nil
+  elseif old_opt == nil or new_opt == nil then
+    old.opt = nil
+    old.manual_opt = nil
+  elseif old_opt == true and new_opt == true then
+    if old.manual_opt or new.manual_opt then old.manual_opt = true end
+  end
+
+  -- config, setup: append
+  for _, key in ipairs{'config', 'setup'} do
+    if new[key] then
+      if type(old[key]) ~= 'table' then old[key] = {old[key]} end
+      if type(new[key]) ~= 'table' then new[key] = {new[key]} end
+      for _, item in ipairs(new[key]) do
+        table.insert(old[key], item)
+      end
+    end
+  end
+
+  -- installer, updater: Warn & ignore
+  for _, key in ipairs{'installer', 'updater'} do
+    if new[key] then
+      log.warn('Plugin "' .. name .. '" wants to overwrite "' .. key .. '" setting. Keeping original value')
+    end
+  end
+
+  -- run, rtp: Keep old key if it's set, otherwise set new one
+  for _, key in ipairs{'run', 'rtp'} do
+    if new[key] then
+      if old[key] then
+        local old_type = type(old[key])
+        local new_type = type(new[key])
+        if (old_type ~= new_type) or
+            -- If the first condition doesn't hit, old and new have the same type
+            (old_type == 'string' and old[key] ~= new[key]) or
+            (old_type == 'function' and (string.dump(old[key], true) ~= string.dump(new[key], true)))
+        then
+          log.warn('Plugin "' .. name .. '" wants to overwrite "' .. key .. '" setting. Keeping original value')
+        end
+      else
+        old[key] = new[key]
+      end
+    end
+  end
+
+  -- cmd, cond, event, ft, keys: append (dedup is done in compile)
+  for _, key in ipairs{'cmd', 'cond', 'event', 'ft', 'keys'} do
+    if new[key] then
+      if type(old[key]) ~= 'table' then old[key] = {old[key]} end
+      if type(new[key]) ~= 'table' then new[key] = {new[key]} end
+      for _, item in ipairs(new[key]) do
+        if not vim.tbl_contains(old[key], item) then
+          table.insert(old[key], item)
+        end
+      end
+    end
+  end
+
+  -- require, after: Handled in manage()/compile()
+  return old
+end
+
 --- The main logic for adding a plugin (and any dependencies) to the managed set
 -- Can be invoked with (1) a single plugin spec as a string, (2) a single plugin spec table, or (3)
 -- a list of plugin specs
@@ -206,7 +309,7 @@ manage = function(plugin_data)
     return
   end
 
-  if plugins[name] and not plugins[name].from_requires then
+  if plugins[name] and not plugins[name].from_requires and not config.merge_duplicates then
     log.warn('Plugin "' .. name .. '" is used twice! (line ' .. spec_line .. ')')
     return
   end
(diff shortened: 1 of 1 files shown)


lua/packer.lua
```diff
@@ -188,13 +291,19 @@ manage = function(plugin)
     end
   end
 
-  plugin.install_path = util.join_paths(plugin.opt and config.opt_dir or config.start_dir,
```
runiq, line 191: This chunk maybe looks more complicated than it actually is—All I'm doing here is to ask, "has this already been `use`d before? If not, do <whatever packer did before>. If we've seen it before, skip the `setup` stuff (since it's already been done the first time around) and merge the new spec into the old one".

lua/packer.lua
```diff
@@ -188,13 +291,19 @@ manage = function(plugin)
...
 
-  if not plugin.type then plugin_utils.guess_type(plugin) end
-  if plugin.type ~= plugin_utils.custom_plugin_type then plugin_types[plugin.type].setup(plugin) end
-  for k, v in pairs(plugin) do if handlers[k] then handlers[k](plugins, plugin, v) end end
-  plugins[plugin.short_name] = plugin
+    plugins[plugin.short_name] = plugin
+  end
   if plugin.rocks then use_rocks(plugin.rocks) end
 
   if plugin.requires and config.ensure_dependencies then
```
runiq, line 309: This entire chunk maybe looks more complicated than it actually is—all I'm doing here is to ask, "has this already been `use`d before? If not, do \<whatever packer did before\>. If we've seen it before, skip the `setup` stuff (since it's already been done the first time around) and merge the new spec into the old one".

wbthomason:
@runiq No worries if you're not working on this at the moment, but just so you know, I'm ignoring this PR while it's marked as draft/unless you ping me with a question (sorry if you've already done this and I forgot!)

runiq:
I'm not exactly *working* on it, but I'm dogfooding! :) The PR itself seems 'done' to me. Haven't unearthed anything serious so far, I believe. As soon as I've had a chance to go over it again (probably next week), I'll move it out of draft.

wbthomason:
@runiq Please note that, while there's zero rush, I am waiting on you to move this out of draft before I take a serious look at it. Again, no rush from my end, though!

runiq:
Alright, let's pull the plug and move this out of draft. I've tried this out for long enough, I believe, and I've only found minor issues that I believe I've fixed in the meantime.

I'm still not too sure about the handling of some of the keys. Some of it feels pretty ad-hoc to me now. You're the plugin author, @wbthomason, so if you have ideas on how to improve the consistency here, I'm all ears! :)

runiq:
Ping! :) Is there still desire to get this PR merged? I've been using it in my (admittedly fairly static) config for a few months and kept up with the changes to packer, which were fairly easy to integrate. I can still understand if you don't want to merge this :)

runiq:
Alright, so I finally figured out how to do this in my config without packer's involvement. And I assume this isn't as useful to others as I though it'd be, so… How about I just close the PR? One less thing for you to worry about. :)

wbthomason:
@runiq Hey! Sorry, for some reason I still had this in my mental queue as "draft"; I must've missed your message forever ago.

I think this would actually still be useful to others - config merging would make some requested features for plugins that (partially) specify other plugins as requirements easier to use. I'll give it a review now. Thank you very much for your persistence with this, and I'm really sorry that I dropped the ball on review so hard here!

lua/packer.lua
```diff
@@ -13,6 +13,7 @@ local config_defaults = {
   snapshot_path = join_paths(stdpath 'cache', 'packer.nvim'),
   package_root = join_paths(stdpath 'data', 'site', 'pack'),
   compile_path = join_paths(stdpath 'config', 'plugin', 'packer_compiled.lua'),
+  merge_duplicates = false,
```
wbthomason, line 16: Nitpick: maybe this should be slightly more descriptive, e.g. `merge_duplicate_configs`? Or is that too verbose?

lua/packer.lua
```diff
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
   end
 end
 
+-- Merges new into old; returns merged spec.
+local function merge_specs(old, new)
+  local name = new.short_name
+
+  -- disable: Warn & ignore on disabling a previously enabled plugin
+  if old.disable then
+    if not new.disable then log.warn('Trying to disable previously non-disabled mod "' .. name .. '", ignoring') end
```
wbthomason, line 177: NItpick: this should be 'Trying to enable previously non-disabled plugin "'

lua/packer.lua
```diff
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
   end
 end
 
+-- Merges new into old; returns merged spec.
+local function merge_specs(old, new)
+  local name = new.short_name
+
+  -- disable: Warn & ignore on disabling a previously enabled plugin
```
wbthomason, line 175: I may answer this question for myself further down the file, but: how is the old/new iteration order determined? If it's unstable, e.g. Lua table iteration order, then this could cause inconsistent behavior, since we only check the "old" version of the spec. That said, the inconsistency is only whether or not a warning gets printed, so it's not that big of a problem.

lua/packer.lua
```diff
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
...
+  if old.disable then
+    if not new.disable then log.warn('Trying to disable previously non-disabled mod "' .. name .. '", ignoring') end
+  else
+    old.disable = new.disable
+  end
+
+  -- branch, tag, commit, lock: Must all be the same, else warning
+  for _, key in ipairs{'branch', 'commit', 'lock', 'tag'} do
+    if new[key] ~= old[key] then
+      log.warn('Plugin "' .. name .. '" wants to overwrite "' .. key .. '" setting. Keeping original value')
```
wbthomason, line 185: Is it worth printing the two values in this warning, e.g. to make determining which spec is "new" easier for users?

lua/packer.lua
```diff
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
...
+  -- - Rule of thumb: true < nil < false. Set the lowest necessary--if nil suffices, don't set false
+  -- - For manual_opt:
+  --   - If result is start, remove mo. Else, if omo or nmo, keep mo.
+  --   - There is no situation where opt = falsy but manual_opt = true
+  --
+  -- old new result (1 = true, - = nil, 0 = false, s = start, o = opt)
+  -- ---+---+------
+  --   -   -     -s
+  --   1   -     -s
+  --   -   1     -s
```
wbthomason, line 200: I'm somewhat confused as to why these lines have the same result. For example, couldn't we have a case where `old.opt = true` because the "original" spec sets up some lazy-loading, but `new.opt = nil` because the new spec just customizes e.g. `config` and wants to leave the original lazy-loading intact?

lua/packer.lua
```diff
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
...
+  --   0   0     0s
+  --   1   0     0s
+  --   0   1     0s
```
wbthomason, line 203: Same question here - what if `new.opt = true` because it's adding some lazy-loading?

lua/packer.lua
```diff
@@ -167,6 +168,108 @@ packer.use_rocks = function(rock)
...
+  -- run, rtp: Keep old key if it's set, otherwise set new one
+  for _, key in ipairs{'run', 'rtp'} do
+    if new[key] then
+      if old[key] then
+        local old_type = type(old[key])
+        local new_type = type(new[key])
+        if (old_type ~= new_type) or
+            -- If the first condition doesn't hit, old and new have the same type
+            (old_type == 'string' and old[key] ~= new[key]) or
+            (old_type == 'function' and (string.dump(old[key], true) ~= string.dump(new[key], true)))
```
wbthomason, line 246: Does direct function comparison not work here? I suppose users are most likely to specify anonymous functions which will compare unequal, but unless the user has specified exactly the same function in the new spec as the old, the bytecode will compare unequal too.

lua/packer.lua
```diff
@@ -206,7 +309,7 @@ manage = function(plugin_data)
     return
   end
 
-  if plugins[name] and not plugins[name].from_requires then
+  if plugins[name] and not plugins[name].from_requires and not config.merge_duplicates then
```
wbthomason, line 312: Maybe we should emit a `debug` level log message even if we are merging duplicates, in case of accidental override?

wbthomason: Nevermind; I now see the log call on line 352.

wbthomason (changes requested):
Thanks for this! I think it's mostly good to go, except for a few nitpick comments and some confusion over the right thing to do with merging `opt`ness.

I also want to test the assumption that appending to `config`, `setup`, etc. (for the other compile-relevant keys) is correct - the compile step can handle multiple such values, but what if users legitimately want to override these, rather than just adding to them?

Finally, do you think there's value in adding an `override` key to the spec to allow merging to be skipped at the individual spec level?

lua/packer.lua
runiq: Yup, will change that!

lua/packer.lua
runiq: Yeah, it'd definitely be useful. Last night I finally understood what the `plugin_spec.line` and IIUC we could use it to highlight the offending lines here.

lua/packer.lua
runiq: Ah, this is probably a blind spot in my thinking, since I generally give packer several lists of plugin specs in the same order (and lists themselves are ordered), so I'm getting deterministic behavior.

I'll have to noodle on this a bit, since it may have an effect on the other options as well…

lua/packer.lua
runiq: That would definitely make sense, yeah.

And that reveals one of the major pain points I've had while deciding on the policies for each setting: Sometimes more than one behavior may make sense, depending on the use-case, but I had to choose one for this PR. That only became apparent to me when I moved `merge_specs()` over to my own config, where deciding on those policies was surprisingly easy since I was the final arbiter on how these settings should behave.

This was actually another reason for me wanting to close this PR (one that I should have expressed explicitly, I apologize): In its current form, once you activate the `merge_duplicates` key, it might result in behavior that doesn't fit your use-case, and you cannot override it without another PR. This wouldn't be a problem if one could override how a key is handled, but that is blocked on migrating all the setting keys to handlers.

I feel this is a pretty big deal, so maybe we should in fact wait with merging this PR until the migration of the "official" keys to handlers has been made?

lua/packer.lua
runiq: Oof, I totally missed that, thank you. I completely forgot that a lot of keys implicitly enable `opt`.

lua/packer.lua
runiq: True, I'll change it.

runiq:
> I also want to test the assumption that appending to `config`, `setup`, etc. (for the other compile-relevant keys) is correct - the compile step can handle multiple such values, but what if users legitimately want to override these, rather than just adding to them?

I'm actually not sure anymore! :) See my [inline comment](https://github.com/wbthomason/packer.nvim/pull/246#discussion_r833165460).

> Finally, do you think there's value in adding an `override` key to the spec to allow merging to be skipped at the individual spec level?

I would rather wait for the migration of the official keys to handlers—it'd be just as powerful and much easier to conceptualize. I've had to read the code to understand how some of the keys interact, and adding a not-quite-orthogonal way to override would make this even more complicated.

lua/packer.lua
wbthomason: I think we might actually be OK here - the only place this is called is within `manage`, and `manage` is only called within an `ipairs` loop, which should have a deterministic iteration order (since - afaik - plugins are always added with `use` in a deterministic order since they have to be given either as separate calls or in an arraylike table).

lua/packer.lua
wbthomason: That's a fair point, and if this PR no longer serves your needs, I understand if you don't want to dedicate more time to it - I should be able to take it over from here, if you'd prefer.

That said, one workaround is the `override` key I mentioned in another comment. Or, we can wait until I finally finish the refactoring to make everything handlers - maybe this will serve as good motivation for me to move that PR along....

PR #248, opened 2021-03-04
Lazy load package on require

### LINKED ISSUE #108: Lazy load on require Lua module - possible?
Hey,
this plugin manager looks pretty cool! I like the native package support with all those fancy features on top. :+1: 

I'm always very interested into lazy loading functionality. I simply have a too big setup and use `nvim` too often for many small things where I simply don't need everything. With the rise of Lua plugins, key mappings that require a Lua module and execute a function becomes more and more common. So I was wondering if the good old lazy load on command/function call could also work for such cases? I saw that there is the feature to lazy load on keys. But that creates a (horrible - at least for this case) dependency that I would love to prevent under all circumstances.

Since I basically have too less knowledge about the internal details, do you know/think it is possible to make such a feature possible? Thanks for your feedback!

akinsho:
This PR borrows heavily from dein's implementation of `on_lua` to lazy load a vim package if its lua module is required e.g. 
```lua
-- somewhere in init.lua
use {'akinsho/nvim-bufferline', module = 'bufferline'}
require('bufferline').setup{}
```

#### Todo
- [ ] Should this take a list optionally? why? there's usually just one package name for a lua module and one vim plugin that corresponds to is @wbthomason thoughts?
- [ ] Use `load` function maybe. Currently `packer_load` expects a list of plugins and seems to be designed to load multiple plugins at a time, is it suited to this use case?

Will fix #108

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -56,7 +56,7 @@ local function lazy_load_module(module_name)
 end
 
 if not vim.g.packer_custom_loader_enabled then
-  table.insert(package.loaders, 1, lazy_load_module)
+  table.insert(package.loaders, 2, lazy_load_module)
   vim.g.packer_custom_loader_enabled = true
 end
 ]]
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -138,6 +138,7 @@ local function make_loaders(_, plugins)
   local after = {}
   local fns = {}
   local ftdetect_paths = {}
+  local module_lazy_loads = {}
   for name, plugin in pairs(plugins) do
     if not plugin.disable then
       local quote_name = "'" .. name .. "'"
@@ -278,6 +279,13 @@ local function make_loaders(_, plugins)
           fns[fn] = fns[fn] or {}
           table.insert(fns[fn], quote_name)
         end
+      end
+
+      if plugin.module then
+        loaders[name].only_sequence = false
+        loaders[name].only_setup = false
+        if type(plugin.module) == 'string' then plugin.module = {plugin.module} end
+        for _, module_name in ipairs(plugin.module) do module_lazy_loads[module_name] = name end
       end
 
       if plugin.config and (not plugin.opt or loaders[name].only_setup) then
@@ -457,6 +465,12 @@ then
   if rtp_line ~= '' then
     table.insert(result, '-- Runtimepath customization')
     table.insert(result, rtp_line)
+  end
+
+  -- Then the module lazy loads
+  if next(module_lazy_loads) then
+    table.insert(result, 'local module_lazy_loads = ' .. vim.inspect(module_lazy_loads))
+    table.insert(result, module_loader)
   end
 
   if next(setup_lines) then vim.list_extend(result, setup_lines) end
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -46,13 +46,19 @@ local function lazy_load_module(module_name)
   local to_load = {}
   local i = 1
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if string.match(module_name, "^" .. module_pat) then to_load[i] = plugin_name end
+    -- check the plugin has not already been loaded to avoid an infinite loop
+    if string.match(module_name, "^" .. module_pat) and not _G.packer_plugins[plugin_name].loaded then
+      to_load[i] = plugin_name
+    end
   end
 
   require('packer.load')(to_load, {module = module_name}, _G.packer_plugins)
 end
 
-table.insert(package.loaders, 1, lazy_load_module)
+if not vim.g.packer_custom_loader_enabled then
+  table.insert(package.loaders, 1, lazy_load_module)
+  vim.g.packer_custom_loader_enabled = true
+end
 ]]
 
 local function dump_loaders(loaders)
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -207,8 +207,8 @@ local function make_loaders(_, plugins)
         end
 
         for _, event in ipairs(plugin.event) do
-          if event:sub(#event,-1) ~= '*' and not event:find('%s') then
-            event = event ..' *'
+          if event:sub(#event, -1) ~= '*' and not event:find('%s') then
+            event = event .. ' *'
           end
           events[event] = events[event] or {}
           table.insert(events[event], quote_name)
@@ -272,9 +272,7 @@ local function make_loaders(_, plugins)
       if plugin.fn then
         loaders[name].only_sequence = false
         loaders[name].only_setup = false
-
         if type(plugin.fn) == 'string' then plugin.fn = {plugin.fn} end
-
         for _, fn in ipairs(plugin.fn) do
           fns[fn] = fns[fn] or {}
           table.insert(fns[fn], quote_name)
@@ -473,6 +471,7 @@ then
     table.insert(result, module_loader)
   end
 
+  -- Then setups, configs, and conditionals
   if next(setup_lines) then vim.list_extend(result, setup_lines) end
   if next(config_lines) then vim.list_extend(result, config_lines) end
   if next(conditionals) then
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -46,7 +46,7 @@ local function lazy_load_module(module_name)
   local to_load = {}
   local i = 1
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if string.match(module_name, module_pat) then to_load[i] = plugin_name end
+    if string.match(module_name, "^" .. module_pat) then to_load[i] = plugin_name end
   end
 
   require('packer.load')(to_load, {module = module_name}, _G.packer_plugins)
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -510,6 +510,6 @@ end
 
 local compile = setmetatable({cfg = cfg}, {__call = make_loaders})
 
-compile.opt_keys = {'after', 'cmd', 'ft', 'keys', 'event', 'cond', 'setup', 'fn'}
+compile.opt_keys = {'after', 'cmd', 'ft', 'keys', 'event', 'cond', 'setup', 'fn', 'module'}
 
 return compile
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -46,8 +46,7 @@ local function lazy_load_module(module_name)
   local to_load = {}
   local i = 1
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    -- check the plugin has not already been loaded to avoid an infinite loop
-    if string.match(module_name, "^" .. module_pat) and not _G.packer_plugins[plugin_name].loaded then
+    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. module_pat) then
       to_load[i] = plugin_name
     end
   end
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -40,6 +40,21 @@ local function try_loadstring(s, component, name)
 end
 ]]
 
+local module_loader = [[
+local function lazy_load_module(module_name)
+  if module_name == 'packer.load' then return nil end
+  local to_load = {}
+  local i = 1
+  for module_pat, plugin_name in pairs(module_lazy_loads) do
+    if string.match(module_name, module_pat) then to_load[i] = plugin_name end
+  end
+
+  require('packer.load')(to_load, {module = module_name}, _G.packer_plugins)
+end
+
+table.insert(package.loaders, 1, lazy_load_module)
+]]
+
 local function dump_loaders(loaders)
   local result = vim.deepcopy(loaders)
   for k, _ in pairs(result) do
--- a/README.md
+++ b/README.md
@@ -329,6 +329,7 @@ use {
   fn = string or list          -- Specifies functions which load this plugin.
   cond = string, function, or list of strings/functions,   -- Specifies a conditional test to load this plugin
   setup = string or function,  -- Specifies code to run before this plugin is loaded.
+  module = string or list      -- Specifies patterns (e.g. for string.match) of Lua module names which, when required, load this plugin
 }
 ```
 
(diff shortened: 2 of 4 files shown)


akinsho:
@wbthomason I could use some advice regarding how to actually load the plugin, should I just do a plain `packadd` seems like we probably want to do all the extra magic like loading after dirs etc but the load function isn't really designed for use with 1 plugin and also makes assumptions about things like `config` being a string etc. I've tried tweaking it with some extra conditionals etc but to me this seems best as 2 separate functions that might both depend on some helper functions.

An issue with that though is that `load` doesn't export a module just the wrapper function so I'd have to change all it's usages.

Alternatively I could just duplicate the subset of load logic I need in `packer.lua` but that seems messy to me.

I wanted to check what your preference would be before going down any specific path.

wbthomason (commented):
Thanks for starting on this! The `load` function should be fine with a single plugin, afaik...additionally, I think the "config is a string" assumption should still be OK here.

What I'd envision is that the loader - which has access to the `packer_plugins` table - would invoke the `load` function with a similar table to the uses by the other lazy-loaders, which should suffice for `load` to work as usual. If you'd like, I'd be happy to push a commit or two to this branch to implement that. Or, if you think that's not a good design, I'm open to alternatives - this just seems like the simplest/least redundant/most consistent way to implement the other load checks (`after` dirs, etc.)

akinsho:
@wbthomason if it works I'm happy to go with using the existing loader I'm not sure how you'd get round the loadstring for config values without either some conditional logic or something. Anyway feel free to push commits to this branch 👍🏾

wbthomason:
@akinsho I pushed a few commits, if you have a chance to take a look. The current code is tested to work with your `bufferline` example; I've not tested it much beyond that yet.

The gist of my changes are:
- Move the lazy-loading to happen in packer_compiled, since we can't assume that packer itself has been loaded
- Simplify the lazy-load logic by (1) allowing plugin specs to give a Lua pattern to match (since this includes the common use case of a module name, but allows matching richer categories of module names e.g. in the case of a library like plenary) and (2) simply collecting the matching plugin names and passing them to `packer.load`
- I also added code to make a "short list" table of the modules that are to be lazy-loaded, to minimize the overhead on overall package loading (since this custom loader will run for every `require`)

Please let me know what you think!

lua/packer/compile.lua
```diff
@@ -40,6 +40,21 @@ local function try_loadstring(s, component, name)
...
+  local to_load = {}
+  local i = 1
+  for module_pat, plugin_name in pairs(module_lazy_loads) do
+    if string.match(module_name, "^" .. module_pat) then to_load[i] = plugin_name end
+  end
+
+  require('packer.load')(to_load, {module = module_name}, _G.packer_plugins)
+end
+
+table.insert(package.loaders, 1, lazy_load_module)
```
akinsho, line 55: Since the compiled file can be run several times maybe this should be inside a guard statement to prevent it from being re-run, so it doesn't end up adding several loaders

akinsho:
@wbthomason I hadn't thought about packer itself not being loaded. Makes sense to move it to the compiled file. I just added one comment about how it would work if you resource the packer compiled file. Looks good overall. Haven't had a chance to test it yet though

wbthomason:
Ah, good point re: reloading. We'll have to add a check to prevent the custom loader from being added multiple times.

akinsho:
Just tried this out @wbthomason and realised that it's missing a check that I noticed in the dein implementation [here](https://github.com/Shougo/dein.vim/commit/c6ddf04183606413870a0f8a86f61f8d58f85982#diff-4f851e2cf75e336d1775d0ecad81fb006dd85e4cb6853de7fef17edcf4868509R164). There's a check to prevent an infinite loop which I'm hitting whilst running this. I've tried to add a conditional that prevents attempting to reload a plugin if already loaded but I think maybe loaded isn't being set to true since this condition doesn't work
```lua
    -- check the plugin has not already been loaded to avoid an infinite loop
    if string.match(module_name, "^" .. module_pat) and not _G.packer_plugins[plugin_name].loaded then
      to_load[i] = plugin_name
    end

```
Not sure if I'm missing something about how to track when a plugin has already been required, not I also tried similart to dein's implementation to use a local table of loaded modules but that also seems not to work 🤔

wbthomason:
Ah, hmm. I'll take a look when I'm back at a computer, but one thought is that perhaps we should put the custom loader at position 2 in the list, since the first loader is responsible for checking for modules that are already loaded? That seems better for performance and like it could fix the infinite loop issue.

akinsho:
Actually the conditional does work to stop the infinite loop I was getting thrown by the fact that the package loader function gets called for all submodules e.g. `bufferline`, `bufferline/buffers`, `bufferline/colors` etc so it generally results in several similar/repeated errors. Although the original/main problem remains which is that it actually doesn't load the module. I believe it was working in the initial implementation so not sure if something about how it works now is incompatible 🤔

akinsho:
Ahhh I think I understand what's happening now based on my example. In my case I'm testing my own plugin `bufferline` which I use some utils it contains in some parts of my `init.lua`. I was wondering why I was still hitting a module not loaded error when I wasn't before but I believe this is because adding the custom loader is now in `packer_compiled.vim` which hasn't run yet so the loader hasn't been added to the loaders till *after* my init.lua is loaded which means this functionality wouldn't work for any requires inside an `init.lua`. 

I guess if someone use a require in a mapping then this would work but that seems a lot less useful to me since surely you'd usually have the require as part of some lua code you'd already written in your init.lua?

wbthomason:
The point about not having access in `init.lua` is unfortunately fair. If you are organizing your dotfiles in the way `packer` encourages (e.g. mostly via `setup` and `config` functions, etc.), then this isn't an issue, but plenty of people don't do that.

To be fair, though, if you're using this lazy-loader for a plugin that you need in your `init.lua`, then you won't get any benefit from it anyway - you may as well just make the plugin a `start` plugin, right? Since you would effectively need it loaded immediately upon start anyway?

akinsho:
@wbthomason I guess that makes sense, although I think it might be quite confusing to users. Since the specific reason it won't work vs say `dein` is that it happens in the compiled file which a user might not realise, and is sort of an implementation detail of packer. Currently it errors quite messily if you make the mistake of using `module = <name>` if packer compiled hasn't been loaded yet.

I'm wondering what the use cases for this would be since other than a mapping or autocommand I'm not sure I know why you'd use this. The usages I imagined were along the lines of say a plugin is conditionally loaded and it in turn requires another plugin that is also conditionally loaded then if the condition is true and it wanted to then require the dependent plugin's module that would work via specifying the module.

So for example you could do
```lua
use {"plenary", module = "plenary"}
use {"plugin-needs-plenary", config = some_conf_fn} -- this will require plenary on load
```

This is potentially a bit contrived as there may be other ways around this but it does seem like a strange limitation for it to not work during initialisation especially if you wanted to module key to be used for plugins that may or may not  be loaded during startup

wbthomason:
That's fair - though I imagine the `dein` version of this feature also only works once your `dein` setup file has been sourced?

One of the reasons I hadn't prioritized this feature earlier is that I think it is more limited in utility than it seems on its face. *All* Lua code is effectively "loaded on demand" anyway, since it isn't read until it's required. As such, there's no startup benefit to making a purely Lua plugin `opt` - only to deferring the use of expensive functions from that plugin, e.g. when an autocmd, keybinding, etc. is triggered.

The `module` lazy-load, then, seems almost redundant. Your `plenary` example is still reasonable since `plenary` includes files in `plugin` which impose a startup cost (though I've made a PR to optionally disable the main culprit via https://github.com/nvim-lua/plenary.nvim/pull/77) - I think plugins like this are the primary motivation for `module` lazy-loading (i.e. plugins which mix Lua and non-autoloaded Vimscript).

All this to say: I think the "right" way to handle this is still to put things in `config`/`setup` keys, and I don't think the limitations are actually all that different from `dein` (e.g. one could manually source `packer_compiled` at the start of their init file). But, I acknowledge that the cases you suggest are realistic and these restrictions are likely to confuse users.

Given that...I wonder if an appropriate solution (to this and other packages-related problems) is to change the architecture of `packer` a bit, such that:
- we add a `packer.end` (similar to RTP-based plugin managers) which will source `packer_compiled` if it exists and run `packer.compile` if it's missing
- `packer.startup` does the same thing
- The main `packer` module is refactored to be super cheap to `require`, e.g. by lazily-requiring the modules providing other functionalities

If we opt for these changes, I might also add a default-true config var that basically means "my plugin specification is all in the file that `init`/`startup` was called from". That enables us to do things like auto-recompile if `packer_compiled` is out of sync with the specification file, while still letting more advanced users who split their specification across multiple files do their thing.

This would mean that (1) `packer_compiled` gets moved out of `plugin/` by default and (2) users are expected to call `packer.init` or `packer.startup` somewhere in their `init` file. That's kind of a big breaking change, but it would enable an overall smoother use of `packer`, I think.

akinsho:
This is a bit of a tangled thought around improvements loading lua plugins. With some lua plugins I use/have encountered I believe it's becoming common to do a bit of potentially expensive setup logic in the `require"plugin".setup_or_other_name{}` calls which I think is where some of the startup cost is being seen (usually not excessive but they add up). Not sure I know exactly when the `config` is run but if it occurs during startup maybe that could be a place worth optimising. I've definitely observed that deferring plugins like `nvim-lspconfig` to `VimEnter` improves my startup time.

Regarding `dein` tbh I hadn't realised that it also depended on a compiled file, don't know much about how it handles that, I guess if it's similar then the issues are the same.

I think if there is something that `module = <name>` provides that can't be accomplished some other way then there is still value in this. Maybe the `autocmd BufEnter require('thing').method()` case etc. I think even if it couldn't be in the `init.lua` that could be documented but personally I think things should work the easy way rather than requiring users to learn about details they likely don't care about.

Regarding a larger refactor
> we add a packer.end (similar to RTP-based plugin managers) which will source packer_compiled if it exists and run packer.compile if it's missing

At what point would it then be sourced? before the configs are read?
What are the main issues it would solve? it would help here but the value of this seems less clear, and with things like using `_G.packer_plugins` in an `init.lua` but there's a not a great need for  that using `config/setup`?

>If we opt for these changes, I might also add a default-true config var that basically means "my plugin specification is all in the file that init/startup was called from". That enables us to do things like auto-recompile if packer_compiled is out of sync with the specification file, while still letting more advanced users who split their specification across multiple files do their thing.

How might this work? Why for only users with everything in one file? seems odd to lose functionality due to that? if it has something to do with knowing where to look to know if things are out of sync then perhaps packer could provide a wrapper around require for config like `packer.config(<path>)` or potentially change the behaviour of require in the context of the packer startup function.

>users are expected to call packer.init or packer.startup somewhere in their init file. That's kind of a big breaking change

Is this something many people aren't doing? I'm vaguely aware that this exists but is it normal for people to use it this way? In either case potentially packer could print a warning for a few days in the lead up to the change and pin an issue in the repo

wbthomason:
Thanks for your thoughts!

> With some lua plugins I use/have encountered I believe it's becoming common to do a bit of potentially expensive setup logic in the require"plugin".setup_or_other_name{} calls which I think is where some of the startup cost is being seen (usually not excessive but they add up). Not sure I know exactly when the config is run but if it occurs during startup maybe that could be a place worth optimising

Yup, I agree. `config` runs when the plugin is loaded (so at startup for `start` plugins, on lazy-load for lazy-loaded plugins).

> Regarding dein tbh I hadn't realised that it also depended on a compiled file, don't know much about how it handles that, I guess if it's similar then the issues are the same.

I could well be wrong here (it's been some years since I last used `dein`), but iirc it relies on a "compiled" file with a precomputed `rtp` value and some other information. Rather than that, though, I more meant to say that `dein` users still have to call *something* to get loading on `require` to work, which is not so different a requirement from e.g. manually sourcing `packer_compiled` to get it to work.

> personally I think things should work the easy way rather than requiring users to learn about details they likely don't care about.

Agreed, and I think this is an area of weakness for `packer` - it can be a bit overwhelming to start with.

> At what point would it then be sourced? before the configs are read?
What are the main issues it would solve? it would help here but the value of this seems less clear, and with things like using _G.packer_plugins in an init.lua but there's a not a great need for that using config/setup?

Right - the main point of this would be making it so that at least some parts of `packer` are loaded during `init` (which we currently do not assume). That would make it easier to do things like `module` lazy-loading in `init`, but I think the biggest thing this would enable is automatic (re)generation of `packer_compiled` under certain circumstances, which would obviate 90% of the need for manual `PackerCompile` calls.

> How might this work? Why for only users with everything in one file? seems odd to lose functionality due to that? if it has something to do with knowing where to look to know if things are out of sync then perhaps packer could provide a wrapper around require for config like packer.config(<path>) or potentially change the behaviour of require in the context of the packer startup function.

As I've thought about it, this would work by adding a check for `mtime` to the start of `packer_compiled`, comparing the last compiled time to the `mtime` of the plugin specification file. An older compiled time would trigger a recompile. The issue with a single file vs. multiple (and I agree that it's not great to lose functionality if you opt for multiple files) is that `packer` needs to know what file to run in order to load the set of plugin specifications. If this is one file, we can assume that `require(<FILE PATH>)` or something will do it; if it's multiple files, we can load them all, but there might be a higher chance that those files have side effects that we don't want.

What automatic recompilation needs is just a way to get at the full current set of managed plugins. Thinking "out loud" here, `use` could certainly track the files from which it's called - the issue is really just side effects resulting from blindly requiring/running those files. The easy way out of this is requiring plugin specification files to not have side effects, but idk if that's too restrictive for some uses. That's where I get stuck - on a way to run all the `use` calls without getting side effects from other code in the same file.

> Is this something many people aren't doing? I'm vaguely aware that this exists but is it normal for people to use it this way? In either case potentially packer could print a warning for a few days in the lead up to the change and pin an issue in the repo

Heh, I don't do this. I'm probably in the minority, though, and printing a warning is a reasonable way to handle this.

wbthomason:
More to the point of this particular issue/feature: I think the best way forward may be to merge with the limitation that it won't work in `init.lua`, and giving real consideration to the broader changes we've discussed here to enable (1) `init.lua` use of this feature and (2) automatic recompilation of the loader file.

akinsho:
Makes sense 👍🏿

wbthomason:
Thanks again for this! I made one small tweak to check if a plugin is loaded before looking for a pattern match, but I think it's good to go. I'll add some docs and then merge.

PR #259, opened 2021-03-19
Implement User autocmds

jdelkins:
This feature can enable automations that require knowing when an
asynchronous Packer operation completes. For example, to automatically
sync and quit nvim, you can do something like this:

```viml
function! SyncPluginsAndQuit()
  autocmd User PackerComplete qa
  PackerSync
endfunction
```


wbthomason (approved):
Thanks, this is a great idea! I did some minor refactoring to trigger the user autocommands in functions (in case we want to add more behavior to the "on completion" hooks), and updated the docs and README. If you're good with this, I'll go ahead and merge.

Thanks again!

jdelkins:
Perfect, that's awesome. Thanks!

Issue #262, opened 2021-03-22
packer_plugins not available globally


mjacobus:
Hi there

I got that error while trying the code snippet found in the README:

> E5113: Error while calling lua chunk: config/nvim/init.lua:254: attempt to index global 'packer_plugins' (a nil value)

```lua
-- init.lua
require('packer').startup(function(use)
--...
end)

-- ...

if packer_plugins["vim-fugitive"] and packer_plugins["vim-fugitive"].loaded then
print("Vim fugitive is loaded")
-- other custom logic
end
```

Am I missing something or is that a bug? I am new to lua so I might have missed something that is obvious to more experienced users.

akinsho:
@mjacobus `packer_plugins` is only available *after* `packer_compiled.vim` has been sourced so it can't be used anywhere that is loaded before packer compiled e.g. you can use it for things like autocommands when vim has already loaded or in `ftplugin` files etc.

mjacobus:
Thank you @akinsho for your help!

So far I did not quite understand what the packer_compiled.vim is and how/why to use it, but I will try to catch up with the entire README soon. I also need to understand how to do proper lua configuration.

I found [this](https://github.com/9710willy/dotfiles/tree/master/nvim) config and I think a few things started to click, just by looking at it. I think what I really need is to use the config option.

akinsho:
@mjacobus yeah with packer you really shouldn't need the `packer_plugins` table that much. Configuration options for plugins should be inside a function in the `config` key of the plugins declaration and that will automatically be loaded after a plugin. If you want to add thing before the plugin is loaded you use the `setup` key.

Yeah the README has this all in there but it's quite long and likely very commonly skimmed. I think packer's docs could do with a succinct quick start section maybe

wbthomason:
Hypothetically, https://github.com/wbthomason/packer.nvim#quickstart is that succinct quick start, but it could probably do with a rewrite. There's a lot to cover concisely...

akinsho:
😆 Forgot there was already a section called quickstart. @wbthomason I think since packer is a power tool more or less it's quite easy to get overwhelmed by the documentation. Re. the quick start I think like small code block with explanatory comments explaining the *most common* use cases might be more digestible. I think all the examples in the big block are useful but maybe not super scannable.

Something like

# Quick start

## A simple plugin installation
This will install the plugin when `PackerSync` is run and always load it on start up
```lua
[...]

PR #263, opened 2021-03-22
Add note about usage of packer_plugins table

akinsho:
@wbthomason this PR adds a small note about `packer_plugins` not being available till after `packer_compiled` is loaded since this I think has come up a few times e.g. #262


wbthomason (approved):
Thanks!

Issue #250, opened 2021-03-06
How to manage a dependency between two lazy load plugins?


weilbith:
Hey,
I have the following concrete example:

```vim
lua packer.use({
      \   'nvim-telescope/telescope.nvim',
      \   requires = {{'nvim-lua/popup.nvim'}, {'nvim-lua/plenary.nvim'}, {'kyazdani42/nvim-web-devicons'}},
      \   cmd = 'Telescope',
      \   config = 'vim.cmd("packadd telescope")',
      \ })
```

and 

```vim
lua packer.use({
      \   'pwntester/octo.nvim',
      \   requires = {{'nvim-lua/popup.nvim'}, {'nvim-lua/plenary.nvim'}},
      \   after = 'telescope.nvim',
      \   cmd = 'Octo',
      \   config = 'vim.cmd("packadd octo")',
      \ })
```

These `opt` packages I load in the `config` properties are just my way of loading plugin configuration. These packages just include all the setup calls, setting variables, highlights, signs, custom functions etc. pp.

The problem now is that both of these plugins get lazy loaded. But the `octo.nvim` plugin requires the `telescope.nvim` one. My approach to resolve this is with the `after` property.Telescope by itself just works fine. But Octo does not work until I load Telescope somehow else (i.e. run as `:Telescope...` command, cancel it, and then call `:Octo...`).
How can I resolve this issue to make it just work?

EDIT: Back then I tried to add `telescope.nvim` to the `require` list of Octo. But that gave me errors by Packer.

wbthomason:
Thanks for the report and example! I think this may be somewhere that `packer`'s sequenced loading needs to be improved...there's a way to hack this right now, but it's pretty gross. I think what we'd want is a `before` key or similar, so that you could trigger `telescope` to be loaded whenever `octo` is loaded (or the long-outstanding improvements to `requires` mentioned in #87).

The issue right now is that - while `octo` will be loaded whenever `telescope` is - there's nothing specifying that `telescope` should be loaded *before* `octo` is loaded. You could hack this by lazy-loading `telescope` on the `Octo` command (and keeping the `after = 'telescope.nvim' for `octo`), but that is inelegant and requires too much manual specification.

weilbith:
Sorry for the late response. :expressionless: 

So is there already an open issue for the `before` behavior? I thought to remember of an issue where this was part of the conversation. But by a quick search I can't find it again. If it exists, I assume we can close this issue here?

[...]

Issue #256, opened 2021-03-16
Options for multiple plugins


AckslD:
Would it be possible to add functionality to set options to multiple plugins? I.e. something as
```lua
    use {
        {'SidOfc/mkdx', config = 'require("plugin_settings/mkdx")'},
        -- table of contents
        {'mzlogin/vim-markdown-toc'},
        -- preview
        {'iamcco/markdown-preview.nvim', run = vim.fn['mkdp#util#install'] },
        ft = {'md'},
    }
```
where the `ft`-option  is set for all of them

muchosteechos:
This is also discussed here #139.

AckslD:
Thanks @muchosteechos! That seems slightly different though, but indeed related.

PR #264, opened 2021-03-22
An attempt to resolve lazy loading requirements of plugins

weilbith:
Attempts to close: #250
I have now run multiple times into this issue and it bothers me. While this topic seems to be controversial I decided to give it a try and share my first simple (and working) version here.
Next to the description in the issue, the commit message also includes a kinda generic description of the issue to solve. I settled on the new configuration option name `wants` as a reference to the naming of Systemd service units. So it is not out of nowhere (hopefully many users are already used to it/know its usage) and it also makes sense from the wording to me.
What do you think about the solution? It is pretty simple and flat. I saw some special stuff around (which I have missed since touching the code last time), but it did not look like I need that. Do I miss some important stuff. As I said, for me it is working. See the example `octo.nvim` + `telescope.nvim` and `vim-ultest` + `vim-test`.

---

While I made the attempt to also extend the documentation, I realized that is kinda hard to integrate in combination with the `after` option.  So I think it is required to rephrase this option and the whole **Sequencing** section too.
So I must admit that the `after` option is a little confusing for me. I think there is even an issue for that open? So for me this is actually rather a lazy load kind feature: Load this plugin when the other plugin got loaded. The description/example in the section actually even reads like this where the depending on plugin also is lazy loaded itself.
Could you help?

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -276,6 +276,11 @@ local function make_loaders(_, plugins)
         end
       end
 
+      if plugin.wants then
+        if type(plugin.wants) == 'string' then plugin.wants = {plugin.wants} end
+        loaders[name].wants = plugin.wants
+      end
+
       if plugin.fn then
         loaders[name].only_sequence = false
         loaders[name].only_setup = false
--- a/lua/packer/load.lua
+++ b/lua/packer/load.lua
@@ -12,6 +12,14 @@ packer_load = function(names, cause, plugins)
     if not plugin.loaded then
       some_unloaded = true
       needs_bufread = needs_bufread or plugin.needs_bufread
+
+      if plugin.wants then
+        for _, wanted_name in ipairs(plugin.wants) do
+          local wanted_plugin = plugins[wanted_name]
+          packer_load({wanted_name}, {}, plugins)
+        end
+      end
+
       if plugin.commands then
         for _, del_cmd in ipairs(plugin.commands) do cmd('silent! delcommand ' .. del_cmd) end
       end


wbthomason (approved):
Thanks! This looks pretty reasonable to me. There's some additional work that will need to be done - in particular, I don't think we want all of `after`, `requires`, and `wants` to exist together, so I'll need to merge/remove some of those.

However, I'm OK merging this as-is for the time being, to get the basic feature added. I can try to get to cleaning up the confusing overlapping options (and finally resolving #87) in the next week or two.

Thanks again!

wbthomason:
One note about this is that it will currently only work for dependencies of lazy-loaded plugins, i.e. this does not, in its current form, replace `after`. That will be resolved by me merging/removing this, `after`, and `requires`.

wbthomason:
Note to myself: we might eventually want to add a check for cycles to `wants` resolution.

weilbith:
Wow you were quite fast in just merging this now. ^^
We don't even have docs for this now. 😂 
Okay, I'm curious how this new interface will look like. Maybe my reference to Systemd is useful. I mean they do manage tons of dependencies and load order of different kind. And I personally think they do a great job and have an excellent implementation. So maybe checkout their [documentation](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#%5BUnit%5D%20Section%20Options) (especially `wants`, `requires`, `requisite`, `before`, `after`). And then there are the [reverse properties](https://www.freedesktop.org/software/systemd/man/systemd.unit.html#Mapping%20of%20unit%20properties%20to%20their%20inverses) like `wantedBy` for example. 🤷🏾

weilbith:
> Note to myself: we might eventually want to add a check for cycles to `wants` resolution.

Good point. That would actually have been one of the point I meant we need to check in detail. 🤣

wbthomason:
> Wow you were quite fast in just merging this now. ^^
We don't even have docs for this now. joy

Yep - this is actually intentional. Because I'm going to change the interface soon by fixing up the `after`, `requires`, etc. confusion, I don't want to document and advertise this interface which may also change. However, merging this won't break anything for anybody not using it and clears the path for you to use the feature without needing to use a PR branch, and I think the broader interface changes should be in a different PR.

> That would actually have been one of the point I meant we need to check in detail.

Yep, again, I figured merging would only affect people already using the feature, which is essentially just you, so it was low-risk.

> Maybe my reference to Systemd is useful. I mean they do manage tons of dependencies and load order of different kind. And I personally think they do a great job and have an excellent implementation.

It's worth consideration, though I'll note Systemd also has a reputation for unnecessary and confusing complexity, so I'm not entirely sure I want to mimic their interface too closely.

What I'm thinking (based in part on the discussion in #87) is:
- merge `wants` and `requires` under some name reflecting a predecessor dependency relationship (probably `requires`), i.e. for `use {A, requires = B}`, `B` must (1) be installed if `A` is installed and (2) always be loaded before `A`.
- Rename `after` to `load_after`, which indicates a successor dependency relationship, i.e. for `use {A, load_after = B}`, `A` must be loaded after `B` any time `B` is loaded.
- Resolve #256 to make it possible to apply options to groups of plugins, so that you can write `use {{A1, A2, A3, A4}, load_after = B}` to indicate that `A1` through `A4` should all be loaded after `B` is loaded.

Does this seem reasonable and less confusing?

weilbith:
For the top part: you have some good points! 👍🏾 

---

The `load_after` still suffers to be easy to grasp I think. For me it sounds more like an order only definition. But actually it is in first place a trigger/lazy load property. So maybe a `wanted_by` is more expressive? Or you differ strongly between order definitions and trigger. So the order definitions do not trigger the load of another plugin but just make sure none is loaded in a wrong order. But maybe you are right and I'm too much into Systemd which is not liked by everyone. 😆

akinsho:
Tbh I like the naming you've proposed, I can't imagine a clearer way to express

> A must be loaded after B any time B is loaded.

than using `load_after` . I think

> it sounds more like an order only definition. But actually it is in first place a trigger/lazy load property. So maybe a wanted_by or load_with/load_by

This sounds like it pre-supposes some knowledge of another system which I don't think packer users implicitly are thinking about, I'm not sure the idea of a load property and order property etc. is something people need to explicitly consider. I think it's much simpler for the keys to just express as succinctly as possible what will happen and some of the stuff around it therefore being lazy loaded etc be implementation details e.g.

If something is supposed to `load_after` something else I think it's clear that implicitly it will happen after something so if that thing is itself lazy loaded or not a user maybe doesn't have to think about that just know that whatever the case it will be after the other thing.

wbthomason:
I agree with @akinsho, but I will note that I like how `load_with` makes it clear that the plugins load at the same time as (technically immediately after) the target plugin.

nyngwang:
I think it would be good to have `wants` documented.

PR #267, opened 2021-03-25
Feature: add packer status

### LINKED ISSUE #12: Command to list installed plugins
It would be useful (sometimes) to have a function that lists the plugins managed by packer together with some rudimentary information -- mainly whether it's a `start` or an `opt` plugin, and maybe the source (commit hash, branch, local dir...). As a bonus, it could show the full config dictionary when selected (just like the `update` command shows the git log after updating).

I saw there's already a `packer/display` function, but that is not exposed?

akinsho:
image

<details>

packer_status_1

packer_status_2

#### Initial attempts
image

#### with string highlights
image
</details>

This PR will provide a new `PackerStatus` function which will open a window listing the currently installed packer plugins and their loaded status. I plan to start minimally by just adding the window with some basic info in this PR but hopefully I can iterate on this window so users can see the most recent commits as well as carry out actions like delete,revert etc. through the UI

This PR will fix #12

#### Follow up PR
- view latest update
- ability to revert
- ability to delete
- ability to load

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -100,6 +100,7 @@ packer.init = function(user_config)
     vim.cmd [[command! PackerSync     lua require('packer').sync()]]
     vim.cmd [[command! PackerClean    lua require('packer').clean()]]
     vim.cmd [[command! PackerCompile  lua require('packer').compile()]]
+    vim.cmd [[command! PackerStatus  lua require('packer').status()]]
   end
 end
 
@@ -396,6 +397,13 @@ packer.sync = function(...)
     local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
     display_win:final_results(results, delta)
     packer.on_complete()
+  end)()
+end
+
+packer.status = function ()
+  async(function ()
+    local display_win = display.open(config.display.open_fn or config.display.open_cmd)
+    display_win:status(_G.packer_plugins)
   end)()
 end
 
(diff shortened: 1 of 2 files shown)


weilbith:
@akinsho nice feature! What do you think about adding folds to each each plugin? So the initial view provides a better overview? Just a small idea. 🤷🏾

akinsho:
@weilbith that's a cool idea, I think it'll be good to add but might not do it in this PR. I want to keep the PRs for this functionality small and easy to review/ merge so I can iterate more quickly hopefully rather than one large PR that might get abandoned if I get busy 😅 .

The toggle logic looks like it'd need to be re-implemented or adapted for this particular use case since atm it's designed for the other windows usage.

akinsho:
@wbthomason this is now ready for review. I've kept it fairly light for now vs what I'd like to add in the long term, since that would be a much larger PR I think.

Things to note:
* I've created an allow list of keys so that as packer adds new keys they don't just appear in whatever format or potentially break due to some unexpected format of the key.
* I've added a few more syntax groups but they should be unique enough to only match things I've added or well booleans and strings but thought that might be a nice to have generally.

I've opted to leave stuff out (as I mentioned) so the details are pretty light at the moment but once I've added folding and can source git info then will be able to add more stuff in

akinsho:
Decided to add toggling since it was actually quite straightforward to use the existing logic
packer_status_folds

wbthomason:
This looks great to me! Thanks very much, @akinsho!

PR #271, opened 2021-03-31
Compile cmd as silent commands

eth3lbert:
This pr will compile cmd as silent commands which could bypass the `Invalid command name` error:
```
Error in packer_compiled: Vim(lua):E5108: Error executing lua [string ":lua"]:373: Vim(command):E182: Invalid command name
```

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -365,7 +365,7 @@ local function make_loaders(_, plugins)
   local command_defs = {}
   for command, names in pairs(commands) do
     local command_line = fmt(
-                           'vim.cmd [[command! -nargs=* -range -bang -complete=file %s lua require("packer.load")({%s}, { cmd = "%s", l1 = <line1>, l2 = <line2>, bang = <q-bang>, args = <q-args> }, _G.packer_plugins)]]',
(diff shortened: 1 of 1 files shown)


wbthomason:
Moving discussion here from https://github.com/wbthomason/packer.nvim/discussions/270

Sorry for the delay on this!

> Actually, it should be treat as different command, but I do agree with your suggestion and that's why I ask first

Yeah, so what I was saying is that `fzf.vim` actually doesn't do this: https://github.com/junegunn/fzf.vim/blob/master/plugin/fzf.vim#L50

It defines the one `GFiles` command and then handles the case of a trailing `?` separately. In my testing, if you just define `GFiles` as the lazy-load command, you get `GFiles?` too, just like in the eagerly-loaded plugin.


> Maybe we could also accept {"GFiles", silent=true} as cmd value

I'm OK with doing this if we ever have a need for adding `silent` commands. However, I don't think that this is the case - I think `silent` just hides the errors, and that - in your case - you get the `GFiles?` command not from your `silent` lazy-load command, but from the `GFiles` lazy-load command, which works without `silent`.

I'm open to changing my mind on this if there is a use case for `silent` after all, though!

eth3lbert:
Agree,  `silent` just hide the error, and potentially a harmful choice if we just silent command by default and without given any option to disable it at the same time. I have no other use case for `silent` command neither.

I'm open to just close this pr, or wait others to discuss.
Thanks 😄

wbthomason:
Given it's been a few days, I think I'll close this for now. Thank you again, though!

PR #275, opened 2021-04-05
Fix incorrect rtp directory concatenation

### LINKED ISSUE #274: rtp without initial seperator
If I want to specify a subdirectory as mentioned in the README, e.g. `vim` then I need to set this to `rtp = "/vim"`. Maybe the `/` should be added automatically?

EDIT: In fact there seems to be an issue since if I set `rtp = "vim"` then I get
```
    path = "/home/axel/.local/share/nvim/site/pack/packer/start/himalayavim"
```
but
```
vim.o.runtimepath = vim.o.runtimepath .. ",/home/axel/.local/share/nvim/site/pack/packer/start/himalaya/vim"
```
however if I set `rtp = "/vim" I get:
```
    path = "/home/axel/.local/share/nvim/site/pack/packer/start/himalaya/vim"
```
but
```
vim.o.runtimepath = vim.o.runtimepath .. ",/home/axel/.local/share/nvim/site/pack/packer/start/himalaya//vim"
```
i.e. in neither case I get what I want.

akinsho:
This PR fixes an issue where the plugin's install path was being directly concatenated with the `rtp` path rather than joining the two with a path separator as is being done for the actual `&rtp`.

Note for some reason the plugin that was mentioned in #274, `himalaya` still does not appear to load so it's not clear if this is an issue with that plugin or with packer.

Also note (unrelated fix): I took the liberty of fixing a global function I noticed with a very generic name that felt highly likely to eventually cause a clash with some user defined functionality. `restore_cursor`, I think going forward all global callbacks should probably be name spaced under an object like `__packer.callback` to reduce the likely hood of clashes

Fixes #274

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -163,12 +163,16 @@ local function make_loaders(_, plugins)
         end
       end
 
-      if plugin.rtp then table.insert(rtps, util.join_paths(plugin.install_path, plugin.rtp)) end
+      local path = plugin.install_path
+      if plugin.rtp then
+        path = util.join_paths(plugin.install_path, plugin.rtp)
+        table.insert(rtps, path)
+      end
 
       loaders[name] = {
         loaded = not plugin.opt,
         config = plugin.config,
-        path = plugin.install_path .. (plugin.rtp and plugin.rtp or ''),
+        path = path,
         only_sequence = plugin.manual_opt == nil,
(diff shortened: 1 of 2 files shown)


wbthomason (approved):
Thanks for tracking this down and fixing it! I agree re: namespacing any global functions we require.

PR #278, opened 2021-04-06
Fix #273: Escape any backslashes in event pattern strings

### LINKED ISSUE #273: Entries in `event` key aren't escaped in `packer_compiled.vim`
Example

A plugin is configured to load on `CmdlineEnter \?`, which in the plugin spec means that it has
```lua
use {
	...
	event = 'CmdlineEnter \\?',
	...
}
```

But in the compiled spec, we end up with a line containing the following:
```lua
vim.cmd [[au CmdlineEnter \? ++once lua require("packer.load")({...}, { event = "CmdlineEnter \?" }, _G.packer_plugins)]]
--																							  ^
--																							  this is the problem
```

As a result, the plugin will fail to load when it should, and throw an E5107:
```
E5500: autocmd has thrown an exception: Vim(lua):E5107: Error loading lua [string ":lua"]:1: invalid escape sequence near '"CmdlineEnter '
```

---

This is admittedly a bit of an edge case, and it can be mitigated by just not lazy loading the plugin, but I figured it'd be worth mentioning.

wbthomason:
Attempts to fix #273 by escaping backslashes in event patterns. I'm not 100% convinced that this simple fix won't break something with patterns, e.g. by escaping already-escaped backslashes, but I haven't put together a counterexample yet.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -320,7 +320,7 @@ local function make_loaders(_, plugins)
   for event, names in pairs(events) do
     table.insert(event_aucmds, fmt(
                    'vim.cmd [[au %s ++once lua require("packer.load")({%s}, { event = "%s" }, _G.packer_plugins)]]',
-                   event, table.concat(names, ', '), event))
+                   event, table.concat(names, ', '), event:gsub([[\]], [[\\]])))
   end
 
   local config_lines = {}


wbthomason:
I'm curious - has anyone on Windows run into issues with event patterns before this PR? Given the difference in file path separators, I want to make sure that this doesn't break anything on Windows.

YaBoiBurner:
I hadn't been paying attention to this PR much, but I'll try and test stuff out on Windows later today, or some time tomorrow.

If there's anywhere it'll break, it's windows.

That or it fixes a bug on Windows that nobody has run into yet.

YaBoiBurner:
Finally got around to testint it out.
This does fix a bug that appears to have gone unnoticed on Windows thus far.

I tested it with the following spec (no logic to it since i just wanna boot back into linux lol)

```lua
vim.cmd [[packadd packer.nvim]]
return require('packer').startup(function()
	use {'wbthomason/packer.nvim', opt = true}
	use {
		'9mm/vim-closer',
		event = [[BufRead C:/Users/Jaden\ Pleasants/Downloads/nvim-win64/Neovim/share/nvim/runtime/*]]
	}
end)
```

Actually, while trying to just get a working autocmd, I stubled upon this note around the help page for `autocmd-pattern`:
```vimhelp
Note that for all systems the '/' character is used for path separator (even
Windows). This was done because the backslash is difficult to use in a pattern
and to make the autocommands portable across different systems.

It is possible to use |pattern| items, but they may not work as expected,
because of the translation done for the above.
```
So if I'm not missing something here, this PR shouldn't actually have any Windows-specific effects, beyond maybe paths with whitespace characters being more prevalent than they are on Linux.

wbthomason:
Oh, nice catch on `autocmd-pattern`. Thanks for testing this - I think we're OK to merge for now. If any Windows users report breakage, we'll revisit this.

PR #279, opened 2021-04-06
Fix lazy loading plugins via require

akinsho:
This PR builds on #248 which at present doesn't actually work for a few reasons.

1. The name of the module passed in is not escaped and can contain special characters for lua patterns such as `-`, `.` etc. so the match actually fails in most cases.
2. With the names escaped packages are still not lazy loaded immediately because our loader is second in the list and seems the first loader failing will end up throwing an error.
3. With our loader as first in the list you get an infinite loop error because `require('lazy-load')` will go on cause a bunch of other requires so we need to cache what we have already checked.

@wbthomason this implementation is much closer to [`dein`'s](https://github.com/Shougo/dein.vim/commit/c6ddf04183606413870a0f8a86f61f8d58f85982) which I now believe was the way it is for the reasons above and so ours needs to match that.

Things I think will be of note:
1. Making the custom loader first in the order. I thought this might be a point of contention so I tested several times if it was possible to move it to 2 and have this work. Unfortunately it isn't. The first loader not finding our module seems to result in an error the first time. So with the loader as second the flow is `trigger lazy load via mapping -> errors first time -> trigger mapping -> success`. Whereas with the loader as first it works straight away.

2. Add the table of all previously checked modules. `dein` does a similar thing which I was wary of leaving out the first time around since I encountered some infinite loops initially. The reason we weren't seeing this was that the first loader handles keeping track of what was called. Since this loader is now first it needs to check so that subsequent calls to lazy load module don't try and check all modules infinitely.

I'm a little fuzzy on some aspects of how this works and fails primarily around the infinite looping and the exact behaviour of the first loader. Though I'm pretty sure that this now works.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -42,10 +42,6 @@ end
 
 local module_loader = [[
 
-local function escape_pattern(text)
-    return text:gsub("([^%w])", "%%%1")
-end
-
 local lazy_load_called = {}
 local to_load = {}
 
@@ -53,7 +49,7 @@ local function lazy_load_module(module_name)
   if module_name == 'packer.load' or lazy_load_called[module_name] then return nil end
   lazy_load_called[module_name] = true
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. escape_pattern(module_pat)) then
+    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. vim.pesc(module_pat)) then
       to_load[#to_load + 1] = plugin_name
     end
   end
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -41,11 +41,9 @@ end
 ]]
 
 local module_loader = [[
-
 local lazy_load_called = {}
-local to_load = {}
-
 local function lazy_load_module(module_name)
+  local to_load = {}
   if module_name == 'packer.load' or lazy_load_called[module_name] then return nil end
   lazy_load_called[module_name] = true
   for module_pat, plugin_name in pairs(module_lazy_loads) do
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -41,13 +41,20 @@ end
 ]]
 
 local module_loader = [[
+
+local function escape_pattern(text)
+    return text:gsub("([^%w])", "%%%1")
+end
+
+local lazy_load_called = {}
+local to_load = {}
+
 local function lazy_load_module(module_name)
-  if module_name == 'packer.load' then return nil end
-  local to_load = {}
-  local i = 1
+  if module_name == 'packer.load' or lazy_load_called[module_name] then return nil end
+  lazy_load_called[module_name] = true
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. module_pat) then
-      to_load[i] = plugin_name
+    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. escape_pattern(module_pat)) then
+      to_load[#to_load + 1] = plugin_name
     end
   end
 
@@ -55,7 +62,7 @@ local function lazy_load_module(module_name)
 end
 
 if not vim.g.packer_custom_loader_enabled then
-  table.insert(package.loaders, 2, lazy_load_module)
+  table.insert(package.loaders, 1, lazy_load_module)
   vim.g.packer_custom_loader_enabled = true
 end
 ]]


lua/packer/compile.lua
```diff
@@ -41,21 +41,28 @@ end
...
 local function lazy_load_module(module_name)
-  if module_name == 'packer.load' then return nil end
-  local to_load = {}
-  local i = 1
+  if module_name == 'packer.load' or lazy_load_called[module_name] then return nil end
+  lazy_load_called[module_name] = true
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. module_pat) then
-      to_load[i] = plugin_name
+    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. escape_pattern(module_pat)) then
```
wbthomason, line 56: Could `escape_pattern` be replaced with `vim.pesc`?

lua/packer/compile.lua
```diff
@@ -41,21 +41,28 @@ end
...
+
 local function lazy_load_module(module_name)
-  if module_name == 'packer.load' then return nil end
-  local to_load = {}
-  local i = 1
+  if module_name == 'packer.load' or lazy_load_called[module_name] then return nil end
+  lazy_load_called[module_name] = true
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. module_pat) then
-      to_load[i] = plugin_name
```
wbthomason, line 50: Whoops, nice catch on this.

lua/packer/compile.lua
```diff
@@ -41,21 +41,28 @@ end
...
+      to_load[#to_load + 1] = plugin_name
     end
   end
 
   require('packer.load')(to_load, {module = module_name}, _G.packer_plugins)
 end
 
 if not vim.g.packer_custom_loader_enabled then
-  table.insert(package.loaders, 2, lazy_load_module)
+  table.insert(package.loaders, 1, lazy_load_module)
```
wbthomason, line 65: I want to try one thing to potentially let us put this loader second; I'll experiment and push a commit if I get it working.

wbthomason (commented):
Thanks! I have admittedly not used this feature myself yet, so I'm glad you noticed it was broken. Everything mostly looks good - I have one question about using the built-in `vim.pesc` instead of `escape_pattern`, and I want to try something to fix the loader insertion order before I merge.

wbthomason:
Oh also @akinsho: do you have an example of a plugin spec that was causing the ordering issue to be apparent (i.e. showing that the first loader throws an error)?

lua/packer/compile.lua
akinsho: Yep just tested it and it works the same 👍🏿 hadn't used that one before 😄

akinsho:
The example which lead me to realise that this wasn't working is a little long
essentially I lazy load `nvim-treesitter-playground` since I use it super infrequently (not sure the merit of lazy loading but thats another issue) 😄 .

```lua
-- In plugins.lua
 use {
  "nvim-treesitter/playground",
   cmd = "TSPlaygroundToggle",
   module = "nvim-treesitter-playground",
}

--- This function is mapped e.g.
vim.cmd [[nnoremap ,E <cmd>lua token_inspect()<CR>]]
function __G.token_inspect()
  if not ts_playground_loaded then
    ts_playground_loaded, ts_hl_info = pcall(require, "nvim-treesitter-playground.hl-info")
  end
  if vim.tbl_contains(as_utils.ts.get_filetypes(), vim.bo.filetype) then
    ts_hl_info.show_hl_captures()
  end
end
```

I essentially ran this mapping to test. The ideal case is that on pressing `,E` it would just work. Which at present with the loader as first it does. With the loader as second it errors first and doesn't work till the second attempt

akinsho:
@wbthomason ☝🏿 this is a slightly tweaked version of what I've been doing/testing with locally. It's only slightly different i.e. I've removed some of my utility functions and left just the core logic.

wbthomason:
Ok! That took a bit of hacking, but I've gotten things working with the lazy-loader in the second slot of `package.loaders`.
To explain: (one of) the problems with the old version was that a searcher function (the function placed in `package.loaders` is supposed to return a function that returns the module, rather than returning the module directly (I do not know why this indirection is present, but it's how Lua does things...)

So, I modified the loader to return such a function, and allowed the main load function to forward this along through the module lazy-loader.

However, this doesn't quite work, because Lua has some logic to prevent circular requires that meant we could never load the actual module (https://github.com/LuaJIT/LuaJIT/blob/1e66d0f9e6698fdee672c40a9a5b4159c9254091/src/lib_package.c#L438)

So, I added a (potentially horrific) hack, using the `debug` library to modify the registry table to skip this circular require check in this case, since we know that we cannot cause an infinite loop.

I'm not sure if this is worth it - in 90% of cases, this is probably just a micro-optimization. It doesn't complicate the code much, but it does do some questionable things with `debug`. If we opt to go with the original approach (placing the loader in first position), we should keep the commit I added that makes `to_load` function-local (otherwise it will accumulate entries as more modules are lazy-loaded).

@akinsho, does this work for you? It seems to work for me on the examples you've given, though I still need to do more thorough testing.

akinsho:
@wbthomason I just tried it out and does work 👍🏿 . Ultimately I think I'll go with whatever you prefer, but IMO I'm a big fan *personally* of doing the simplest and often dumbest thing if it gives the almost as good a result as the clever solution.

In this case I think it will make the workings of what is already a somewhat obscure mechanism completely opaque to anyone looking to contribute. I also think perhaps more importantly that having to work around the way lua works by using the debug library to skip some checks seems like something to definitely avoid unless absolutely necessary.

I think if this proves costly for users or ends up needing to be optimised weeks, months or years down the line then it's worth revisiting this solution. I (perhaps incorrectly) think of this in terms of the actual overall cost of packer vs a) the value the functionality provides b) how much any performance cost here (which I think will already be quite minimal) will be dwarfed by a user adding a single expensive plugin to their config.

akinsho:
@wbthomason one optimisation I wonder if we could add if we go with the original implementation, is conditionally adding this loader to `package.loaders`. So checking the plugins a user has specified to see if there are any that use the `module` key. If a user doesn't use this key then we don't really need to ever add our own loader

wbthomason:
Yeah, I buy that argument/like the proposed optimization. I'll pick out the bad commits and add that in now.

wbthomason:
We actually already have that optimization: https://github.com/wbthomason/packer.nvim/blob/b495895dffca9aabfead066a860b87ab3a67cf7e/lua/packer/compile.lua#L487

wbthomason:
Ok! Rebased the hack away (keeping the fix with `to_load`), and it works on the example you sent above. I think we're good to merge. Thanks again!

Issue #287, opened 2021-04-12
Crashes with path with spaces in its installation directory


sam20908:
When Packer tries to install itself into a path with spaces, it fails because It doesn't handle it.

![image](https://user-images.githubusercontent.com/18388795/114470726-38834180-9ba4-11eb-9f13-6838468882b6.png)

This also applies to installing **anything** too, even if I manually cloned packer to the right directory.

![image](https://user-images.githubusercontent.com/18388795/114470861-708a8480-9ba4-11eb-855c-3331093253e3.png)

wbthomason:
Thanks for the report!

Your first problem (with cloning `packer`) is just how spaces in paths are handled by `git` and most CLI tools - you need to manually escape them. `packer` does not, in general, clone itself (I assume this is from a bootstrap function?)

The second problem is related to the stalled PR #166. Could you please give that a try (I just rebased it onto current `master`) and confirm that it does not solve the problem for you? Assuming it doesn't, could you please post the contents of `stdpath('cache')\packer.nvim.log`?

sam20908:
Confirmed that #291 does fix the first issue. I'll test the second issue tomorrow.

sam20908:
Confirmed that #166 does work as well.

sam20908:
> Confirmed that #166 does work as well.

It does *partially* work, but I get this:
![image](https://user-images.githubusercontent.com/18388795/114958215-79cf5780-9e17-11eb-85f3-ac9d08f5cd7c.png)

sam20908:
I can confirm that #166 doesn't work on Windows with space in path.

sam20908:
I think we need to use `system` to invoke cloning, similar to #291.

nanotee:
That's odd, the error code is `E739`, which according to the Vim documentation is caused by `vim.fn.mkdir()`, but on my Linux system the function handles spaces just fine. I'll have to test this on Windows

wbthomason:
> I think we need to use `system` to invoke cloning, similar to #291.

@sam20908 `system` is blocking, so it's not an option. Using the `luv` spawn API, as we do, should be fine with proper string escaping.

Given that `fnameescape` is a built-in function, could you test if it works as expected for escaping when used manually on Windows?

sam20908:
> That's odd, the error code is `E739`, which according to the Vim documentation is caused by `vim.fn.mkdir()`, but on my Linux system the function handles spaces just fine. I'll have to test this on Windows

[...]

PR #291, opened 2021-04-13
docs: Use `system()` in the bootstrapping snippet

nanotee:
Passing a List to `system()` saves users the trouble of escaping spaces and other problematic characters

(You can test this with `:call system(['git', 'clone', 'https://github.com/wbthomason/packer.nvim', 'my folder with spaces'])`)

Partially addresses #287


wbthomason (approved):
Thanks, this is a good suggestion!

PR #292, opened 2021-04-13
Fix possible lazy loading recursion when running plugin config

bphilly96:
I ran into this problem when lazy loading on require. If the config for a plugin required a module which was set to lazy load that same plugin, recursion would occur because the plugin has not been set to loaded before running the config.

I originally changed `lazy_load_module()` in compile.lua which fixed it specifically for lazy loading on require however I then thought it might be simpler to just set the plugin as loaded before config is run inside the main load function. This way it will also fix this in case there is any other way to load a plugin inside it's own config.

### DIFF
--- a/lua/packer/load.lua
+++ b/lua/packer/load.lua
@@ -35,10 +35,14 @@ packer_load = function(names, cause, plugins)
         end
       end
 
+      -- set plugin as loaded before config is run in case something in the config
+      -- tries to load this same plugin again
+      plugin.loaded = true
       if plugin.config then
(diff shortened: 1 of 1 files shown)


wbthomason:
Sorry, I hadn't gotten around to merging this, but ran into a related issue in #300. I think I'll try to add you as a co-author on that PR and close this one, since I added a bit of other useful context and refactoring in #300.

bphilly96:
Ok no worries 👍

wbthomason:
Thanks again! I added you as a co-author on #300.

PR #295, opened 2021-04-15
docs: Don't mention patterns in module key (#293)

### LINKED ISSUE #293: Confusion over pattern in module key of use
When lazy loading a plugin using the module key in use, the docs state the value of module should be a pattern. However in the source code at https://github.com/wbthomason/packer.nvim/blob/fdf1851c6121dee98294791c72aebff92b99b733/lua/packer/compile.lua#L50 the pattern is escaped meaning the pattern will be treated literally. Is this behaviour intended? Shouldn't the user escape the pattern manually so that they can choose which characters to treat literally and which to treat as special characters?

For example is the user wanted to lazy load on any module name consisting of upper-case letters or punctuation they would set `module = '[%u%p]+$'`. The current implementation would escape all of these special characters and cause string.match to treat them literally.

Furthermore, should the leading `"^"` that's in the source code be documented as well as documenting that there is no trailing `"$"` inserted for you? I'm happy to add this myself, just want to make sure I'm not misunderstanding something.

bphilly96:
Closes #293

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,8 @@ use {
   event = string or list,      -- Specifies autocommand events which load this plugin.
   fn = string or list          -- Specifies functions which load this plugin.
   cond = string, function, or list of strings/functions,   -- Specifies a conditional test to load this plugin
-  module = string or list      -- Specifies patterns (e.g. for string.match) of Lua module names which, when required, load this plugin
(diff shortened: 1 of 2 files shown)


akinsho (approved):
LGTM

wbthomason (approved):
Yup, thanks @bphilly96!

Issue #299, opened 2021-04-17
BUG: module config causes error while loading plugin


shadmansaleh:
I was trying to load telescope with module config. But I got the following error.

<details>
<summary>Error message</summary>

```
Error running config for telescope.nvim
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:43: [string "..."]:0: loop or previous error loading mod
ule 'telescope'
Please check your config for correctness
Error running config for telescope.nvim
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:43: [string "..."]:0: loop or previous error loading mod
ule 'telescope'
Please check your config for correctness
Error running config for telescope.nvim
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:43: ...k/experimental/opt/telescope.nvim/lua/telescope/i
nit.lua:39: loop or previous error loading module 'telescope.config'
Please check your config for correctness
Error running config for telescope.nvim
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:43: ...k/experimental/opt/telescope.nvim/lua/telescope/i
nit.lua:39: loop or previous error loading module 'telescope.config'
Please check your config for correctness
Error running config for telescope.nvim
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:43: ...k/experimental/opt/telescope.nvim/lua/telescope/i
nit.lua:39: loop or previous error loading module 'telescope.config'
Please check your config for correctness
Error running config for telescope.nvim
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:43: ...k/experimental/opt/telescope.nvim/lua/telescope/i
nit.lua:39: loop or previous error loading module 'telescope.config'
Please check your config for correctness
Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:66: Vim:E492: Not an editor command: Telescope
Please check your config for correctness
Press ENTER or type command to continue
```

</details>

### Minimal config
```lua
require('packer').startup(function()
  use { 'nvim-telescope/telescope.nvim',
    opt = true,
    module = 'telescope',
    cmd = 'Telescope',
    config = function()
      require('telescope').setup()
    end,
    requires = {
      'nvim-lua/popup.nvim',
[...]

PR #300, opened 2021-04-18
Fix first half of #299: Correct module load ordering

wbthomason:
Previously, we set a plugin as `loaded = true` at the *end* of its part of the load function. However, with the addition of Lua module lazy-loading, the load function is re-entrant. This means that we try to load a plugin in the middle of loading it, which breaks all sorts of things.

These changes move the `loaded = true` mark earlier to prevent this, and include miscellaneous fixes and refactorings for the module lazy-loader.

This does **not** fix the part of #299 that arises from transitive optness with `require`'d plugins; that's a duplicate of #87 and will be fixed in another PR (hopefully in the next few days, if I can get to it).

@shadmansaleh, can you please confirm that this solves #299 for you if you ensure that `plenary` and `popup` are not `opt` first (or wait until the second PR to fix that part too)?

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -47,7 +47,7 @@ local function lazy_load_module(module_name)
   if lazy_load_called[module_name] then return nil end
   lazy_load_called[module_name] = true
   for module_pat, plugin_name in pairs(module_lazy_loads) do
-    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, "^" .. vim.pesc(module_pat)) then
+    if not _G.packer_plugins[plugin_name].loaded and string.match(module_name, module_pat)then
       to_load[#to_load + 1] = plugin_name
     end
   end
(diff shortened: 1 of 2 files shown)


shadmansaleh:
It definatly works for me :)

bew:
stylistic suggestion: instead of setting `loaded` at the beginning (which does not make much sense since the plugin is not really loaded yet), what about adding a `loading` var, that you remove before setting `loaded`

wbthomason:
@bew: Thanks for the suggestion. I agree that the `loaded` variable should probably be renamed, but I'm hesitant to add another temporary state variable. My thinking is that we don't currently have any mechanism for re-trying a failed load, so the only state we actually want to represent is whether we've _tried_ to load a plugin - which is what (I think) this PR makes `loaded` represent.

The name is a bit unclear, though, so I'm open to changes there. For now, I'll go ahead and merge this PR to get the fix out sooner.

PR #310, opened 2021-04-21
Add border to remove prompt

shadmansaleh:
I honestly sometimes have hard time finding this float :P Colorscheme might have a hand in that . But borders are cool and with bfredl making them so easy to add I see no reason not to have them :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ default configuration values (and structure of the configuration table) are:
     moved_sym = '→', -- The symbol for a plugin which was moved (e.g. from opt to start)
     header_sym = '━', -- The symbol for the header line in packer's display
     show_all_info = true, -- Should packer show all update details automatically?
+    prompt_border = 'double', -- Border style of prompt popups.
     keybindings = { -- Keybindings for the display window
       quit = 'q',
       toggle_info = '<CR>',
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -220,6 +220,7 @@ default values: >
       moved_sym = '→', -- The symbol for a plugin which was moved (e.g. from opt to start)
       header_sym = '━', -- The symbol for the header line in packer's display
       show_all_info = true, -- Should packer show all update details automatically?
+      prompt_border = 'double', -- Border style of prompt popups.
       keybindings = { -- Keybindings for the display window
         quit = 'q',
         toggle_info = '<CR>',
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -69,6 +69,7 @@ local config_defaults = {
     header_lines = 2,
     title = 'packer.nvim',
     show_all_info = true,
+    prompt_border = 'double',
     keybindings = {quit = 'q', toggle_info = '<CR>', diff = 'd', prompt_revert = 'r'}
   },
   luarocks = {python_cmd = 'python'},
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -144,7 +144,8 @@ local function prompt_user(headline, body, callback)
     col = x,
     row = y,
     focusable = false,
-    style = 'minimal'
+    style = 'minimal',
+    border = config.prompt_border,
   }
 
   local win = api.nvim_open_win(buf, false, opts)


akinsho:
@shadmansaleh maybe this should be configurable so users can choose how the border shows. There is a `display` subtable in the config where you can add the key and a user can just specify whatever key is valid for `nvim_open_win`'s border key. Each module should already have access to the config at the top of the file

shadmansaleh:
> @shadmansaleh maybe this should be configurable so users can choose how the border shows.

Added .

I feel like [this](https://github.com/wbthomason/packer.nvim/blob/f9dc29914f34cb2371960236d514191b9feba8b5/lua/packer/display.lua#L93) line should be at the top of the file so it's mpre noticable . Currently it seems to be in middle of nowhere :D

akinsho:
@shadmansaleh I agree re. the config line's location, feel free to move it somewhere more visible

akinsho (approved):
This LGTM, it's easily configurable but is `double` a good default. A single line border seems like a more un-opinionated default than the double one but really not a big deal either way

wbthomason (approved):
LGTM, thanks! I hadn't heard about the border change; that's exciting for general UI things.

yamatsum:
I updated neovim and packer to the latest version, but the border does not appear.

Screen Shot 2021-04-26 at 13 04 13

shadmansaleh:
> Do I need any other settings? Am I doing something wrong?

Humm I didn't add border to `PackerSync` float . I thought it was a split for all this time as it opened in one side of screen :P .  You can try disabling a plugin and the `PackerSync` than should open a prompt with border . It should be really easy to add border to `PackerSyncs` float too if it's a wanted feature :)

yamatsum:
To avoid confusion, I think it's better to border all other floats, not just PackerSync.

akinsho:
@shadmansaleh the deletion float is actually a bit of an outlier since normally packer allows a user to specify how it should open and provides a util to allow for floats which is what @yamatsum is referring to. By default packer uses splits but a user can opt for packers float implementation which is where that change could go. Tbh I'm not sure why the removal float doesn't use the existing float implementation so all packer floats are created the same way.

wbthomason:
The deletion float existed before `packer` included a default floating window function for displaying progress, status, etc.

@akinsho are you suggesting that the removal prompt should use the `win_fn`/`win_cmd` that the rest of the displays use? The original reason for *not* doing this was that, during `sync`, we already have a split open (by default), so I wanted to show a float rather than using another split.

akinsho:
@wbthomason nope wasn't thing of using another split since that would definitely be weird just more whatever specific function is used to create a float so that if that function is updated the removal float would also be updated. Not sure what the function is called I don't use it but I presume there is one somewhere.

wbthomason:
The removal float is the only one that `packer` creates by default; we could make that use https://github.com/wbthomason/packer.nvim/blob/master/lua/packer/util.lua#L103.

yamatsum:
Is it possible to add a border like telescope.nvim?

akinsho:
@yamatsum telescope's border still relies as far as I know on creating a fake border window so they can add in titles and such which isn't supported natively. Tbh I *personally* think for packer's use case that we should just keep things simple and use the builtin border and what that supports (i.e. not titles in the border line). 

It's quite alot of extra logic for a bit of eye candy, I think if someone wanted to implement it that way though 🤷🏾 but I think the whole fake border window thing is not worth it IMO given the native support, especially since it will mean that in order to support user customisation they'd have to also factor in customising the border with user specified characters.

yamatsum:
> Tbh I personally think for packer's use case that we should just keep things simple and use the builtin border and what that supports (i.e. not titles in the border line).

I think so.
I didn't want the title in the window, I wanted the four corners to be rounded.

akinsho:
@yamatsum that should be easy enough to either configure in this plugin or as a user. With the native borders you can do
```lua
vim.api.nvim_win_open(buf, arg, {
      border = {"╭", "─", "╮", "│", "╯", "─", "╰", "│"}
}
```

yamatsum:
I was able to set it! Is it possible to allow the border to change by default?
Screen Shot 2021-04-29 at 1 29 00

akinsho:
@yamatsum I believe someone has already submitted a PR to allow passing your own args to packers float function #325

PR #313, opened 2021-04-22
Fix #312: Correctly generate profiling code for conditionals

### LINKED ISSUE #312: optional plugins loads even if cond is false
For a while now I have used the cond-option for a plugin, to temporarily disable plugins with `function() return true end` and `function() return false end` to enable and disable plugins. Today I found all my plugins loaded even if I had it turned off with the function returning false. Example:

This worked previously, but does not work now:

```Lua
use {'seblj/nvim-tabline',
    cond = function() return false end
}
```

wbthomason:
Fixes #312 by special-casing the profiling code generation for conditionals. This is necessary because - unlike for other kinds of timed chunk - we cannot easily put `time` calls around the chunk in the conditional syntax.

@seblj, @yutkat, @yujinyuz: can you confirm that this fixes #312 for you without causing new issues? It works for me on the minimal reproducer. Thanks!

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -31,7 +31,7 @@ endtry
 
 ---@param should_profile boolean
 ---@return string
-local profile_time = function (should_profile)
+local profile_time = function(should_profile)
   return fmt([[
   local time
   local profile_info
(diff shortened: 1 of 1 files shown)


seblj:
It works for me!

yujinyuz:
works for me too! had to delete the previous `packer_compiled.vim` before it worked

yutkat:
It works! Thank you for your quick fix.

wbthomason:
Sorry for the quick fix but slow merge! :sweat_smile: I got caught up with other work and forgot to merge this. Doing so now - thanks for the help testing the patch!

PR #314, opened 2021-04-22
Feature/add testing infrastructure

akinsho:
@wbthomason should probably have run this past you in an issue but it's pretty simple to setup so I just did it.

This PR adds the infrastructure required to run tests for packer using `plenary`'s busted test lib. It doesn't require users to have plenary just CI and anyone developing locally. If this seems alright to you I'll add a contributing section that describes how to run tests locally.

The tests currently in place are more examples/placeholders and something to run for now

### DIFF
--- a/lua/packer/.github/workflows/test.yaml
+++ b/lua/packer/.github/workflows/test.yaml
@@ -1,0 +1,31 @@
+---
+name: Run tests
+on:
+  pull_request: ~
+  push:
+    branches:
+      - master
+
+    jobs:
+      build:
+        name: Run tests
+        runs-on: ubuntu-latest
+
+        steps:
+          - uses: actions/checkout@v2
+          - run: date +%F > todays-date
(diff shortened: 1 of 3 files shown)


wbthomason (approved):
Thanks, this is well overdue. My ad hoc testing may have been sufficient initially, but it's not cutting it anymore.

I'll make a new issue for adding some real tests.

akinsho:
@wbthomason I just realised 🤦🏾 🤦🏾 the github workflow is in the wrong directory. It should be in the root 🤕 I'll push a fix now 😆

wbthomason:
Oh, whoops. Sorry, I should've caught that, but I know very little about Github CI. I should read up on how that works.

akinsho:
I've only recently started adding this workflow to some of my plugins and it generally works quite well but there are very minor gotchas like making sure things are in the correct location (but that's a fairly standard requirement 😅 )

PR #315, opened 2021-04-22
chore: add issue templates

nanotee:
This PR adds issue templates (largely inspired by the ones in the Neovim repository).

Hopefully encouraging users to answer these questions upfront will reduce the amount of back and forth needed to diagnose issues since a lot of them tend to be related to software/os versions


wbthomason:
Thanks, this is overdue (I've been lazy in creating these)!

I'm going to push one small modification to the template (I think it makes sense to request the log file and compiled file by default for bugs), then merge.

wbthomason (approved):
Thanks again!

PR #317, opened 2021-04-24
docs: Add diff keybinding

pockata:
I noticed that the `diff` keybinding isn't listed in the README and vimdoc so I added it


akinsho (approved):
LGTM @pockata thanks

wbthomason:
Thanks @pockata and @akinsho!

PR #320, opened 2021-04-25
fix: move github test workflow to root directory

akinsho:
This PR fixes the location of the test github workflow file to the correct directory so tests actually run


wbthomason:
Thanks!

Issue #319, opened 2021-04-25
Chore: add tests


wbthomason:
@akinsho put basic `busted` testing infrastructure in place with #314. Now, we need to add real tests. These should cover:
- [ ] Different package_roots
- [ ] All flavors of specification (e.g. `startup` and variants, bare functions, etc.)
- [ ] All kinds of lazy-loaders
- [ ] Ideally most combinations of lazy-loaders and other features
- [ ] Luarocks installation (both hererocks and actual rocks, test that we can require as expected)
- [ ] Different config values like max_jobs, git commands
- [ ] Probably more things?

We also might want to add Windows and macOS environments, if possible.

PR #321, opened 2021-04-25
bugfix: indentation of workflow yaml

akinsho:
Sorry @wbthomason should have marked the initial PR as draft since I saw it fail once I pushed it. Seems when I copied across the workflow the yaml's indentation broke so it wasn't running correctly. I'll leave this as draft for a few mins just till I see if it passes then change it if it builds successfully


wbthomason:
Oh, ok - my bad for jumping the gun. I'll stay hands off from this one, sorry.

akinsho:
@wbthomason Ok this now passes and correctly runs the workflow. I had `tests` dir at the wrong level as well 🤕 , so fixed that as well as the test cases. One gotta is that running plenary tests locally unless you use the directory function you can't specify an `init.vim` which can mean that your local config interfered. So I needed to update the test cases once I was running a proper minimal init.vim

wbthomason:
Looks good, thanks. I made #319 to track implementing actual functional tests using this infrastructure; thanks again for getting it set up!

PR #325, opened 2021-04-27
Allow setting extra opts for floating window

gbrlsnchs:


lua/packer/util.lua
```diff
@@ -110,14 +110,14 @@ util.float = function()
   local left = math.ceil((columns - width) * 0.5)
   local top = math.ceil((lines - height) * 0.5 - 1)
 
-  local opts = {
+  opts = vim.tbl_deep_extend('keep', {
```
akinsho, line 113: @gbrlsnchs I think you might want to use `force` here since presumably this should prefer the user's options over packer's?

akinsho: > "keep": use value from the leftmost map
from the help docs, this means the left side map takes priority, so you won't actually be able to override anything explicitly set in packer

gbrlsnchs: Hi @akinsho, thanks for the review. My initial idea was to not be able to override what Packer explicitly sets so one wouldn't be able to mess with the size of the window. Do you think it would be better if a user could do such?

akinsho: I guess it means that a user can completely mess with the window but ultimately it's their prerogative 🤷🏾‍♂️ I think if we give them the ability to change this we should just give them full control since it only affects their system and who knows what someone might prefer.

They might want to put the window in the bottom right of the editor and completely mess with the height. More importantly I think it will be confusing to people that they can pass in some settings and those work but others just don't.

gbrlsnchs: I think that makes total sense. I'll rebase with the proper parameter. Thanks for the insight.

gbrlsnchs:
I have just rebased the commit in order to allow _all_ settings passed to `nvim_open_win` to be overridden.

akinsho (commented):
LGTM 👍🏾 I guess it's worth documenting this somewhere so people know they can pass these options in. Maybe in the help docs or the README wherever there might already be a reference to the `float` function

wbthomason (approved):
Thanks! I was waiting to modify this PR to allow overriding all the options, but it looks like you did that already. I agree with @akinsho re: docs; otherwise, I think this is ready to merge.

yamatsum:
I'm waiting for this PR merge

akinsho:
@yamatsum this is waiting on a `docs` update then will be good to merge 🤷🏾‍♂️

gbrlsnchs:
I'll work on docs right now, hold on...

gbrlsnchs:
Done @akinsho @wbthomason, added docs.

akinsho (approved):
Awesome thanks for adding in the docs 💯 🥇 (didn't realise there were zero docs on floats)  this looks good to merge now, will give it an hour or so for @wbthomason to chime in, if not will hit the button 👍🏾

gbrlsnchs:
All right @akinsho, thank you too!

clason:
I think that broke something: on sync
```
[packer.nvim] [ERROR 19:26:14] async.lua:20: Error in coroutine: vim/shared.lua:219: after the second argument
: expected table, got string
[packer.nvim] stack traceback:
[packer.nvim] ^Ivim/shared.lua:219: in function 'tbl_deep_extend'
[packer.nvim] ^I...m/site/pack/packer/start/packer.nvim/lua/packer/util.lua:113: in function 'opener'
[packer.nvim] ^I...ite/pack/packer/start/packer.nvim/lua/packer/display.lua:674: in function 'open'
[packer.nvim] ^I...site/pack/packer/start/packer.nvim/lua/packer/update.lua:105: in function 'update'
[packer.nvim] ^I...e/nvim/site/pack/packer/start/packer.nvim/lua/packer.lua:386: in function <...e/nvim/site/p
ack/packer/start/packer.nvim/lua/packer.lua:369>
```

I use the following config:
```lua
local config = {
  compile_path = vim.fn.stdpath('data')..'/site/pack/loader/start/packer.nvim/plugin/packer_compiled.vim',
  display = {
    open_fn = require'packer.util'.float,
  }
}

return require'packer'.startup {plugins, config = config}
```

akinsho:
Ahh the `open_fn` is being called with a string argument 😓 `[packer]`. 🤔 The `float` function doesn't use it though. In the interim @clason can you change the signature to ignore the argument e.g.
```lua
open_fn = function(name)
  require'packer.util'.float()
end
```

clason:
```
Error executing vim.schedule lua callback: .../site/pack/packer/start/packer.nvim/lua/packer/async.lua:20: attempt to concatenate local 'val' (a nil value)
```
Sorry :/

akinsho:
Can you try the PR instead of the workaround, that seems to resolve the issue for me

PR #329, opened 2021-04-29
compile.lua: better error messages when nil

ray-x:
It is an issue for me from time to time when I update vimrc and failed to compile. Sometimes the error message is obvious and sometimes not.
This PR is to address the case when dependency(after) was not setup up correctly and compile.lua failed of nil dereference.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -480,6 +480,9 @@ local function make_loaders(_, plugins, should_profile)
 
   local sequence_loads = {}
   for pre, posts in pairs(after) do
+    if plugins[pre] == nil then
+     error(string.format("Compile.lua dependency not loaded %s %s", pre, vim.inspect(posts))) 
+    end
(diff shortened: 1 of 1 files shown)


wbthomason:
Thanks for the PR! Can you give an example of when this error happens?

ray-x:
yes. The setup I got trouble:
```lua
use {'hrsh7th/nvim-compe', event = 'InsertEnter', config = function()...end }
use {'hrsh7th/vim-vsnip'', event = 'InsertCharPre',  setup = function()...end, after = "hrsh7th/nvim-compe"}
```

wbthomason:
Sorry for taking so long to get back to this! This looks good - I've tweaked the error message somewhat, but if you're still OK with the changed message I'm good to merge this.

ray-x:
Thanks. The new message looks good to me.

wbthomason:
Great, thanks again!

PR #335, opened 2021-05-03
feat: add ability to allow unrelated git histories

### LINKED ISSUE #327: Automatically handled failed updates due to rebases
### Describe the feature
Occasionally when updating a plugin it fails because the author has rebased their branch, so you see output that looks like
```
 ✗ Failed to update ray-x/lsp_signature.nvim
  Errors:
    fatal: Not possible to fast-forward, aborting.
    e4852e0
```
Ideally plugin makers wouldn't do that but really it's quite a helpful thing to be able to do in git and inevitable that people will. This is even more common if using a branch of a plugin that hasn't been merged.

I've tried setting `--rebase=true` in the update command which works but this removes the ability to inspect individual commits from other updates. I briefly looked into adding a new key binding which would allow a user to rebase manually by clicking over a plugin, but am wondering if this can't shouldn't just be automatic. The user in all likelihood just wants things to update regardless of the git workflow of the author so this seems like a good place to automate things instead.

I think a potential solution would be to inspect the error returned and if it matches a rebase error then just perform a rebase. This could potentially be configurable if for some reason someone didn't want this. It would also only apply in the cases where there has been a rebase and an update fail due to that.

Thoughts @wbthomason? also if you don't hate the idea where might I look, I've had a read through `git.lua` so assume it would be at some point in the `updater` function just not sure exactly where/if you have any ideas re. how

I guess alternatively there might be some way to update the `updater` so it also handles needing to rebase in the first place. I don't know what logic `vim-plug` uses but I don't think I've ever had to worry specifically about if there's been a rebase or not whilst using it.

akinsho:
This will hopefully fix #327 by adding back in the ability for git to all merging branches with unrelated histories. This is happening as plugin makers are rebasing feature branches as well as occasionally the main branch of their plugins which essentially rewrites their projects history in a way a normal git merge post git 2.9 cannot automatically handle.

### TODO
- [ ] Test this out on a rebased branch/project

As a side note this seems reasonable to set as the default since it won't affect most normal cases only ones where the histories have diverged. Also git 2.9 was released in 2016 so presumable people have updated their git versions in the last *5 years*


wbthomason:
I'd also like us to test this on repositories which have diverged due to local commits/changes. We may or may not care if these changes are overwritten by using this flag with `git`, but we should know what the effect is and document it either way.

akinsho:
@wbthomason testing out the local changes issue and this flag actually won't stop that from failing because the `git pull --ff-only` command is what is failing i.e. if you have local changes it's impossible to fast forward the branch. So I guess the behaviour here then is that that will still fail. Tbh that particular case I'd say is different from what the issue was specifically trying to fix since tbh if a user has changed things locally it's not really clear what should actually happen.

akinsho:
Also just tested this with the main use case i.e. a rebase on a branch i.e. main or other branch and it doesn't actually work. Primarily because the `--ff-only` key still causes the command to fail despite the `--allow-unrelated-histories` flag in the `update` command. Removing `--ff-only` leads to a can of worms without `rebase=true` though since this results in merge conflicts etc.

I think this might need the initial solution we discussed of automatically retrying not sure yet

wbthomason (approved):
These changes look good to me, thanks. I'm not a `git` maestro, though, so one or both of us should test them thoroughly.

akinsho:
@wbthomason nope actually in this state it doesn't actually work at all. Locally I've been working on something like the original discussion about checking the error output and automatically retrying which kind of works. I just have questions about how to structure the retry using packer's async implementation

wbthomason:
Oh, sorry - I badly misread that comment. Happy to answer questions about the structuring.

lua/packer/plugin_types/git.lua
```diff
@@ -264,12 +264,22 @@ git.setup = function(plugin)
...
-          return {
-            msg = fmt('Error pulling updates for %s: %s', plugin_name,
-                      table.concat(update_info.output, '\n')),
-            data = err
-          }
+          -- TODO should retrying be inside of the map error function
+          local git_err = update_info.err and update_info.err[1] or nil
+          if git_err and git_err:match(vim.pesc("Not possible to fast-forward")) then
+            local update_with_rebase = update_cmd:gsub('rebase=false', 'rebase=true')
+            r = await(jobs.run, update_with_rebase,
```
akinsho, line 271: @wbthomason so the advice I needed was how to use the current job/async module to handle hitting an error and wanting to start off another job. Can I just do this inside of the `map_err` as I've done. This sort of works, as in the logic is triggered (though incomplete) but looks kind of wrong (for no concrete reasons, just a feeling).

Or is there some other mechanism for this or should there be?

wbthomason: The `Result` type semantics are uglier than I wanted them to be in general; I've thought about redesigning that mechanism but haven't had any great ideas for it yet.

The better way to do what you want here is with `or_else`: this takes and runs a result-returning function, so you could pass it this function doing the retry and return the result of your `await():and_then` sequence.

For what it's worth, the `Result` "type" is meant to provide similar semantics to e.g. https://doc.rust-lang.org/std/result/ or Haskell's `Either`, if you're familiar with those. Unfortunately, Lua's syntax (e.g. repeating `function()` everywhere) make it rather ugly.

akinsho: Ah @wbthomason that makes sense I wasn't aware of the `or_else` method. Re. `Result` and `Eithers` and other monad/monad like patterns I'm generally a big fan of. Will give that a go 👍🏾

akinsho:
@wbthomason could use a hand here, I've switched to the `or_else` helper but the subsequent command I run inside the `or_else` seems to just hang indefinitely. I wrapped the function I passed to the or else in an async func which is probably pointless since the parent function is already async but not sure why it won't proceed tbh 😓

wbthomason:
@akinsho Taking a look now!

wbthomason:
@akinsho How have you been testing this? Making a local commit?

akinsho:
I have a `test_branch` on one of my plugins that I'm just rebasing and then trying to update.

```lua
    use {
      "akinsho/dependency-assist.nvim",
      branch = "test_branch",
      config = function()
        return require("dependency_assist").setup()
      end
    }
```

This won't work for you since you won't have a copy of the branch before it was rebased but you could do a similar thing with any plugin repo you have control over.

It's kind of an annoying one to test but I guess you could also just rebase the local master of a plugin you have installed

wbthomason:
Thanks! I am still looking at this, but having trouble myself. I'm a bit worried this is a bad interaction between the `result` type and async logic, i.e. that the `await` within the `or_else` function isn't getting added to the async work loop correctly.

akinsho:
@wbthomason any further ideas on the behaviour of this one? Can dig in myself at some point, otherwise just not sure if you managed to find anything?

wbthomason:
@akinsho I didn't write an update because I didn't have a fix yet, but it does seem that the coroutine used by async doesn't resume correctly for the nested await here (the job run in the `or_else`. I'm not sure why this happens --- one ugly hack workaround is to use `if r.err ~= nil then` {the current body of the `or_else` function} `end` instead of `or_else`.

PR #340, opened 2021-05-04
feat: add ability to detect when a different repo with a similar name is added

### LINKED ISSUE #231: Packer doesn't detect when switching to a different plugin with the same repo name
Whilst trialing PR fixes for various plugins, I've noticed that moving from
```lua
-- from
use {"person-1/cool-plugin"}
-- to
use {"person-2/cool-plugin"}
```

Doesn't trigger a packer clean and then subsequent installation of the fork of the package, I think this is because packer primarily distinguishes based on plugin type i.e. local or remote but not the author i.e. the `person-x` prefix

akinsho:
This PR adds in an async check whilst cleaning,updating, and installing plugins i.e. places where the `find_missing_plugins` function was being used, for a git plugins remote url and checks if the repo name from the plugin matches the plugins name in memory, if it doesn't the plugin is marked as missing.

This will fix #231 

### TODO
- [x] Figure out why/how to correctly await `vim.schedule` in `find_missing_plugins`
- [x] Tidy up git command i.e. add to config?? - No because we need a consistent result here
- [x] Tidy up extraction of repo name from remote url
- [x] Ensure plugin is correctly removed/installed if url is different (this should work but just needs more testing)

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -266,9 +266,7 @@ local function args_or_all(...) return util.nonempty_or({...}, vim.tbl_keys(plug
 -- Installs missing plugins, then updates helptags and rplugins
 packer.install = function(...)
   local install_plugins
-  if ... then
-    install_plugins = {...}
-  end
+  if ... then install_plugins = {...} end
   async(function()
     if not install_plugins then
       install_plugins = await(plugin_utils.find_missing_plugins(plugins))
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -268,17 +268,18 @@ packer.install = function(...)
   local install_plugins
   if ... then
     install_plugins = {...}
-  else
-    install_plugins = plugin_utils.find_missing_plugins(plugins)
   end
-
-  if #install_plugins == 0 then
-    log.info('All configured plugins are installed')
-    packer.on_complete()
-    return
-  end
-
   async(function()
+    if not install_plugins then
+      install_plugins = await(plugin_utils.find_missing_plugins(plugins))
+    end
+
+    if #install_plugins == 0 then
+      log.info('All configured plugins are installed')
+      packer.on_complete()
+      return
+    end
+
     local start_time = vim.fn.reltime()
     local results = {}
     await(clean(plugins, results))
@@ -320,9 +321,8 @@ packer.update = function(...)
     local start_time = vim.fn.reltime()
     local results = {}
     await(clean(plugins, results))
-    local missing_plugins, installed_plugins = util.partition(
-                                                 plugin_utils.find_missing_plugins(plugins),
-                                                 update_plugins)
+    local missing = await(plugin_utils.find_missing_plugins(plugins))
+    local missing_plugins, installed_plugins = util.partition(missing, update_plugins)
 
     update.fix_plugin_types(plugins, missing_plugins, results)
     local _
@@ -369,9 +369,8 @@ packer.sync = function(...)
   async(function()
     local start_time = vim.fn.reltime()
     local results = {}
-    local missing_plugins, installed_plugins = util.partition(
-                                                 plugin_utils.find_missing_plugins(plugins),
-                                                 sync_plugins)
+    local r = await(plugin_utils.find_missing_plugins(plugins))
+    local missing_plugins, installed_plugins = util.partition(r or {}, sync_plugins)
 
     update.fix_plugin_types(plugins, missing_plugins, results)
     local _
--- a/lua/packer/clean.lua
+++ b/lua/packer/clean.lua
@@ -25,7 +25,7 @@ local clean_plugins = function(_, plugins, results)
     results.removals = results.removals or {}
     await(a.main)
     local opt_plugins, start_plugins = plugin_utils.list_installed_plugins()
-    local missing_plugins = plugin_utils.find_missing_plugins(plugins, opt_plugins, start_plugins)
+    local missing_plugins = await(plugin_utils.find_missing_plugins(plugins))
     -- turn the list into a hashset-like structure
     for idx, plugin_name in ipairs(missing_plugins) do
       missing_plugins[plugin_name] = true
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -173,6 +173,15 @@ git.setup = function(plugin)
     end)
   end
 
+  plugin.remote_url = function ()
+    return async(function ()
+      return await(jobs.run(fmt("%s remote get-url origin", config.exec_cmd),
+      {capture_output = true, cwd = plugin.install_path })):map_ok(function (data)
+        return { remote = data.output.data.stdout[1] }
+      end)
+    end)
+  end
+
   plugin.updater = function(disp)
     return async(function()
       local update_info = {err = {}, revs = {}, output = {}, messages = {}}
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -38,8 +38,8 @@ local handle_checkouts = function(plugin, dest, disp)
         disp:task_update(plugin_name, fmt('checking out %s %s...',
                                           plugin.branch and 'branch' or 'tag', branch_or_tag))
       end
-      r:and_then(await, jobs.run(config.exec_cmd
-                                   .. fmt(config.subcommands.checkout, dest, branch_or_tag), opts))
+      r:and_then(await, jobs.run(
+                   config.exec_cmd .. fmt(config.subcommands.checkout, dest, branch_or_tag), opts))
         :map_err(function(err)
           return {
             msg = fmt('Error checking out %s %s for %s', plugin.branch and 'branch' or 'tag',
@@ -54,8 +54,8 @@ local handle_checkouts = function(plugin, dest, disp)
       if disp ~= nil then
         disp:task_update(plugin_name, fmt('checking out %s...', plugin.commit))
       end
-      r:and_then(await, jobs.run(config.exec_cmd
-                                   .. fmt(config.subcommands.checkout, dest, plugin.commit), opts))
+      r:and_then(await, jobs.run(
+                   config.exec_cmd .. fmt(config.subcommands.checkout, dest, plugin.commit), opts))
         :map_err(function(err)
           return {
             msg = fmt('Error checking out commit %s for %s', plugin.commit, plugin_name),
@@ -65,19 +65,18 @@ local handle_checkouts = function(plugin, dest, disp)
         end)
     end
 
-    return r:map_ok(function(ok) return {status = ok, output = output} end):map_err(
-             function(err)
-        if not err.msg then
-          return {
-            msg = fmt('Error updating %s: %s', plugin_name, table.concat(err, '\n')),
-            data = err,
-            output = output
-          }
-        end
+    return r:map_ok(function(ok) return {status = ok, output = output} end):map_err(function(err)
+      if not err.msg then
+        return {
+          msg = fmt('Error updating %s: %s', plugin_name, table.concat(err, '\n')),
+          data = err,
+          output = output
+        }
+      end
 
-        err.output = output
-        return err
-      end)
+      err.output = output
+      return err
+    end)
   end)
 end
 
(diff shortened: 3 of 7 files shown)


lua/packer/plugin_utils.lua
```diff
@@ -96,26 +96,44 @@ plugin_utils.ensure_dirs = function()
...
+            table.insert(missing_plugins, plugin_name)
+          end
+        end
+      end
     end
-  end
 
-  return missing_plugins
+    -- TODO why do we need to await vim.schedule here
+    await(vim.schedule)
```
akinsho, line 134: @wbthomason without this line subsequent functions which use `vim.fn` or `nvim_exec` all fail. I'm not sure exactly what's going on but I found that putting this here fixed those issues 😓

akinsho:
@wbthomason despite this being a draft I wouldn't mind some feedback regarding the general approach especially regarding the use of the `async` functions I'm not entirely sure I'm using these correctly.

lua/packer.lua
```diff
@@ -263,20 +263,22 @@ local function args_or_all(...) return util.nonempty_or({...}, vim.tbl_keys(plug
 -- managed plugins.
 -- Installs missing plugins, then updates helptags and rplugins
 packer.install = function(...)
-  local install_plugins
-  if ... then
-    install_plugins = {...}
-  else
-    install_plugins = plugin_utils.find_missing_plugins(plugins)
-  end
+  local args = ...
```
nanotee, line 266: I believe this only keeps the first argument passed to `packer.install()`, is this intended?

akinsho: Nope if that's the case then definitely not the intention, this is more to get past the fact that I can't use `varargs` directly since where they are needed is now within a new function scope although, I could just collect them into a table and unpack that later e.g. `local args = {...}`

lua/packer/plugin_utils.lua
```diff
@@ -96,26 +96,44 @@ plugin_utils.ensure_dirs = function()
...
+      await(vim.schedule)
+      local guessed_type = plugin_utils.guess_dir_type(plugin_path)
+      if not plugin_installed or plugin.type ~= guessed_type then
+        table.insert(missing_plugins, plugin_name)
+      end
+      if guessed_type == plugin_utils.git_plugin_type then
+        local r = await(plugin.remote_url())
+        local remote = r.ok.remote
+        if r.ok and r.ok.remote then
+          local parts = vim.split(remote, "/")
```
akinsho, line 123: I know they are not paths but will the separators consistently be `/` on windows (I know very little about how windows behaves tbh).

lua/packer/plugin_utils.lua
wbthomason: Try `await(main)` - it's the same thing, but semantically a bit clearer. This is necessary to ensure that you're on the main loop, i.e. to do things outside of `api-fast`.

wbthomason:
> @wbthomason despite this being a draft I wouldn't mind some feedback regarding the general approach especially regarding the use of the `async` functions I'm not entirely sure I'm using these correctly.

Sorry for the radio silence here/on other issues and PRs. I'll be able to take an in-depth look in a couple of days (and merge some of the other pending changes, etc.). At a glance, your usage of the async stuff seems fine.

lua/packer/plugin_utils.lua
akinsho: This is a dump point in hindsight since it's a url not path 🤦🏾

lua/packer.lua
```diff
@@ -268,17 +268,19 @@ packer.install = function(...)
   local install_plugins
   if ... then
     install_plugins = {...}
-  else
-    install_plugins = plugin_utils.find_missing_plugins(plugins)
   end
+  async(function()
+    if not install_plugins then
```
wbthomason, line 273: Dumb Lua question: is `install_plugins` reliably falsey without explicit initialization to `nil`? I assume so?

lua/packer.lua
```diff
@@ -320,9 +322,8 @@ packer.update = function(...)
     local start_time = vim.fn.reltime()
     local results = {}
     await(clean(plugins, results))
-    local missing_plugins, installed_plugins = util.partition(
-                                                 plugin_utils.find_missing_plugins(plugins),
-                                                 update_plugins)
+    local missing = await(plugin_utils.find_missing_plugins(plugins))
+    local missing_plugins, installed_plugins = util.partition(missing, update_plugins)
```
wbthomason, line 326: Does this need the `missing or {}` check included in the `sync` case?

lua/packer/clean.lua
```diff
@@ -25,7 +25,9 @@ local clean_plugins = function(_, plugins, results)
     results.removals = results.removals or {}
     await(a.main)
     local opt_plugins, start_plugins = plugin_utils.list_installed_plugins()
-    local missing_plugins = plugin_utils.find_missing_plugins(plugins, opt_plugins, start_plugins)
+    local missing_plugins = {}
+    local r = await(plugin_utils.find_missing_plugins(plugins))
+    missing_plugins = r or {}
```
wbthomason, line 30: Nitpick, but maybe we shorten this to `local missing_plugins = await(plugin_utils.find_missing_plugins(plugins)) or {}`?

lua/packer/plugin_utils.lua
```diff
@@ -100,26 +100,43 @@ plugin_utils.ensure_dirs = function()
...
-    local plugin_path = util.join_paths(config[plugin.opt and 'opt_dir' or 'start_dir'],
-                                        plugin.short_name)
-    local plugin_installed = (plugin.opt and opt_plugins or start_plugins)[plugin_path]
+      local plugin_path = util.join_paths(config[plugin.opt and 'opt_dir' or 'start_dir'],
+      plugin.short_name)
+      local plugin_installed = (plugin.opt and opt_plugins or start_plugins)[plugin_path]
 
-    if not plugin_installed or plugin.type ~= plugin_utils.guess_dir_type(plugin_path) then
-      table.insert(missing_plugins, plugin_name)
+      await(vim.schedule)
```
wbthomason, line 118: Nitpick: Let's use `await(a.main)` for consistency. It's the same thing, but in case we have to change the async implementation in the future, I'd prefer to use `a.main`.

wbthomason (changes requested):
Thanks! This looks good overall - I've left a few nitpicky comments (mostly on style), and I'd prefer that we run `lua-format` using the repo format file before merging.

We should also add a test case for this - this behavior is complex enough that I think it's worth including in the test set to prevent regressions. Which plugins have you been using while manually testing this, @akinsho?

lua/packer.lua
akinsho: Probably 😅

lua/packer.lua
akinsho: @wbthomason I've done this quite a lot before, you don't actually need to explicitly assign to nil, any random variable reference will being assigned to nil (the local bit makes it local but would otherwise be a global nil value).

akinsho:
@wbthomason regarding formatting I've tried to run luaformatter on the project before using the local `.lua-format` and everything just gets completely ruined not sure if it's something in the config file or on my end. I don't really use `luaformatter` for anything but I did attempt setting it up for this project but it always formats things incorrectly 🤔 .

Regarding testing I had a similar thought but I was thinking that without a lot of the testing helpers/infrastructure already built it would probably require the same again or more amount of code to get this tested 😓 . Not sure if there is a simpler way to test this without having to create git repositories with similar names but different remotes since that's sort of the crux of how this works.

On the testing front I've forked a lot of projects historically so I just tested changing from my fork to the original and back

lua/packer.lua
akinsho: Actually looking at the functions implementation it always returns a table so I don't think any of the `or {}` are required.

akinsho:
Actually just realised it's much easier to test this than I though by just focusing on testing the function I changed and ensuring it's behaviour

nanotee:
I think the test failure is because of a missing "p" flag in a call to `mkdir()` in one of my previous tests (`packer_plugin_utils_spec.lua`). Not sure why CI didn't complain before, sorry about that

akinsho:
@wbthomason added some tests here now, would still be good to get someone else testing it though. It should be easy enough to test if you try and change any of your repo's with a fork. I try `akinsho/packer.nvim` to `wbthomason/packer.nvim` and that should trigger a clean and download

lua/packer/plugin_types/git.lua
```diff
@@ -173,6 +173,15 @@ git.setup = function(plugin)
     end)
   end
 
+  plugin.remote_url = function ()
+    return async(function ()
+      return await(jobs.run(fmt("%s -C %s remote get-url origin", config.exec_cmd, plugin.install_path),
+      {capture_output = true})):map_ok(function (data)
```
nanotee, line 179: Just a heads up: if `plugin.install_path` contains a space, the command will fail because `jobs.run()` uses spaces as a separator when splitting the string.

nanotee:
@akinsho I tested this locally, changing the repository works as expected

lua/packer/plugin_types/git.lua
akinsho: Thanks for the heads up 👍🏾 just switched to specifying the directory using the `cwd` argument of the `job` instead

akinsho:
Thanks for trying it out @nanotee 👍🏾

wbthomason (approved):
This is looking good and works in my local testing. Thanks @akinsho! I'm going to run lua-format on the changed files (should really make this a commit hook, though if it causes strange formatting output on your machine maybe that's not a good idea), then I think we're good to merge?

wbthomason:
Damn, as soon as I say it's working and merge, I run into an issue. I'm temporarily reverting this since roughly everything broke, sorry for the churn.

akinsho:
@wbthomason yikes sorry about that what broke when you merged it? just having a look now

wbthomason:
Still figuring it out - when I tested locally, I was just using `PackerInstall`, which seemed fine, but when I ran `PackerUpdate` after pulling the changes, it seems to try to delete all my installed plugins?

wbthomason:
This is complicated by #354

akinsho:
Hmm I see the window, I've noticed that it's sometimes surprisingly slow to appear i.e. the prompt is immediate but the window is not

wbthomason:
Ah! I see the problem, and understand why my local testing worked.

I had also been tweaking some settings in my `.gitconfig` between local testing and merging the PR. Part of the changes involved uncommenting these lines:
```
[url "ssh://git@github.com/"]
  insteadOf = https://github.com/
```
which have the effect of making the `repo_name` used in `find_missing_plugins` look like `//git@github.com/someauthor/someplugin` rather than just `someauthor/someplugin`, so everything looks like it has a different remote and gets marked for removal.

This should be a fairly easy fix, though I'm going to have to figure out the right `git-fu` to clean up the `git` history on this repo...

This also makes me wonder if we'll see problems with fully-specified URLs, e.g. `https://gitlab.com/someauthor/someplugin` in a plugin spec.

wbthomason:
> Hmm I see the window, I've noticed that it's sometimes surprisingly slow to appear i.e. the prompt is immediate but the window is not

Strange. It never seems to appear for me.

akinsho:
> This also makes me wonder if we'll see problems with fully-specified URLs, e.g. https://gitlab.com/someauthor/someplugin in a plugin spec.

@wbthomason the situation you had sounds like the exact edge case that this doesn't cover since in testing and developing this I added a check to handle ssh git remote urls which take the form `git@github.com:akinsho/dotfiles.git` vs `https:blah/user/repo` so this PR handles those cases fine. Your case though is a combination of both? so there is a `match` for `git@github.com` which presumes it will just match a url like I shared but would match yours but wouldn't be parsed properly.

A very simple fix would be to match at the start of the string for the ssh urls.

Regarding the git fu one simple dump solution is if I rebase my branch it will rewrite the history of it I can add in this extra change and we can just merge it as something different and I'll just reference this in the PR description for future purposes.

rather than trying to revert a revert etc

wbthomason:
@akinsho Yeah, what happens is that the remote URL starts with the form `ssh://git@github.com/someauthor/someplugin.git` (I think), so taking the last segment after splitting on `:` doesn't work as expected. I've implemented a fix (basically, always split by `:` *and* `/`, and take the last two chunks). I'm also wondering, though, if this is going to cause Windows issues as written - it uses `util.join_paths`, but the path separator on Windows is `\` rather than `/`, which probably won't match the plugin's name?

wbthomason:
re: git: I was thinking that maybe I rebase and force-push `master` to get rid of the PR commit and revert commit, then we somehow add the new commit fixing the issue to your branch and re-merge? This might be exactly what you're proposing; either way, I'm good with whatever you think is easiest and cleanest.

akinsho:
I think re. rebasing master without stuff like the changes in my other PR it's a pain for users when plugins rebase since packer won't handle that and in a lot of cases I end up going to the plugin's repo in my filesystem and pulling or changing packer temporarily to use `rebase=true` could just be me not sure how other people handle it but unless I'm missing some workflow it's pretty annoying 😆 my suggestion would just be to accept the history and open another branch with this change but different commits since it's rebased.

Re. using `util.join_paths` there's no need for that in hindsight was my bad since urls won't be any different on windows so it should just use `/`

wbthomason:
Ok. I'm fine with that. I'll make a new PR with the changes in just a sec, thanks.

PR #341, opened 2021-05-05
✨ Add Hiperlink

UltiRequiem:


wbthomason:
Thanks for the suggestion! I'd be OK adding hyperlinks to all the contributors named in the README, but it should probably be either nobody or everybody.

PR #342, opened 2021-05-05
Fix highlight being disabled due to redrawtimeout

### LINKED ISSUE #323: Bug: Syntax highlighting gets disabled after redrawtime
### Steps to reproduce
Run `:PackerSync` with lots of plugins
### Actual behaviour
After some time highlighting gets disabled with following messeage
> `redrawtime` exceeded, syntax highlighting disabled

IMG
### Expected behaviour
Syntax highlighting should remain all the time 

### Aditional Info
According to `:h redrawtime` If redrawing takes longer then redrawtime then syntax highlighting gets disabled . Though I can't understand why Packers redrawing will take that long . I'ven't seen it ever happen on other plugins . Can it be because of extmarks ? If it is then it might be a upstream issue . Otherwise running processes(git) in background slowing the redraw? Can probably be discarded as it never happened with VimPlug unless Packer is doing something weird :P .

shadmansaleh:
According to docs `\@<=` is really inefficient and recommends using `\\zs` instead of it . 

### Doc
```
\@<=	Matches with zero width if the preceding atom matches just before what
	follows. |/zero-width|
	Like "(?<=pattern)" in Perl, but Vim allows non-fixed-width patterns.
	Example			matches ~
	\(an\_s\+\)\@<=file	"file" after "an" and white space or an
				end-of-line
	For speed it's often much better to avoid this multi.  Try using "\zs"
	instead |/\zs|.  To match the same as the above example:
		an\_s\+\zsfile
	At least set a limit for the look-behind, see below.

	"\@<=" and "\@<!" check for matches just before what follows.
	Theoretically these matches could start anywhere before this position.
	But to limit the time needed, only the line where what follows matches
	is searched, and one line before that (if there is one).  This should
	be sufficient to match most things and not be too slow.

	In the old regexp engine the part of the pattern after "\@<=" and
	"\@<!" are checked for a match first, thus things like "\1" don't work
	to reference \(\) inside the preceding atom.  It does work the other
	way around:
	Bad example			matches ~
	\%#=1\1\@<=,\([a-z]\+\)		",abc" in "abc,abc"

	However, the new regexp engine works differently, it is better to not
	rely on this behavior, do not use \@<= if it can be avoided:
	Example				matches ~
	\([a-z]\+\)\zs,\1		",abc" in "abc,abc"

```

### Changes 
I've replace the `\@<=` with `\zs` in syntax rules. I've cleaned all the plugins and installed them in those 29 seconds highlight didn't disapear so That's an improvement :D

For packerStatusSuccess and packerStatusFail I've limited their look behind to 5 bytes . Couldn't get them to work with `\zs` . I'm assuming it's because both `^ working_sym` is already matches packerSuccess. Navertheless looking behind 5 bytes instead of till start of file still better

I'm actually not sure how to test all possible highlights. So anyone reviewing please test it out and let me know if any syntax highlighting is off.

@akinsho can you check if you can repoduce [it](https://github.com/wbthomason/packer.nvim/issues/299#issuecomment-824208219) in this branch ?

closes #323

### DIFF
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -590,30 +590,39 @@ display_mt.__index = display_mt
 
 -- TODO: Option for no colors
 local function make_filetype_cmds(working_sym, done_sym, error_sym)
+  local function look_back(str)
+    return string.format([[\(%s\)\@%d<=]], str, #str)
+  end
   return {
     -- Adapted from https://github.com/kristijanhusak/vim-packager
     'setlocal buftype=nofile bufhidden=wipe nobuflisted nolist noswapfile nowrap nospell nonumber norelativenumber nofoldenable signcolumn=no',
-    'syntax clear', 'syn match packerWorking /^ ' .. working_sym .. '/',
+    'syntax clear',
+    'syn match packerWorking /^ ' .. working_sym .. '/',
     'syn match packerSuccess /^ ' .. done_sym .. '/',
     'syn match packerFail /^ ' .. error_sym .. '/',
-    'syn match packerStatus /\\(^+.*—\\)\\@<=\\s.*$/',
-    'syn match packerStatusSuccess /\\(^ ' .. done_sym .. '.*\\)\\@<=\\s.*$/',
-    'syn match packerStatusFail /\\(^ ' .. error_sym .. '.*\\)\\@<=\\s.*$/',
-    'syn match packerStatusCommit /\\(^\\*.*—\\)\\@<=\\s.*$/',
+    'syn match packerStatus /^+.*—\\zs\\s.*$/',
+    'syn match packerStatusSuccess /' .. look_back('^ ' .. done_sym) .. '\\s.*$/',
+    'syn match packerStatusFail /' .. look_back('^ ' .. error_sym) .. '\\s.*$/',
+    'syn match packerStatusCommit /^\\*.*—\\zs\\s.*$/',
     'syn match packerHash /\\(\\s\\)[0-9a-f]\\{7,8}\\(\\s\\)/',
-    'syn match packerRelDate /([^)]*)$/', 'syn match packerProgress /\\(\\[\\)\\@<=[\\=]*/',
+    'syn match packerRelDate /([^)]*)$/',
+    'syn match packerProgress /\\[\\zs[\\=]*/',
     'syn match packerOutput /\\(Output:\\)\\|\\(Commits:\\)\\|\\(Errors:\\)/',
     [[syn match packerTimeHigh /\d\{3\}\.\d\+ms/]],
     [[syn match packerTimeMedium /\d\{2\}\.\d\+ms/]],
     [[syn match packerTimeLow /\d\.\d\+ms/]],
     [[syn match packerTimeTrivial /0\.\d\+ms/]],
     [[syn match packerPackageNotLoaded /(not loaded)$/]],
-    [[syn match packerPackageName /\(^\ • \)\@<=[^ ]*/]],
+    [[syn match packerPackageName /^\ • \zs[^ ]*/]],
     [[syn match packerString /\v(''|""|(['"]).{-}[^\\]\2)/]],
     [[syn match packerBool /\<\(false\|true\)\>/]],
-    'hi def link packerWorking        SpecialKey', 'hi def link packerSuccess        Question',
-    'hi def link packerFail           ErrorMsg', 'hi def link packerHash           Identifier',
-    'hi def link packerRelDate        Comment', 'hi def link packerProgress       Boolean',
+
+    'hi def link packerWorking        SpecialKey',
+    'hi def link packerSuccess        Question',
+    'hi def link packerFail           ErrorMsg',
+    'hi def link packerHash           Identifier',
+    'hi def link packerRelDate        Comment',
+    'hi def link packerProgress       Boolean',
     'hi def link packerOutput         Type'
   }
 end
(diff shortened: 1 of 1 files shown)


akinsho:
@shadmansaleh thanks for digging into this 👍🏾 had no idea about the speed of `\@<=` I'll test it out today and see if it resolves the issue.

lua/packer/display.lua
```diff
@@ -593,27 +593,33 @@ local function make_filetype_cmds(working_sym, done_sym, error_sym)
...
+    'syntax clear',
+    'syn match packerWorking /^ ' .. working_sym .. '/',
     'syn match packerSuccess /^ ' .. done_sym .. '/',
     'syn match packerFail /^ ' .. error_sym .. '/',
-    'syn match packerStatus /\\(^+.*—\\)\\@<=\\s.*$/',
-    'syn match packerStatusSuccess /\\(^ ' .. done_sym .. '.*\\)\\@<=\\s.*$/',
-    'syn match packerStatusFail /\\(^ ' .. error_sym .. '.*\\)\\@<=\\s.*$/',
-    'syn match packerStatusCommit /\\(^\\*.*—\\)\\@<=\\s.*$/',
+    'syn match packerStatus /^+.*—\\zs\\s.*$/',
+    'syn match packerStatusSuccess /\\(^ ' .. done_sym .. '\\)\\@5<=\\s.*$/',
```
akinsho, line 601: OOI why specifically 5 for the number of bytes to check here? only just began reading through the docs on this seems good to have the limit just curious about how you arrived at 5?

shadmansaleh: Lowest it worked on was 4 + 1 . Nothing scientific about it lol

akinsho:
Been using this all day and haven't hit the redraw issue so seems good, but tbh it used to happen quite infrequently in my case. I did try removing and re-installing all my plugins and didn't hit the issue as well.

shadmansaleh:
Well it happed almost every `PackerSync` for me (alow device + emuluation) . With this haven't seen it yet :)

lua/packer/display.lua
akinsho: Could this then recur for users who have changed their `done_sym` and `err_sym` to other bigger characters since a user can configure this? The docs seemed to imply there was a way to do this but limiting the check to the current line or would that still bring on this bug?

shadmansaleh: Maybe we can check len of `done_sym` and `err_sym`?

akinsho: Lua's len function returns the size in bytes so seems like it'd work for this case. Read through the docs and seems the default i.e. what we already had was the one that is limited to the current line and the line above but seems ambiguous about it being slow and was already causing this issue so is best avoided

shadmansaleh: Ok . @akinsho  Now the highlight should work no mater what user has in config :)

shadmansaleh: > Lua's len function returns the size in bytes so seems like it'd work for this case.

I learned that when `string.sub()` managed to split a multibyte char in two :P

wbthomason (approved):
Thanks! I added a couple of commits fixing a style nitpick and running lua-format, but this looks good to me. Sorry for the delay in reviewing.

akinsho:
Just a heads up @shadmansaleh I don't think this fully works, just saw the error recur during a packer sync with this  commit in place 😭

shadmansaleh:
> Just a heads up @shadmansaleh I don't think this fully works, just saw the error recur during a packer sync with this commit in place 😭

That's unfortunate . At least it's a lot rarer now . Seeing your comment I lowered the 'redrawtime' in my device and for me the highlights survived upto redrawtime set to 400 . Do you know any other optimization we could do ? `syn-nextgroup` and `syn-contained` looks interesting

akinsho:
I use the default redraw time of `2000` which is much larger than your example and it still occurs but a bit of a moot point. It did seem to be resolved or maybe that was just luck for me before we changed the amount used based on the icon size, although this really isn't scientific since I used it for like an hour or so and didn't see the bug meanwhile I also didn't see it with this commit till just now and it's been merged in for a day.

Haven't looked into any syntax highlighting stuff in a while thankfully post treesitter but those look interesting or alternatively as ever having a look at what vim plug does with it's highlighting for inspiration

shadmansaleh:
> I use the default redraw time of 2000 which is much larger than your example and it still occurs but a bit of a moot point.

I use 2000 too . I didn't see it occur for me yet that's why when I saw your comment I tried with lower values. 

> It did seem to be resolved or maybe that was just luck for me before we changed the amount used based on the icon size, although this really isn't scientific since I used it for like an hour or so and didn't see the bug meanwhile I also didn't see it with this commit till just now and it's been merged in for a day.

I guess we'll just have to see how often it happens now 

Since there was now proper way to reproduce it on the first place no way to know if it's been fixed or not .

> having a look at what vim plug does with it's highlighting for inspiration

I can see lots of `contained` & `contains` in [vimPlug](https://github.com/junegunn/vim-plug/blob/fc2813ef4484c7a5c080021ceaa6d1f70390d920/plug.vim#L804-L855)

Makes sense currently all rules are tried against entire buffer . While `contained` rules are only tried against the block that says it `contains` them . That should improve performence.

shadmansaleh:
@akinsho did you see the issue reoccur after that time ? It still hasn't happened for me yet.

akinsho:
@shadmansaleh it still happens very rarely for me, like once every 2 weeks or more it's much less common now.

par4m:
I know it's been almost 1 year but I'm still experiencing this issue, not sure why
image

PR #343, opened 2021-05-05
Enhance: don't panic on config error at startup

shadmansaleh:
### Prevent packer from panicking for config error

Currently if a error occurs while running config for a startup plugin. Packer panics and doesn't load any other plugin, neither does create lazy loading bindings . This patch changes that . With this packer will show an error message for failure and go on with usual business .

There's no point in throwing an error when call to `try_loadstring` is unprotected . Instead error message can be displayed with `nvim_echo` and error is gracefully handled inside `try_loadstring`. Also in my opinion `try` should mean it anticipates an error and handles it without causing a new one :P

### Motivation

I found current behavior really annoying especially when experimenting with plugins . As an error in one plugin just takes down the entire setup .

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -89,8 +89,9 @@ local try_loadstring = [[
 local function try_loadstring(s, component, name)
   local success, result = pcall(loadstring(s))
   if not success then
-    print('Error running ' .. component .. ' for ' .. name)
-    error(result)
+    vim.schedule(function()
+      vim.api.nvim_notify('Packer: Error running ' .. component .. ' for ' .. name..'\n        '..result, vim.log.levels.ERROR, {})
+    end)
   end
   return result
 end
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -90,7 +90,7 @@ local function try_loadstring(s, component, name)
   local success, result = pcall(loadstring(s))
   if not success then
     vim.schedule(function()
-      vim.api.nvim_notify('Packer: Error running ' .. component .. ' for ' .. name..'\n        '..result, vim.log.levels.ERROR, {})
+      vim.api.nvim_notify('packer.nvim: Error running ' .. component .. ' for ' .. name .. ': ' .. result, vim.log.levels.ERROR, {})
     end)
   end
   return result


akinsho (approved):
This looks good to me, I definitely find it super annoying to have one plugin collapse my entire setup 👍🏾 , this needs to be vetted by @wbthomason as he might have had a reason for wanting it like this. FWIW I agree that trying to load a plugin should be a best attempt and shouldn't abort everything.

I guess one concern with trying to go on is that since some plugins are linked e.g. required by others etc. allowing things to go on could lead to cascading error messages else where but I assume these would just do the same the first plugin did i.e. fail print an error and let everything else carry on

shadmansaleh:
> I guess one concern with trying to go on is that since some plugins are linked e.g. required by others etc. allowing things to go on could lead to cascading error messages else where but I assume these would just do the same the first plugin did i.e. fail print an error and let everything else carry on

Even in that case the cascading error will be isolated to that group and the rest will be fine .

Also does `after` & `require` have affect for non opt plugins ? Since loading startup plugins is out of hands of packer and are directly loaded by vim-package .

akinsho:
I believe `after` will force a plugin to be `opt` and `require` is only relevant for installing and removing plugins, it was more just a random example of linkages causing a series of plugins to fail especially if things like the groups feature that there are a few issues for gets implemented, but as we've mentioned that won't really matter since all affected plugins will error.

As a complete tangent I wonder if erroring in this plugin should use `vim.notify` or `nvim_notify` which can now be overriden by a user and is slowly being used to replace native lsp errors that way a user can override them. I've overriden this in my config to show a float instead since I find the commandline area spam and the press enter to continue prompts super annoying.

shadmansaleh:
> As a complete tangent I wonder if erroring in this plugin should use vim.notify or nvim_notify which can now be overriden by a user and is slowly being used to replace native lsp errors that way a user can override them. 

Can `vim.notify` specify highlight groups ? I can't find docs for it anywhere . I chose `nvim_echo` as with that the message can be highlighted with `ErrorMsg` looking more like an error then a normal info :)  .

> I've overriden this in my config to show a float instead since I find the commandline area spam and the press enter to continue prompts super annoying.

Will your overwrite work as at this stage. As neovim is still starting up and the config might yet not be loaded ?

akinsho:
`vim.notify` is higher level than the implementation of `nvim_echo` it gets a `vim.notify(msg: string, level: int, opts: future-use)` it's more of an idle though for the future. It would be nice to see it being more widely adopted so users can control how neovim errors/info are presented through one place.

Re. if my override would work, I believe even start packages are loaded after the rest of your `.vimrc/init.vim/init.lua`, from `:h packages`

> When Vim starts up, after processing your .vimrc, it scans all directories in
'packpath' for plugins under the "pack/*/start" directory.  First all those
directories are added to 'runtimepath'.  Then all the plugins are loaded.

shadmansaleh:
> Re. if my override would work, I believe even start packages are loaded after the rest of your .vimrc/init.vim/init.lua, from :h packages

kind of not related as packer calls them from `plugin/` directory . Though I checked `plugin/` gets called after init . I was asking because I've configured vim.notify to open a float at top right corner . But that doesn't work at startup . If I replace the calls to `nvim_echo` with equavelent nvim_notify calls . My implementation of vim.notify gets called the message is written to the buffer but I don't see a float .Basically it becomes a silent error unless I go checking the log . I'm assuming float cann't be opened because tui isn't fully initiated yet. Can you check if your implementation of `vim.notify` work ?

```diff
-    vim.api.nvim_echo({
-                        {'Error running ' .. component .. ' for ' .. name..'\n', 'ErrorMsg'},
-                        {result, 'ErrorMsg'}}, true, {})
+    vim.api.nvim_notify('Error running ' .. component .. ' for ' .. name..'\n', 1, {})
+    vim.api.nvim_notify(result, 1, {})
```

akinsho:
@shadmansaleh I'll give it a try but tbh I think regardless messages like this should be deferred in either case since currently they block vim from starting till you hit enter which I find particularly frustrating. I want to know that something has gone wrong sure but the experience of starting your editor and it opening to a blank screen except for red text crawling up the screen is pretty rubbish IMO. 

I think it would be nicer if the the call where wrapped in a `defer_fn` and waited till vim had started before giving me the bad news rather than stressing me out with error messages immediately it opens
```lua
vim.defer_fn(function()
 --- Stuff
end, 3000)
```

shadmansaleh:
> I think it would be nicer if the the call where wrapped in a defer_fn and waited till vim had started before giving me the bad news rather than stressing me out with error messages immediately it opens

I think this is a good idea . I've just wraped it with `vim.schedule` . That way the message will be immediately available when nvim opens and we don't need to guess how long startup time will take.

Also I looked at default implementation of vim.notify . It does treat error messages properly . So I've switched to that :)

wbthomason:
@shadmansaleh @akinsho Thank you! This looks great. Tbh I have no recollection of why I made `try_loadstring` re-throw the error; in retrospect that seems like a stupid choice. Also, my apologies for letting this PR fall into a black hole for a month.

I've pushed a commit with some minor formatting changes to the error message. @shadmansaleh, if you're OK with the current format, I'm happy to merge this. Thanks again!

shadmansaleh:
> I've pushed a commit with some minor formatting changes to the error message. @shadmansaleh, if you're OK with the current format, I'm happy to merge this. Thanks again!

👍🏽 . Thanks for looking at it . This format is fine .

wbthomason:
Great, thanks again!

wbthomason:
There is one addition to this feature that I'd like to add: although this fixes the `try_loadstring` case, it does nothing if someone has an error in a setup/config key specified as a string. Handling that case seems trickier; it would be pretty gratuitous to wrap every such key in its own anonymous function, but Lua only provides `pcall` for catching errors (as far as I'm aware), limiting what we can do with just blocks of code at top level.

wbthomason:
Maybe it's best to leave this case be for now, since specifying a string for config/setup ought to be obviated once we move away from compiling the setup/config code, as mentioned in #331

akinsho:
@wbthomason maybe you can slightly hijack the `xpcall` function to use here since that function also takes an `error_handler_function` argument it's intended to be used to get more information from the stack trace but think you could also use it to do schedule an error message or something.

eg. [1.](https://www.lua.org/pil/8.5.html), [2.](https://stackoverflow.com/questions/30125726/how-to-use-xpcall-with-a-function-which-has-parameters)

shadmansaleh:
@wbthomason  @akinsho  #391

PR #344, opened 2021-05-05
refactor(plugin_utils): use `cwd` instead of `cd` 

nanotee:
This PR changes how `packer` handles the current directory for post-update hooks by using the `cwd` option rather than `cd`-ing into the directory

This is mainly for compatibility across platforms since not all shells support the `&&` operator (IIRC older versions of `fish` didn't).

(Note: it might be useful to investigate the `'shell'` and `'shellcmdflag'` options to make this even more portable)

About the test:

It was not obvious to me how to test an async function, so I ended up leveraging `plenary`'s async test functions. That means I'm using `packer`'s `await()` function inside a `plenary` `Future`. Now, this works (probably because both are sugar over coroutines with similar APIs), but it might be too much of a hack for your taste

@akinsho I'd be curious to know whether this PR improves the situation you describe in #338

### DIFF
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -174,10 +174,8 @@ plugin_utils.post_update_hook = function(plugin, disp)
             stdout = jobs.logging_callback(hook_output.err, hook_output.output, nil, disp,
                                            plugin_name)
           }
-          local cmd = {
-            os.getenv('SHELL'), '-c', 'cd ' .. plugin.install_path .. ' && ' .. plugin.run
-          }
-          return await(jobs.run(cmd, {capture_output = hook_callbacks})):map_err(
+          local cmd = {os.getenv('SHELL') or vim.o.shell, '-c', plugin.run}
+          return await(jobs.run(cmd, {capture_output = hook_callbacks, cwd = plugin.install_path})):map_err(
                    function(err)
               return {
                 msg = string.format('Error running post update hook: %s',
--- a/lua/packer/jobs.lua
+++ b/lua/packer/jobs.lua
@@ -180,6 +180,8 @@ local run_job = function(task, opts)
       options.timeout = 1000 * opts.timeout
     end
 
+    options.cwd = opts.cwd
+
     local stdin = loop.new_pipe(false)
     options.args = {unpack(task, 2)}
     options.stdio = {stdin, stdout, stderr}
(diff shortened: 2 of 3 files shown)


akinsho:
@nanotee this looks great 🤞🏾 I'll give it a try and report back. Might not be till tomorrow since I won't have much time this evening.

akinsho:
Just had a quick 5min go with this PR on my mac and seems to have resolved the issue i.e. not seeing extra directories in the `opt` dir. Will try again tomorrow  for longer when I have a more time presumably one correct install is good enough but not sure if other stuff might have been triggering it so will see.

akinsho:
Actually been using this for the last day and the issue has come back 😭 , so maybe was not related to this, not sure. 

Unfortunately the logs don't include what run scripts were executed.

Maybe that would indicate what's going wrong 😞 , really strange since I use zsh on my linux and mac machines and this only happens on the mac. The also use the same version of `zsh` but for each platform so not sure what the differences are.

nanotee:
Well, it was worth a try, at least we can rule out `cd` as being the cause (thanks for testing this!)

It might be a good idea run `shellcheck` on the install script to see if it has issues with unquoted variables and word splitting

lua/packer/plugin_utils.lua
```diff
@@ -174,10 +174,8 @@ plugin_utils.post_update_hook = function(plugin, disp)
             stdout = jobs.logging_callback(hook_output.err, hook_output.output, nil, disp,
                                            plugin_name)
           }
-          local cmd = {
-            os.getenv('SHELL'), '-c', 'cd ' .. plugin.install_path .. ' && ' .. plugin.run
-          }
-          return await(jobs.run(cmd, {capture_output = hook_callbacks})):map_err(
+          local cmd = {os.getenv('SHELL') or vim.o.shell, '-c', plugin.run}
```
QiBaobin, line 177: shall we use vim.o.shell as default?

nanotee: Ah, right, that does look a bit weird. My reluctance to use `vim.o.shell` by default comes from the fact that the option can contain spaces and quoted arguments, so it'd have to be tokenized properly so it can be split (`:help shell-unquoting` has more info on this). Neovim has an internal function to parse this option (`shell_build_argv()`) but I thought it was a bit overkill to try and make a Lua version of it, at least in this PR. Right now `'shell'` is used as a fallback when there is no `$SHELL` environment variable (the test actually failed on GitHub CI for that very reason)

wbthomason (approved):
Looks great, thank you @nanotee!

PR #345, opened 2021-05-05
Fix incorrect argument passed to util.float

akinsho:
This was raised in https://github.com/wbthomason/packer.nvim/pull/325#issuecomment-832874005 that an unexpected argument is passed to `util.float` when used as the `open_fn` for packer's display. This PR fixes this issue by checking the `opts` args type before trying to use it in the deep extend function

### DIFF
--- a/lua/packer/util.lua
+++ b/lua/packer/util.lua
@@ -110,6 +110,14 @@ util.float = function(opts)
   local left = math.ceil((columns - width) * 0.5)
   local top = math.ceil((lines - height) * 0.5 - 1)
 
+  --- TODO: this is an impromptu fix for
+  --- https://github.com/wbthomason/packer.nvim/pull/325#issuecomment-832874005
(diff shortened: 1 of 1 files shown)


akinsho:
@clason can you please try this branch to see if it fixes the issue you encountered.

gbrlsnchs:
I was about to open this same PR, so thanks for fixing it. Yeah, unexpected `[packer]` string. :disappointed_relieved:

akinsho:
Once I've verified that it works will try to get the hotfix in so as not to break things for everyone. @wbthomason FYI, I've put this in as a temporary fix so things aren't broken not sure what we want to do about the string in the long run.

akinsho:
Okay tested it on my system and seems to work fine, just need some 3rd party corroboration then will merge 😄

clason:
Yep, works here, too! Thanks!

PR #347, opened 2021-05-05
fix(config): escape brackets in open_cmd

### LINKED ISSUE #346: Packer commands generate errors
bug

An error gets generated whenever I use ```:PackerInstall```, ```:PackerUpdate```, or ```PackerSync```. Is this something to do with the number of plugins as the error says ```too many file names```.

For reference, I have around 22 plugins installed.

```
NVIM v0.5.0-dev+1231-g48e805728
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

```
OS: Ubuntu 20.04
Terminal emulator: Kitty
Shell: zsh
git version 2.25.1
```

nanotee:
Fixes #346


akinsho:
I wonder if the name should just be changed to something else given this issue like `Packer` or something. All that escaping might seem a bit daunting to someone just starting out. In either case it's good to fix this 👍🏾

wbthomason (approved):
Thanks! This is definitely an edge case I have never encountered...

I'm fine with renaming the window to avoid this issue, but this escaping seems fine for now.

Issue #253, opened 2021-03-13
Feature Request: Add Retry Option


ndavidq0:
Sometimes a plugin (or multiple plugins) might fail to update/install.
It would make sense to add a retry option (perhaps mapped to `R`, since `r` is already mapped to revert).

PR #348, opened 2021-05-08
Feature: add retry option #253

tiagovla:
This adds the retry option only if an update or install failed.


akinsho:
@tiagovla thanks for the PR 🚀 just had a look a through and it looks good. One thing I notice though is that this always tries to re-install everything or update everything all over again rather than just what failed. If a user has a lot of plugins then that will be a lot of pointlessly duplicated work.

Maybe another approach here would be to keep track of a list of what plugins failed and pass just those plugins to `install` and `update` so only those get retried?

Not sure but think/expect `install` and `update` can take a list of plugins 🤞🏾

tiagovla:
@akinsho I optimized it to take into account only the failed updates. Correct me if I'm wrong on this, but for the install process it will try to install again only the failed ones found by ``plugin_utils.find_missing_plugins(plugins)``, so no changes needed?

lua/packer/display.lua
```diff
@@ -101,6 +101,10 @@ local keymaps = {
   prompt_revert = {
     rhs = '<cmd>lua require"packer.display".prompt_revert()<cr>',
     action = 'revert an update'
+  },
+  retry = {
+    rhs = '<cmd>lua require"packer.display".retry()<cr>',
+    action = 'retry'
```
wbthomason, line 107: Nitpick: could we be more specific with the action name, i.e. "retry failed operations"?

lua/packer/display.lua
```diff
@@ -707,6 +721,14 @@ display.prompt_revert = function()
   if display.status.disp then display.status.disp:prompt_revert() end
 end
 
+display.retry = function()
+  if display.status.any_failed_install then
+      require("packer").install()
+  elseif #display.status.failed_update_list > 0 then
+      require("packer").update(unpack(display.status.failed_update_list))
```
wbthomason, line 728: Is there a reason to build the failed updates list but not do the same for installs? Both `packer.install` and `packer.update` should take a list of plugins to operate upon. It's not a big deal for installs, but it does save some effort i.e. finding missing plugins.

wbthomason (approved):
Thanks! This looks mostly great - I left two small comments (about naming the action and passing a list of failed installs as well as failed updates).

lua/packer/display.lua
tiagovla: I was not sure what to do in case luarock packages failed and I cannot test it because I never used those. So, since ``packer.install`` already finds all missing installations, hopefully including luarock packages, I left the default one.

tiagovla:
I fixed the conflicts and formatting. @akinsho

XVilka:
Are there any updates? It's a major pain-point for people with a poor connection and many plugins. Sometimes with Plug I have to retry failed to updated plugins for up to 6 times per update session.

wbthomason (approved):
LGTM, thank you! Sorry that this one languished for so long; it fell off my radar.

PR #352, opened 2021-05-12
fix(display): setf should after content is loaded

kevinhwang91:
Before patch:
image

After patch:
image


akinsho:
@kevinhwang91 thanks for the PR and the fix 👍🏾

PR #353, opened 2021-05-12
fix(display): use `close` instead of `pclose`

kevinhwang91:
AFAIK, pclose usually used outside preview window.
Buffer map q to pclose is not necessary when we're inside preview window.

And it doesn't work when move preview window to new tabpage by `<C-w>T`.


akinsho (approved):
LGTM 👍🏾 thanks for the fix

wbthomason:
Thanks @kevinhwang91 for the fix and @akinsho for the review and merge!

PR #364, opened 2021-05-20
Feature: allow run key to be a table of mixed types of hook

### LINKED ISSUE #355: MIxed post-install/update hooks
I have a question:

CHADTree plugin needs to execute `python3 -m chadtree deps` as shell command in plugins directory which I can do by: 
```run = 'python3 -m chadtree deps'```
 and after this need to call vim command `CHADdeps`, which I can do as: 
```run =  ':CHADdeps'```
But how to call both commands in turn?

wbthomason:
Fix #355: Allow for `plugin.run` to be a table containing a mix of functions, strings, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,7 +331,7 @@ use {
   tag = string,                -- Specifies a git tag to use
   commit = string,             -- Specifies a git commit to use
   lock = boolean,              -- Skip this plugin in updates/syncs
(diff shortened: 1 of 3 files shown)


wbthomason:
This works as expected in my limited manual testing, so I'm going to go ahead and merge. @anuvyklack, if you still experience issues with this feature, please let me know!

PR #369, opened 2021-05-23
refactor(local): use libuv to create symlinks

nanotee:
This PR replaces the use of external programs to create symlinks with the built-in `vim.loop.fs_symlink()` and adds a test for local plugins

EDIT: I've verified that this works on Windows 8.1 while running neovim-qt as administrator. I'm actually unsure whether developer mode on Windows 10 allows all programs to create symlinks or just `mklink`.

EDIT2: Apparently `libuv` calls the `CreateSymbolicLink` API with the `SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE` flag so this should work in theory, but I don't have a copy of Windows 10 to test this with

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,8 @@ Have a problem or idea? Make an [issue](https://github.com/wbthomason/packer.nvi
 ## Requirements
 - You need to be running Neovim v0.5.0+; `packer` makes use of extmarks and other newly-added Neovim
   features.
-- If you are on Windows 10, you need developer mode enabled in order to use local plugins (`packer`
-  needs to use `mklink`, which requires admin privileges - credit to @TimUntersberger for this note)
+- If you are on Windows 10, you need developer mode enabled in order to use local plugins (creating
+  symbolic links requires admin privileges on Windows - credit to @TimUntersberger for this note)
 
 ## Quickstart
 To get started, first clone this repository to somewhere on your `packpath`, e.g.:
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -26,7 +26,7 @@ REQUIREMENTS                                     *packer-intro-requirements*
   on this to detect whether you are running Windows or a Unix-like OS (for path
   separators)
 - If you are on Windows 10, you need developer mode enabled in order to use
-  local plugins (`packer` needs to use `mklink`, which requires admin privileges
+  local plugins (creating symbolic links requires admin privileges on Windows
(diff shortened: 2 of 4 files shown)


wbthomason (approved):
This looks great, thanks. I am in general all for moving toward `libuv`-based functionality rather than using external programs.

I also unfortunately do not have a Windows 10 environment to test this in. It's probably OK to merge, but I'm going to see if we can solicit Windows testers in the `packer` Gitter first. If nobody has responded within a day, we'll merge and hope for the best.

I don't think it needs to be part of this PR, but I should augment the test workflow to also run on Windows and macOS environments, to prevent this uncertainty in the future.

nanotee:
Thanks for the feedback. This probably isn't super urgent since it doesn't fix bugs or add new features. I'm familiarizing myself with the codebase in the process

akinsho:
@nanotee I just noticed that the new test files are all prefixed with `packer_` I had just created a `plugin_utils_spec.lua` then noticed there was already a packer prefixed version.

What do you think about dropping the prefixes? I don't really think they are necessary since tbh depending on how complex things get we can actually just match packers own structure using directories (rather than prefixes) so tests are always in a predictable place?

But I don't think that's necessary to start but I do think it will be easier to find test files if they just match the actual filename + `_spec.lua`?

nanotee:
I wasn't sure if there was a specific naming convention so I stuck to `packer_*_spec.lua`, dropping the prefix definitely makes sense to me :+1:

wbthomason:
We haven't gotten any Win10 testers, but (given that @nanotee tested on Win8, at least), I think this is OK to merge. @nanotee, any reason you'd like to hold off for now?

nanotee:
I'd like to rebase this PR on `master` once #340 gets merged since there were a few changes in the tests

wbthomason:
@nanotee Please see the discussion on #340: I discovered a bug after merging and quickly reverted, so we'll need to fix the bug, re-merge, and then cleanup the git history.

wbthomason:
@nanotee the new PR to rebase onto is #374

nanotee:
Okay, rebase done. `packer_plugin_utils_spec.lua` should probably get merged into `plugin_utils_spec.lua` at some point but I thought it was out of scope for this PR.

wbthomason:
Sorry for taking so long to merge this!

Issue #298, opened 2021-04-16
[Feature request] Support package snapshot


weihanglo:
Sometimes upgrading packages may results in conflicts between packages. By having a snapshot of all current packages with exact commits, we can restore from the snapshot without much labor.

Reference: [vim-plug](https://github.com/junegunn/vim-plug/blob/cffcfe150bda75177920530c5f1911b52f70a065/doc/plug.txt#L201) supports snapshot.

wbthomason:
Thanks for the idea and reference! I'd be happy to help guide a PR implementing this (it should be pretty straightforward), if you or anyone else is interested.

weihanglo:
Yes, I can help! It would be awesome to have some instruction.

c3n21:
Since already a month has passed and there isn't any progress I would like to step-in and make this feature (since I need this too).

The idea I had is the following:
- create a **PackerSnapshot** command which can take an optional argument which will be the name of the snapshot (all the plugins will be snapshotted except the disabled ones)
- create a **PackerRollback** command which takes an argument and what it does is rollback to the given snapshot (with autocompletion for the snapshots' names which is nice)
- create a **PackerDelete** command which deletes a snapshot given the name
- in the config we could add an option to enable automatic snapshot after each **PackerInstall** or **PackerUpdate** or **PackerClean**
- eventually add a timer which deletes old snapshots created automatically after installs, updates and cleans (so snapshots created with a name will be saved permanently unless the user wants to delete it)

How does this sound to you @wbthomason @weihanglo ?

bew:
_me, dropping in from nowhere_

Imo this gives too much intelligence to the snapshot system.

Taking inspiration from vim-plug, I like how it only generates a simple vim file with 1 line for each plugin, specifying it's name and the revision.
And the file content is in a scratch buffer, it's not saved anywhere the plugin knows about. If I want to keep it I can hit `:w some_place.vim` and manage the snapshot file however I want.

(here is an example snapshot: https://github.com/bew/dotfiles/blob/0d74848b7d8fde734c2ad47e7b4e78e27f43a3b0/nvim/plugins.lock) 

And when you want to restore a snapshot, you simply have to `:source some_place.vim` and run PlugInstall, this will source the snapshot and fill a special vim dictionary, that is used first when installing packages and choosing a revision.

What do you think of this?

c3n21:
> _me, dropping in from nowhere, hi! wave_
> 
[...]

PR #370, opened 2021-05-24
[Feature request] Support package snapshot #298

c3n21:
~~It's just a first implementation and I only tested it with spec(1), namely:~~

~~> spec can be a table with a function as its first element and config overrides as another element: 
`packer.startup({function() use 'tjdevries/colorbuddy.vim' end, config = { ... }})`~~

~~## Usage~~
~~### Make snapshot~~
~~To make a snapshot you just have to run `:PackerSnapshot <snapshot name>`~~

~~### Restore snapshot~~
~~1. To restore a snapshot you have to add **snapshot** key in the **config** table you pass to packer.startup.~~
~~So, if I want to restore a snapshot named **foo** I would have something like this:~~

~~`packer.startup(~~
~~{function() use 'tjdevries/colorbuddy.vim' end, ~~
~~config = { snapshot='foo' })`~~

~~You can also specify the default directory in which retrieve and save snapshots with the following:~~
~~`packer.startup(~~
~~{function() use 'tjdevries/colorbuddy.vim' end, ~~
~~config = { snapshot='foo', snapshot_path='path/to/snapshots' })`~~

~~By default the snapshot's location is in **~/.config/nvim/plugin**~~

~~2. Source the file `:luafile %`~~
~~3. `:PackerUpdate`~~
~~4. Restart NeoVim and enjoy~~

~~If you want to update you will need to remove **snapshot** from **config** table, and then source the file and update.~~

# Final Implementation

### Make snapshot
To make a snapshot you just have to run `:PackerSnapshot <snapshot name>`; you can also append single plugins name to selectively snapshot only the plugins you want like `:PackerSnapshot <snapshot name> <plugin_1> <plugin_2>`

### Restore snapshot
To rollback to a snapshot you have two ways:

1. Add **config.snapshot = 'some snapshot'**. You can specify either the `snapshot name` which will be fetched inside **config.snapshot_path** (the default snapshot directory which is `~/.cache/nvim/packer` but you can change it) or an absolute path.
So, if I want to restore a snapshot named **foo** I would have something like this:

`packer.startup(
{function() use 'tjdevries/colorbuddy.vim' end, 
config = { snapshot='foo' })`

In this way, every time you boot up NeoVim it will always load that snapshot.

2. You can also run `:PackerRollback` which updates your installed plugins' commit to the snapshot you specified. ~~After that you just need to run `:PackerUpdate` and it will revert back~~

### Delete snapshot

1. Just run `PackerDelete <snapshot name>` , you can either provide only the snapshot name or an absolute path.

#### Misc:

- I've also added a packer.snapshot_complete() so that `PackerRollback` and `PackerDelete` can prompt which snapshots are available.
- The snapshot format is a Lua table serialized using `vim.inspect`
- `packer.rollback` hard resets the git repo of the given plugin
#### Known issues:

~~After running the update, the window will show error saying that you are in `detached HEAD` mode~~


c3n21:
Resolve #298

lua/packer.lua
```diff
@@ -80,13 +85,36 @@ local config_defaults = {
...
 local rocks = nil
+local plugins_snapshot = {}
+
+--- Loads and parse snapshot file in the global variable plugins_snapshot
+--@param filename path/to/snapshot
+packer.load_snapshot = function (filename)
+  local f_snapshot, err = io.open(filename, 'r+')
+
+  if err ~= nil then
+    print(err)
```
akinsho, line 96: probs a lingering debug print

lua/packer/install.lua
```diff
@@ -32,6 +32,9 @@ local function install_plugin(plugin, display_win, results)
 end
 
 local function do_install(_, plugins, missing_plugins, results)
+    for index, value in ipairs(plugins) do
+        print(fmt("index: %s, value: %s", index, value))
```
akinsho, line 36: Is this intentional or left over from the initial implementation

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,61 @@
+local a = require('packer.async')
+local async = a.sync
+local await = a.wait
+local fmt = string.format
+local log = require('packer.log')
+local plugin_utils = require('packer.plugin_utils')
+
+local function cfg(_config) config = _config end
+--[[
+property: 1
```
akinsho, line 10: What does this apply to? Is this an emmy lua annotation for something?

lua/packer.lua
```diff
@@ -519,6 +550,24 @@ packer.loader_complete = function(lead, _, _)
   return completion_list
 end
 
+-- Snapshot installed plugins
+-- Given a list of plugins it will only snapshot said plugins,
+-- otherwise it will snapshot all plugins (except for disabled)
+packer.snapshot = function (filename)
```
akinsho, line 556: I might be missing something but this function describes taking a list of plugins otherwise will snap shot all plugins but where is the list it takes?

lua/packer.lua
```diff
@@ -519,6 +550,24 @@ packer.loader_complete = function(lead, _, _)
...
+  async(function()
+    --local start_time = vim.fn.reltime()
+    log.info(string.format("Taking snapshots of currently installed plugins to %s...", filename))
+    filename = util.join_paths(config.snapshot_path, filename)
+    await(snapshot(filename, plugins))
+    await(a.main)
+    --local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
```
akinsho, line 563: Is this left over from some profiling you were doing?

CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
+# CHANGELOG
```
akinsho, line 1: This will be one for @wbthomason to decide but packer hasn't previously maintained a changelog so not sure if we need/should be adding this here since it isn't particularly discoverable. I think this should probably just go in the main docs 🤷🏾

akinsho (commented):
@c3n21 this looks great thanks for taking the time to implement this. I haven't tried it locally or anything just left some comments regarding some comments and a question or two.

Could you also maybe add some tests for this new functionality as well in `tests` dir under `snapshot_spec.lua` or something to validate that they are being created and read correctly. There are a few other tests there which should help guide you. Also there are docs on `plenary`'s repo

I should probably say that given this is whole new thing ultimately @wbthomason will have to review when he's free.

c3n21:
> @c3n21 this looks great thanks for taking the time to implement this. I haven't tried it locally or anything just left some comments regarding some comments and a question or two.
> 
> Could you also maybe add some tests for this new functionality as well in `tests` dir under `snapshot_spec.lua` or something to validate that they are being created and read correctly. There are a few other tests there which should help guide you. Also there are docs on `plenary`'s repo
> 
> I should probably say that given this is whole new thing ultimately @wbthomason will have to review when he's free.

Thank you for taking your time and review my PR, really appreciated.

As soon as I have some free time I will clean up the mess I left in the code and create the tests.

CHANGELOG.md
wbthomason: I actually favor adding a changelog and linking to it from the main docs; the README is too long as it is.

wbthomason: I would prefer to have changes in the changelog be date and commit-stamped, though.

lua/packer.lua
```diff
@@ -23,7 +24,11 @@ _G._packer = _G._packer or {}
 
 -- Config
 local packer = {}
+
+---@class Config
 local config_defaults = {
+  snapshot = nil,
+  snapshot_path = util.join_paths(vim.fn.stdpath('config'), 'plugin'),
```
wbthomason, line 31: I think this should probably default to somewhere in `stdpath('cache')/packer`, unless you intend for people to check their snapshots into VCS or something similar?

CHANGELOG.md
wbthomason: To clarify: we also ought to document this feature in the main docs, but I like having a changelog too. Also, the docs should describe what a snapshot does/does not save.

lua/packer.lua
```diff
@@ -80,13 +85,36 @@ local config_defaults = {
...
+--@param filename path/to/snapshot
+packer.load_snapshot = function (filename)
+  local f_snapshot, err = io.open(filename, 'r+')
+
+  if err ~= nil then
+    print(err)
+    log.info(err)
+  else
+    for line in f_snapshot:lines() do
+      local short_name, commit = unpack(vim.split(line, ' '))
```
wbthomason, line 100: Do I correctly understand that the snapshots save plugin names and commits, but not e.g. configuration other than the commit?

lua/packer.lua
```diff
@@ -181,6 +210,8 @@ manage = function(plugin)
   plugin.name = path
   plugin.path = path
 
+  plugin.commit = plugins_snapshot[plugin.short_name]
```
wbthomason, line 213: I might be missing something, but: what if there's no commit for a plugin in a snapshot, either if the user has no snapshot or if there is a new plugin in the specification not included in an old snapshot?

lua/packer.lua
```diff
@@ -519,6 +550,24 @@ packer.loader_complete = function(lead, _, _)
...
+-- Snapshot installed plugins
+-- Given a list of plugins it will only snapshot said plugins,
+-- otherwise it will snapshot all plugins (except for disabled)
+packer.snapshot = function (filename)
+  async(function()
+    --local start_time = vim.fn.reltime()
+    log.info(string.format("Taking snapshots of currently installed plugins to %s...", filename))
+    filename = util.join_paths(config.snapshot_path, filename)
+    await(snapshot(filename, plugins))
+    await(a.main)
```
wbthomason, line 562: We probably don't need to `await` main here with the below profiling line gone.

lua/packer.lua
```diff
@@ -519,6 +550,24 @@ packer.loader_complete = function(lead, _, _)
...
+    --local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
+    --print(string.format("delta: %s", delta))
+    log.info("Snapshot complete")
+    packer.on_complete()
```
wbthomason, line 566: Need to document that `on_complete` triggers for snapshots too (not just installs, etc.)

lua/packer/plugin_types/git.lua
```diff
@@ -114,6 +114,7 @@ git.setup = function(plugin)
 
   local needs_checkout = plugin.tag ~= nil or plugin.commit ~= nil or plugin.branch ~= nil
 
+
```
wbthomason, line 117: Extraneous blank line; please remove.

lua/packer/plugin_types/git.lua
```diff
@@ -331,6 +332,26 @@ git.setup = function(plugin)
     end)()
     return r
   end
+
+  plugin.get_rev = function ()
```
wbthomason, line 336: Nitpick, but it feels like the main logic here could be separated out into a function not associated with a specific plugin, then the `plugin.get_rev` becomes a simpler function calling the refactored `get_rev` function with the plugin's name and install location. This might be preferable because getting the revision of a plugin is something we do in several places, so reusing the same function simplifies the overall logic.

lua/packer/result.lua
```diff
@@ -1,4 +1,5 @@
--- A simple Result<V, E> type to simplify control flow with installers and updaters
+--- A simple Result<V, E> type to simplify control flow with installers and updaters
```
wbthomason, line 1: Is this a change to make a documentation comment? Not sure what this change is otherwise.

lua/packer.lua
wbthomason: I'm also a bit curious how this will interact with currently specified plugins/config values. It seems as though they can override the snapshotted values, but is that what we want? We should discuss the desired semantics of snapshots. I'm not familiar with this feature in e.g. `plug`, so this may already be a solved problem.

wbthomason (changes requested):
Thanks, as @akinsho said this looks like a great start on the feature. I agree with @akinsho's comments and left some questions of my own on the changes, mostly around documentation and deciding how snapshots will interact with "fresh" plugin specification values.

lua/packer/result.lua
akinsho: @wbthomason I presume so since the type won't appear in lsp hover docs etc if it isn't a doc comment using `---` I've been meaning to suggest using more of this pattern + emmy annotations since once sumneko is setup right it's really helpful to get to know parts of a code base or just remember how things work

akinsho: Actually correction I just checked and sumneko will pick it up, I think it's function annotation `--@param` won't work as an emmy lua annotation only `---@param` so assumed the same was true here but isn't

lua/packer.lua
c3n21: > I might be missing something, but: what if there's no commit for a plugin in a snapshot, either if the user has no snapshot or if there is a new plugin in the specification not included in an old snapshot?

If a user doesn't have a plugin's snapshot it will be `plugin.commit = nil` (which is already the default behaviour when you don't specify the commit in `use`), so it wouldn't make any difference imo

c3n21: > I'm also a bit curious how this will interact with currently specified plugins/config values. It seems as though they can override the snapshotted values, but is that what we want? We should discuss the desired semantics of snapshots. I'm not familiar with this feature in e.g. `plug`, so this may already be a solved problem.

Do you mean if the `plugin.commit` has already a value, should it be rolled back to the snapshotted state?
Didn't really think about that so I guess you have a point.
Maybe we should just ignore the plugins which have commit `specified`? It would make more sense to me since if I want to specify a specific commit version of a package I would like to use that in any case.

lua/packer/result.lua
c3n21: My first thought was to make it more Emmy friendly since I had to figure out on my own how this worked, but since I'm still a newbie in Lua (and contributing in general) I thought it was better leave it as it is and maybe document better in another PR

lua/packer.lua
c3n21: > I'm also a bit curious how this will interact with currently specified plugins/config values. It seems as though they can override the snapshotted values, but is that what we want? We should discuss the desired semantics of snapshots. I'm not familiar with this feature in e.g. `plug`, so this may already be a solved problem.

If I remember correctly when you take a snapshot using `vim-plug` it actually generates a vimscript file which you can `:source` and you have rolled back to the previous state.

In the beginning I wanted to make something similar, but I didn't like that you had to specify the snapshot's location and that you had to source it manually.

But it could be interesting to have a PackerRollback <snapshot> to load a snapshot only in the current session.

lua/packer.lua
c3n21: > Do I correctly understand that the snapshots save plugin names and commits, but not e.g. configuration other than the commit?

Yes, since configs aren't really part of the plugin state I thought it was better to leave it out, or do you think it might be better to back all up in the snapshot (i.e. `config`, `requires` ecc...)

wbthomason: For a first pass on the feature, I think just snapshotting commits is fine - we just need to be clear in the docs that we don't save other config information.

lua/packer.lua
wbthomason: Although it's not really intended for this purpose, `packer` relies on `vim.inspect` to serialize tables for e.g. generating the compiled file. That said, it seems like all we'd really need is to output the plugin names and commits and write a `rollback` function to checkout those commits in the relevant directories. I think those semantics might be cleaner than figuring out the potential interactions between a snapshot and a "fresh" config, tbh.

lua/packer.lua
c3n21: My bad, will fix

lua/packer.lua
c3n21: Yes that was my initial idea, however it's configurable so maybe by default i'll just put use `cache`

lua/packer.lua
c3n21: > We probably don't need to `await` main here with the below profiling line gone.

This is the error I get without `await a.main`
```[packer.nvim] [ERROR 15:30:56] async.lua:20: Error in coroutine: ...Downloads/github/packer.nvim.git/snapshot/lua/packer.lua:275: E5560: 
nvim_exec must not be called in a lua loop callback 
```

Line count might be different from the version in the repo since I made some changes

lua/packer.lua
wbthomason: Although it is configurable, I think it's better to default to `cache` as we cannot really assume that users will read the docs and customize the destination before using the feature.

c3n21:
Just implemented PackerRollback.
Next I would like to:
- add the completion for PackerRollback
- add the tests and it should be good to go

c3n21:
@wbthomason @akinsho I've been working on the tests for the snapshot feature, but I can't get it to work with `plugin_utils_spec.lua` and `packer_snapshot_spec.lua` (which is the one I've created)

Could you give me some advice or point out where am I doing wrong?

akinsho:
@c3n21 are you saying the tests don't run? or that you can't get them to pass? anyway re. running them you need to use `PlenaryBustedDirectory tests/ {minimal_init = 'tests/minimal.vim'}` I advise this over the file only test since that will pick up your `init.vim` and could be causing unpredictable results.

Regarding the tests themselves from what I can see it appears you might be nesting describe blocks inside the tests themselves which won't work

akinsho:
Oh and a side note about the tests can you call them just `snapshot_spec.lua` I don't think the prefix is needed and I mean to try and change any of the existing test files that are prefixed that way since they should just use the filename + `_spec.lua` as the test name

tests/packer_snapshot_spec.lua
```diff
@@ -0,0 +1,34 @@
...
+    local spec = {"tjdevries/colorbuddy.vim"}
+    local short_name = 'colorbuddy'
+    local snapshot_name = 'test'
+    local test_path = path.joinpath(config.snapshot_path, snapshot_name)
+    local expected = {}
+    expected.colorbuddy = '87c80e3'
+    packer.startup(spec)
+    packer.install()
+    packer.snapshot(snapshot_name)
+    describe("snapshot exists", function ()
```
akinsho, line 22: Describe blocks shouldn't be nested within `it` blocks, the describe block should be a top level grouping of what functionality you are testing. Within which each test case should be in an it block.

See [here](http://olivinelabs.com/busted/#overview) which is essentially the library that plenary uses/copied

c3n21:
@akinsho 

I've made removed the describe block inside the it block as you said and at least it runs now.

However, if I add some async code the test gives me this result

`
We had an unexpected error:     { "...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:314: Blocking on future failed ...acker/st
art/plenary.nvim/lua/plenary/async_lib/async.lua:20: expected a single return value\n\nstack traceback:\n\t...packer/start/plenary.nvim/l
ua/plenary/async_lib/util.lua:319: in function <...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:303>\n" }   {
  errs = {},
  fail = {},     
  fatal = { "...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:314: Blocking on future failed ...acker/start/plenary.nvim/lua/p
lenary/async_lib/async.lua:20: expected a single return value\n\nstack traceback:\n\t...packer/start/plenary.nvim/lua/plenary/async_lib/u
til.lua:319: in function <...packer/start/plenary.nvim/lua/plenary/async_lib/util.lua:303>\n" },
  pass = {}                                                                                                                              
}
`

but if I remove all the async code (remove await and `a.` from `describe` and `it` it runs but it makes an empty snapshot.
That's probably because it doesn't install the test plugin, but I can't figure out how to make it work.

akinsho:
@c3n21 the async testing has kind of been borrowed from plenary but packer has it's own async implementation so I think it's a little rough and obscure used together. When I've seen this error in the past it's usually meant that there is an error occurring in your test somewhere and this is how it is bubbling up. It could be the usage/syntax also I'll take a look and see if there are any incorrect usages

akinsho:
@c3n21 I can't see any clear errors. One thing to keep in mind is that it's probably a good idea to test as close to your changes as possible for the time being i.e. just call your functions and assert they do the correct thing rather than trying to setup all of packer. A lot of those code paths are untested so you could easily get blocked by a completely separate issue that is unrelated to your change.

c3n21:
> @c3n21 I can't see any clear errors. One thing to keep in mind is that it's probably a good idea to test as close to your changes as possible for the time being i.e. just call your functions and assert they do the correct thing rather than trying to setup all of packer. A lot of those code paths are untested so you could easily get blocked by a completely separate issue that is unrelated to your change.

The problem is that I need to be sure that at least one plugin is installed using `packer` so that I can snapshot it, check the snapshot format etc, so it's quite mandatory to call `packer.install()`

Without that it's pointless...

Atm I'm only checking if the snapshot have been created or not

akinsho (commented):
It looks like `snapshot` is only really dependent on `list_missing_plugins`, so rather than trying to run install which I think will have quite a lot of side effects you could try to just mock that function. I think plenary has a few mock utilities but I don't know if those will work with a required module so another thing you could do is in the setup of your test you could replace `require` with `fake_require` and check if the value is that module and return a fake module

```lua
function test_require(module_name)
  if module_name == "plugin_utils" then
    return {list_missing_plugins = function() return {value_I_want} end}
  end
  return require(module_name)
end

-- Inside describe
_G.require = test_require

**Comment by c3n21** [comment]
Sorry for the absence but I need to take some exams, so I'll keep working on this once I'll be done (by the end of the month)

Just letting you know that I'm not dead and didn't give up about this feature :)

**Comment by c3n21** [comment]
Update:

finally I've managed to pull it off.

I have updated the first message of this thread so check it out for the "basic usage".

I've also updated the documentation about the snapshot feature and added all the tests I could think of.

**Comment by c3n21** [comment]
@wbthomason is this feature still needed? Because it still needs some improvement but if it's not needed I'm not going to improve this

**Comment by akinsho** [comment]
@c3n21 @.wbthomason is away atm and I haven't really been following things too closely. I don't think the need for this has changed though since it's driven entirely by all the people who would want to use it which I don't think has changed and no alternative has been implemented

**Comment by c3n21** [comment]
> @c3n21 @.wbthomason is away atm and I haven't really been following things too closely. I don't think the need for this has changed though since it's driven entirely by all the people who would want to use it which I don't think has changed and no alternative has been implemented

Ok, I'll mention you for the review when I'll be done

**Comment by c3n21** [comment]
@akinsho it should be good now. I've added the docs and updated the README.md.

**Comment by wbthomason** [comment]
Hi @c3n21! Thanks for your continued work on this, and apologies for my lack of response; as @akinsho mentioned, I was gone.

For reviewing/testing: do you have any suggested operations which would be good to validate that everything is working?

**Comment by c3n21** [comment]
> Hi @c3n21! Thanks for your continued work on this, and apologies for my lack of response; as @akinsho mentioned, I was gone.
> 
> For reviewing/testing: do you have any suggested operations which would be good to validate that everything is working?

Hi and welcome back!
I've already provided some tests for `git.get_rev` (which fails and I don't know why), `git.revert` (not `revert_last`, mine does a hard reset of the git repo to `git.commit`),`snapshot()`, packer.delete().

For testing you could:

1. Make a snapshot with `PackerSnapshot <snapshot name>`;
2. `PackerUpdate` and ensure that `git rev-parse --short HEAD` of the plugin is different with the one reported inside `~/.cache/nvim/packer/<snapshot name>`; 
3.  `PackerRollback <snapshot name>` (you can use TAB to autocomplete the snapshot name) and rollback to <snapshot name> and check with `git rev-parse --short HEAD` that the package have been resetted properly.
4. `PackerDelete <snapshot name>` (TAB if you want autocompletion) if you don't need that snapshot anymore

You can also specify <snapshot name> in the config, e.g. `config = {snapshot = <snapshot name>}` when you run `packer.startup()`

**Comment by c3n21** [comment]
Hi @wbthomason ! Did you manage to review this feature?

**Comment by c3n21** [comment]
I've updated my PR and now it seems to pass all the tests.
I've also cleaned up my fork's history since it was a bit cluttered.
Could you take a look when you have some time? @wbthomason

**Comment by smhc** [comment]
I think this is generally a good change, albeit a few hacks that could be improved.

My main use case is to generate a snapshot file such that on installation those commit hashes are used. Thus I can control clean installations and prevent them pulling down random changes that haven't been tested and may break compatibility etc. I'm using this fork mainly for generating this snapshot file - I do no not need the other functionality (rollback, load etc). I note some other neovim configs are doing something similar.

A few notes:

- The 'PluginSnapshot' currently saves to a cache directory by default. If this doesn't exist it silently fails. I suspect a more appropriate path is 'vim.fn.stdpath(config)..'/lua' - which is more likely to exist, and the snapshot is something you would check in to your configuration.
- I've currently updated all my plugin specs to do:
```lua
local snap=require('packer_snap')
use {
    'chrisbra/csv.vim',
    commit = snap['csv.vim'].commit,
}
```
- It would be helpful if the the 'use' command could use this commit hashes by default. My thinking is something like:

```lua
packer.init {
 commits = require'plugin_snap'
}
use {
    'chrisbra/csv.vim',
    -- commit  this is implicitly looked up from commits table in packer.config
}
```

smhc:
Should also mention: it might be useful to extract this snapshot file generation out to a separate plugin. The rate of pr merges is quite low in this plugin.. separating the functionality out might increase the chance of it being delivered, and lets users choose whether they want it rather than relying on author of packer.

c3n21:
> Should also mention: it might be useful to extract this snapshot file generation out to a separate plugin. The rate of pr merges is quite low in this plugin.. separating the functionality out might increase the chance of it being delivered, and lets users choose whether they want it rather than relying on author of packer.

It could be a good idea, I could just implement the "reset plugin to commit" function in my own plugin and the upstream code wouldn't be affected. Will give it a try when I will have some time, since the author said that packer is undergoing a huge refactor.

> I think this is generally a good change, albeit a few hacks that could be improved.
> 
> My main use case is to generate a snapshot file such that on installation those commit hashes are used. Thus I can control clean installations and prevent them pulling down random changes that haven't been tested and may break compatibility etc. I'm using this fork mainly for generating this snapshot file - I do no not need the other functionality (rollback, load etc). I note some other neovim configs are doing something similar.
> 
> A few notes:
> 
>     * The 'PluginSnapshot' currently saves to a cache directory by default. If this doesn't exist it silently fails. I suspect a more appropriate path is 'vim.fn.stdpath(config)..'/lua' - which is more likely to exist, and the snapshot is something you would check in to your configuration.
Will fix, even though you can change the default path by setting **config.snapshot_path** to something else.
However using **~/.cache/nvim/packer/** was suggested as default so that the snapshots won't clutter your nvim config directory. 
> 
>     * I've currently updated all my plugin specs to do:
> 
> 
> ```lua
> local snap=require('packer_snap')
> use {
>     'chrisbra/csv.vim',
>     commit = snap['csv.vim'].commit,
> }
> ```
> 
>     * It would be helpful if the the 'use' command could use this commit hashes by default. My thinking is something like:
> 
> 
> ```lua
> packer.init {
>  commits = require'plugin_snap'
> }
> use {
>     'chrisbra/csv.vim',
>     -- commit  this is implicitly looked up from commits table in packer.config
> }
> ```

You could just specify **config.snapshot = "snapshot name"** and at startup it will automatically rollback to the commit hashes specified in "snapshot name". But if you want to update you will need to set **config.snapshot = nil** so that it won't get in the way.

doc/packer.txt
```diff
@@ -44,6 +44,7 @@ FEATURES                                         *packer-intro-features*
 - Uses jobs for async installation
 - Support for `git` tags, branches, revisions, submodules
 - Support for local plugins
+- Support snapshots for plugins
```
wbthomason, line 47: Nitpick: would prefer "Support for saving/restoring snapshots of plugin versions" (or similar, just something a bit more descriptive).

doc/packer.txt
```diff
@@ -126,6 +127,14 @@ Perform `PackerUpdate` and then `PackerCompile`.
 `PackerLoad`                                    *packer-commands-load*
 Loads opt plugin immediately
 
+`PackerSnapshot`                                    *packer-commands-snapshot*
```
wbthomason, line 130: This should probably be fixed for the other commands too (not as part of this PR), but we might want to specify the arguments accepted for the snapshot related commands (i.e. the snapshot file).

doc/packer.txt
```diff
@@ -531,6 +540,15 @@ It can be invoked with no arguments or with a list of plugin names to update.
 These plugin names must already be managed by `packer` via a call to
 |packer.use()|.
 
+snapshot()		                                *packer.snapshot()*
+`snapshot` takes the rev of all the installed plugins and serializes them into a lua table which will be saved at `config.snapshot_path` or an absolute path provided by the users. The snapshot can be loaded via `dofile`
```
wbthomason, line 544: Nitpick: please capitalize "Lua".

wbthomason: We might also want to clarify that `config.snapshot_path` is a directory holding multiple snapshot files, and the actual path to which the snapshot is saved has some additional path component.

lua/packer.lua
```diff
@@ -8,6 +8,7 @@ local stdpath = vim.fn.stdpath
 -- Config
 local packer = {}
 local config_defaults = {
+  snapshot_path = join_paths(stdpath 'cache', 'packer'),
```
wbthomason, line 11: For consistency, this should probably default to `join_paths(stdpath 'cache', 'packer.nvim')`

lua/packer.lua
```diff
@@ -123,6 +126,9 @@ packer.init = function(user_config)
 end
 
 packer.make_commands = function()
+  vim.cmd [[command! -nargs=+ PackerSnapshot  lua require('packer').snapshot(<q-args>)]]
```
wbthomason, line 129: I could be wrong, but I think we want to use `<f-args>` instead of `<q-args>` here (and for the other snapshot commands)?

lua/packer.lua
```diff
@@ -784,6 +790,22 @@ packer.loader_complete = function(lead, _, _)
   return completion_list
 end
 
+--- Completion for listing snapshots in `snapshot_path`
+--- Intended to provide completion for PackerRollback command
+--- TODO: using vim.fn.readdir to get entries the snapshot directory
+packer.snapshot_complete = function(lead, _, _)
+  local completion_list = {}
+  local res = io.popen('ls ' .. config.snapshot_path, 'r')
```
wbthomason, line 798: Minor nitpick: we'd generally prefer to use the `vim.loop.fs_*` functions rather than running `ls` directly. This helps `packer` be OS-independent, and may help with performance.

lua/packer.lua
```diff
@@ -784,6 +790,22 @@ packer.loader_complete = function(lead, _, _)
...
+  for entry in res:lines() do
+    if vim.startswith(entry, lead) then
+      table.insert(completion_list, entry)
```
wbthomason, line 801: Super minor nitpick: generally, we'd prefer (for consistency and small potential performance gains) to use the `table_name[#table_name + 1] = item` style of appending over `table.insert`.

lua/packer.lua
```diff
@@ -784,6 +790,22 @@ packer.loader_complete = function(lead, _, _)
...
+    end
+  end
+  res:close()
+  table.sort(completion_list)
```
wbthomason, line 805: Is the sort necessary? Since we're doing prefix matching anyway, I'd expect it doesn't really matter if we sort.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+---Snapshots installed plugins
+---@param snapshot_name string @ can be an absolute path or just a filename
+packer.snapshot = function(snapshot_name)
+  local async = require('packer.async').sync
+  local await = require('packer.async').wait
+  local log = require_and_configure 'log'
+
+  async(function()
+    local snapshot = require 'packer.snapshot'
+    local snapshot_path = snapshot_name
```
wbthomason, line 830: What if `snapshot_name` is nil? Can the user take an anonymous snapshot? We should check for validity either way.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+    local fmt = string.format
+    log.debug(fmt('Taking snapshots of currently installed plugins to %s...', snapshot_name))
+    if vim.fn.fnamemodify(snapshot_name, ":p") ~= snapshot_path then -- is not absolute path
+      snapshot_path = util.join_paths(config.snapshot_path, snapshot_path) -- save to default path
+
+    end
+    await(snapshot(snapshot_path, plugins))
+    local msg = fmt("Snapshot '%s' complete", snapshot_path)
+    log.info(msg)
+    print(msg)
```
wbthomason, line 841: We should probably remove this debug print.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+    packer.on_complete() --not sure if it should fire packer.on_complete()
```
wbthomason, line 842: We shouldn't fire `on_complete` - that's for when plugin update/install/sync operations finish.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+  end)()
+end
+
+---Deletes the snapshot provided
+---@param snapshot_name string @name of the snapshot or the absolute path
+---@return boolean @true if success, false if failed
+packer.delete = function (snapshot_name)
```
wbthomason, line 849: Nitpick: I'd prefer this to be named something like `snapshot_delete`. Or to live in the `snapshot` module.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+  local async = require('packer.async').sync
+  local log = require_and_configure 'log'
+
+    async(function ()
+      local snapshot_path = snapshot_name
+      if vim.fn.fnamemodify(snapshot_path, ":p") ~= snapshot_path then
```
wbthomason, line 855: We use this "absolute path or not" logic a few times - maybe it should be a utility function?

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+  local log = require_and_configure 'log'
+  async(function()
+    manage_all_plugins()
+    local snapshot_path = snapshot_name
+    if vim.fn.fnamemodify(snapshot_path, ":p") ~= snapshot_path then
+      snapshot_path = util.join_paths(config.snapshot_path, snapshot_path)
+    end
+
+    log.debug("Rolling back to " .. snapshot_path)
+    print("Rolling back to " .. snapshot_path)
```
wbthomason, line 879: We should remove this debug print.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+      end
+      log.debug("Deleting " .. snapshot_path)
+      return vim.fn.delete(snapshot_path) == 0
+    end)()
+end
+
+---Instantly rolls back plugins to a previous state specified by `snapshot_name`
+---If `snapshot_name` doesn't exist an error will be displayed
+---@param snapshot_name string @name of the snapshot or the absolute path to the snapshot
+packer.rollback = function(snapshot_name)
```
wbthomason, line 866: Should this logic be in the main module, or the `snapshot` module?

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+        async(function ()
+          await(plugin.revert())
+        end)
+      )
+    end
+
+    wait_all(unpack(jobs))
+    log.info('Rollback complete')
+    packer.on_complete()
+    print('Rollback complete')
```
wbthomason, line 900: We should remove this debug print.

lua/packer.lua
```diff
@@ -796,6 +818,89 @@ packer.plugin_complete = function(lead, _, _)
...
+      if vim.fn.fnamemodify(snapshot_path, ":p") ~= snapshot_path then
+        snapshot_path = util.join_paths(config.snapshot_path, snapshot_path)
+      end
+      log.debug("Deleting " .. snapshot_path)
+      return vim.fn.delete(snapshot_path) == 0
+    end)()
+end
+
+---Instantly rolls back plugins to a previous state specified by `snapshot_name`
+---If `snapshot_name` doesn't exist an error will be displayed
```
wbthomason, line 864: Where do we display the error? I'm missing that in the below code.

lua/packer/plugin_types/git.lua
```diff
@@ -128,6 +140,31 @@ local handle_checkouts = function(plugin, dest, disp)
...
+        { cwd = plugin.install_path, options = { env = git.job_env }, capture_output = true }
+      ))
+    :map_ok(function (ok)
+        local _, r = next(ok.output.data.stdout)
+        return r
+      end)
+    :map_err(function(err)
+        local _, msg = fmt("%s: %s",plugin_name,next(err.output.data.stderr))
+        error(msg)
+        return ""
```
wbthomason, line 161: This will just set `rev.err = ''` - we should probably set it to the error message string. Also, we'll never reach this code anyway because of the `error` call above it.

I'd suggest removing the `error` call and returning `rev` from `get_rev` (instead of `rev.ok`). Then callers of `get_rev` can decide how to handle any potential errors.

lua/packer/plugin_types/git.lua
```diff
@@ -454,6 +491,30 @@ git.setup = function(plugin)
...
+    return async(function ()
+        log.info(fmt("'%s' has no commit to revert to", plugin.name))
+    end)
+  end
+
+  ---Returns HEAD's short hash
+  ---@return string
+  plugin.get_rev = function()
+    return async(function()
+      local res = await(get_rev(plugin))
```
wbthomason, line 514: Nitpick: this wrapper is unnecessary - it's async and needs to be awaited anyway, so we can just have its callers await the `get_rev` function (or move that logic here).

lua/packer/plugin_types/git.lua
```diff
@@ -454,6 +491,30 @@ git.setup = function(plugin)
     end)()
     return r
   end
+
+  ---Reset the plugin to `plugin.commit`
+  plugin.revert = function ()
```
wbthomason, line 496: Maybe this should be called something more like `revert_to` or similar, to avoid confusion with the `revert_last` method?

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,82 @@
...
+    for key, _ in pairs(start) do installed[key] = key end
+
+    for _, plugin in pairs(plugins) do
+      if installed[plugin.install_path] ~= nil then -- this plugin is installed
+        log.debug(fmt("Snapshotting '%s'", plugin.short_name))
+        if plugin.type == plugin_utils.git_plugin_type then
+          local rev = await(plugin.get_rev())
+
+          if rev == "" or rev == nil then
+            local msg = fmt('Snapshotting %s failed', plugin.short_name)
```
wbthomason, line 42: Nitpick: this message should be more descriptive. Maybe more like "Snapshotting to {filename} failed: could not get current revision for {plugin.short_name}"

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,82 @@
...
+--    end))
+-- Doesn't work using vim.fn.writefile
+
+--    vim.schedule_wrap(function ()
+--      if vim.fn.writefile(snapshot, filename) ~= 0 then
+--        log.warn("Couldn't save snapshot")
+--      else
+--        log.info("Snapshot saved succesfully")
+--      end
+--    end)
```
wbthomason, line 63: We should remove these comments before the final merge.

wbthomason (changes requested):
@c3n21: Thanks, and sorry (again!) that it's taken me so long to look at this. I think this is a solid start, and I've left some comments addressing changes that should be made - mostly stylistic tweaks and nitpicks.

c3n21:
Sorry if I didn't reply earlier but I had nearly no time due to university.

Anyway, I fixed all the things you pointed out @wbthomason and moved most of the snapshot logic under `snapshot` module, even the completion functions.

I've also added the possibility to selectively snapshot and rollback plugins, made the code more OS-independent using vim.loop as you said and more robust with assertions and checks.

danymat:
Hello, I just found this PR and I think it could be a great addition for stable neovim configs such as Nvchad, lunarvim, etc. What is the current eta ?

c3n21:
> Hello, I just found this PR and I think it could be a great addition for stable neovim configs such as Nvchad, lunarvim, etc. What is the current eta ?

My fork should work but it's not up to date with upstream

lua/packer.lua
```diff
@@ -38,6 +40,7 @@ local config_defaults = {
       get_bodies = 'log --color=never --pretty=format:"===COMMIT_START===%h%n%s===BODY_START===%b" --no-show-signature HEAD@{1}...HEAD',
       submodules = 'submodule update --init --recursive --progress',
       revert = 'reset --hard HEAD@{1}',
+      revert_to = 'reset --hard %s --',
```
shadmansaleh, line 43: Question: will this work on new installation ? Say I newly install the plugin then I want to restore a snapshot that I have from older installation. It'll be useful if you're setting up config in a new device or temporarily on a server . I'm asking because packer makes shallow clones that means the commit reference might not exist in newly created clone .

lua/packer.lua
```diff
@@ -123,6 +127,9 @@ packer.init = function(user_config)
 end
 
 packer.make_commands = function()
+  vim.cmd [[command! -nargs=+ -complete=customlist,v:lua.require'packer.snapshot'.completion.create PackerSnapshot  lua require('packer').snapshot(<f-args>)]]
+  vim.cmd [[command! -nargs=+ -complete=customlist,v:lua.require'packer.snapshot'.completion.rollback PackerRollback  lua require('packer').rollback(<f-args>)]]
```
shadmansaleh, line 131: These should probably be called `PackerSnapshotRollback` and `PackerSnapshotDelete` especially the delete on It can be easily mistaken for something else like plugin uninstallation or deletion on packer_compiled file.

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,192 @@
...
+            snapshot_plugins[plugin.short_name] = {commit = rev.ok}
+          end
+        end
+      end
+    end
+
+    ---@type string
+    local snapshot_content = "return " .. vim.inspect(snapshot_plugins)
+
+    local fd = loop.fs_open(snapshot_path, "w", tonumber("600", 8))
```
shadmansaleh, line 117: It should be checked if snapshot dir exists if not it should be created . Currently snapshot create fails if the dir doesn't exist . This is inconvenient and the fix is ambiguous.

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,192 @@
...
+
+    if fd == nil then
+      local warn = fmt("Error on creation of snapshot '%s'",snapshot_path)
+      log.warn(warn)
+    else
+      local res = loop.fs_write(fd, snapshot_content)
+      loop.fs_close(fd)
+      if res ~= #snapshot_content then
```
shadmansaleh, line 125: Packer should also notify user when snapshot succeeds.

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,192 @@
...
+              local res = await(plugin.revert_to(commit))
+              if res.err then
+                log.error(res.err)
+              end
+            end)
+          end
+        end
+      end
+
+      wait_all(unpack(jobs))
```
shadmansaleh, line 164: Same as above user should be notified when rollback succeeds/

lua/packer/snapshot.lua
```diff
@@ -0,0 +1,192 @@
...
+            log.warn(warn)
+          else
+            snapshot_plugins[plugin.short_name] = {commit = rev.ok}
+          end
+        end
+      end
+    end
+
+    ---@type string
+    local snapshot_content = "return " .. vim.inspect(snapshot_plugins)
```
shadmansaleh, line 115: Since it's just data . Maybe we can consider storing it serialized as json/msgpack instead of putting it as a lua file and evaluating it . What do you think ?

shadmansaleh:
Over all this looks good and seems functional . Good job @c3n21.

lua/packer/snapshot.lua
c3n21: I still wonder what should I use to notify the user.
Should I use `log.info` or `print` ?

lua/packer/snapshot.lua
c3n21: My first thought was the same, the problem is that I am not aware of any kind of way to deserialize JSON in Lua without implementing a parser on your own.

lua/packer/snapshot.lua
shadmansaleh: > Should I use log.info or print

I think you should use `vim.notify()` . That's what `PackerCompile` uses .

lua/packer/snapshot.lua
shadmansaleh: There's `vim.mpack` (msgpack implementation) & `vim.json` (cjson implementation) . `vim.json.encode` & `vim.json.decode` is better then `json_encode` and `json_decode` but I think this was added on nvim-0.5.1 / 0.6.0 I'm not sure which . @wbthomason what your opinion on using 0.5+ features ?

c3n21: > There's `vim.mpack` (msgpack implementation) & `vim.json` (cjson implementation) . `vim.json.encode` & `vim.json.decode` is better then `json_encode` and `json_decode` but I think this was added on nvim-0.5.1 / 0.6.0 I'm not sure which . @wbthomason what your opinion on using 0.5+ features ?

I could just wrap it in util.
If the user is running 0.5 then use json_encode, else if 0.6+ use vim.json.encode etc.

shadmansaleh: > I could just wrap it in utils.
If the user is running 0.5 then use json_encode, else if 0.6+ use vim.json.encode etc.

Now thinking about it, for such small files difference in using vim.json / json_encode|decode won't be significant . Using different functions based on nvim version is unnecessary complexity . We'll be fine with just using json_encode|decode.

smhc:
> We can consider storing it serialized as json/msgpack instead of putting it as a lua file and evaluating it . What do you think ?

Is there an advantage to using json over lua? Using lua seems to offer the most flexibility and best performance.
e.g I am currently generating a packer snapshot file using this code and then ensuring I use specific commit on installation using the following code:

```lua
  local snapstore = require('packer_snap')
  local snap = setmetatable({}, {
    __index = function(_, plgkey)
      if (not snapstore[plgkey]) then
        print('Warning: no snapshot: ' .. plgkey);
      end
      if (key ~= 'update' and snapstore[plgkey]) then
        return snapstore[plgkey]
      end
      return {
        commit = nil
      }
    end,
  })

  use {
    'nvim-lualine/lualine.nvim',
    commit = snap['lualine.nvim'].commit,
  }
```

Note the above also ensures packer uses the commit on installation. 
IMHO I think it is very important that it supports using specific commits during install - this is likely to be a common use case. e.g I use it for setting up docker containers to a known plugin state.

shadmansaleh:
> Is there an advantage to using json over lua? Using lua seems to offer the most flexibility and best performance.

There's no significant performance difference in in this case (deserialization json might actually be faster then evaluating lua). I don't think we actually want flexibility in  snapshot file . These aren't meant to be created by user instead these are meant to be internal data for packer to restore old state . Is there any reason why user will be writing a script as snapshot ? Also json can be more easily parsed by external tools .

> I am currently generating a packer snapshot file using this code and then ensuring I use specific commit on installation using the following code

You'll be able to do the same . You'll just be deserializing instead of evaluating after that you'll have the same thing.

> IMHO I think it is very important that it supports using specific commits during install - this is likely to be a common use case

We are dicussing how snapshots can work on installation or right after installation https://github.com/wbthomason/packer.nvim/pull/370#discussion_r790172159

shadmansaleh (approved):

lua/packer.lua
c3n21: What is we just run `git fetch --depth 999999 --progress` before rolling back?

c3n21:
The PR is up to date and fixed all the stuff pointed out by @shadmansaleh , even the issue about the snapshot being older than the commit (on rollback it will run `git fetch --depth 999999 --progress`).

wbthomason:
Taking a look now. Thanks for the work @c3n21, guidance @shadmansaleh, and patience on everyone's parts.

c3n21:
Right now I'm having some issues with the async code as mentioned [here](https://github.com/wbthomason/packer.nvim/discussions/817)

If anyone could help me figure out what's going out it would be great.

wbthomason:
@c3n21 I fixed the merge conflict and am re-running the tests; I'll look at the trace and see if I can figure out what's going on.

c3n21:
> @c3n21 Tests seem good to go now. Anything remaining before merge that you know of?

I've made a minor refactor to make the code inside snapshot.create() shorter and a bit more modular.
If you don't like it just reset it and merge it as it is if it's all good to go.

Thanks for your time!

c3n21:
@wbthomason I've polished a bit the code with the advices you gave me above.
I've also added an updated version of the Dockerfile and Makefile I've used during my tests which may be useful in the future.
In the Dockerfile I've create a test user and copied packer.nvim root into the container.

I think this might be useful because rollback hard resets the repo, and using it on the host machine could reset all the work not pushed

wbthomason:
Thanks again, @c3n21! I think we're good to merge.

smhc:
Thanks @c3n21 - good to see this merged. Thanks for all your work on it.

I've been using it since the .lua snapshot generation and have now switched to the more recent incarnation which has .json generation. I do notice that a side effect of async and vim.encode_json is that the .json file generated is not ordered or stable. This is not ideal for checking into source control and easily seeing what changed - I would expect this snapshot file to be source controlled by many people.

I couldn't find a simple solution so I simply use 'jq --sort-keys' on the file afterwards - but if anyone knows a simple fix for packer to do it automatically I can raise a PR.

shadmansaleh:
The output isn't table because lua tables aren't stable . We can use list and sort the list before writing it out .

arathunku:
I've solved keys sorting by additionally processing the final file, given that it's a json dump. 
 
```bash
echo $(jq --sort-keys . packages) > packages
prettier --parser json --write packages
```
Easy on eyes diffs :)

smhc:
Yes I use something like:

```lua
    packer.snapshot_format = function(fn)
      local fullpath=vim.fn.stdpath('config')..'/'..fn
      local tmp=vim.fn.stdpath('config')..'/temp_'..fn
      os.execute('jq --sort-keys . '..fullpath..'> '..tmp)
      os.rename(tmp, fullpath)
    end
```
The jq format seems as good as it can get, I don't think there's any need to run 'prettier'. I'd prefer to chain it with 'packer.snapshot' however I couldn't get the 'await', 'and_then' stuff working correctly, and without that there is race conditions causing empty files etc. So I just run it manually afterwards.

PR #371, opened 2021-05-25
Fix floating window example in README

SmiteshP:


akinsho:
Thanks @SmiteshP this is really helpful, should help people not get tripped up on this step 👍🏾

wbthomason (approved):
Derp, yep, thanks. I forgot to wrap the `startup` function in a table when we added the config change example. It almost makes me think it would be better to make startup just take a function and an optional config table, rather than a function and a table of a function and a config table. The current way seems a bit clunky. Any thoughts?

akinsho:
I agree re. maybe simplifying it i.e. a function and an optional table. I got tripped up by the table as well when I first set up packer. Not sure how it will work with regards to migrating users. Maybe a temporary shim with a warning message that can then be removed in like 2 weeks

wbthomason:
Looks good to me, thanks!

PR #374, opened 2021-05-25
Feature: detect repo changes

wbthomason:
This is #340, plus my commits to fix up normalizing plugin URLs and tweak the awaiting of the main thread (it should happen as close as possible to the slow Vim function calls that require it).

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -281,6 +281,7 @@ packer.install = function(...)
     local start_time = vim.fn.reltime()
     local results = {}
     await(clean(plugins, results))
+    await(a.main)
     local tasks, display_win = install(plugins, install_plugins, results)
     if next(tasks) then
       local luarocks_ensure_task = luarocks.ensure(rocks, results, display_win)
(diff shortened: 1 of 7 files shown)


wbthomason:
@nanotee and @akinsho (and anyone else), if you're able to verify that this still works for you, that would be great!

akinsho:
@wbthomason yep still works 👍🏾

akinsho (approved):
@wbthomason LGTM tested with a non ssh url though but I'm assuming that works for you

nanotee:
@wbthomason I'm not sure what's causing this but I'm getting an error when I run `:PackerSync`:

```
async.lua:20: Error in coroutine: ...site/pack/packer/sta.../lua/packer/update.lua:58: E5560: vimL function must not be called in a lua loop callback
```

What's weird is that this only happens some of the time. Restarting nvim is usually enough to get rid of the error

Changing a plugin's repository still works as expected

akinsho:
It might relate to having moved the `a.wait(main)`, I was getting that during the initial development till I moved it to the line it was previously at.

wbthomason:
Yes, I missed a case to cover with a main thread schedule. I'll fix it when I'm back at my laptop.

wbthomason:
> @wbthomason LGTM tested with a non ssh url though but I'm assuming that works for you

Yup, I tested both cases of my git config.

wbthomason:
@nanotee I believe I've fixed the error you noted, though it required adding more main thread schedules than I expected. It doesn't belong in this PR, but I should see if I can move away from using functions which must be run on the main thread to alleviate this burden.

nanotee:
@wbthomason Seems to have fixed it! I can't reproduce the error anymore

wbthomason:
Ok, I think we're good to merge this (again). Sorry for the churn; hopefully I haven't missed any edge cases this time.

oblitum:
After this commit `:PackerSync` will remove and reinstall any repo I use outside github (e.g. repos from gitlab which use whole https urls).

PR #380, opened 2021-05-26
Fix #378: Normalize separator direction in detecting plugin repo changes

### LINKED ISSUE #378: sourcehut hosted package gets autoremoved
- `nvim --version`: NVIM v0.5.0-dev+1356-g0b905be47
- `git --version`: git version 2.31.1
- Operating system/version: Arch Linux
- Terminal name/version: kitty 0.20.3

### Steps to reproduce
Add a plugin hosted on sourcehut.

### Actual behaviour
Calling `PackerSync` causes packer to want to remove the package (as if I removed it from my plugin file) and download it again. This didn't happen on packer two days ago.

### Expected behaviour
Just update the plugin or do nothing

### packer files
` use {'https://git.xenrox.net/~xenrox/debug-packer'}`
Packer log
https://paste.xenrox.net/~xenrox/fbe3ba404ccdc13724f05597252e6a72b4e264ca

packer compiled file
https://paste.xenrox.net/~xenrox/2f9f632bf1bfdd33d620fea3eeaff6327157ecd4

wbthomason:
Fixes the bug @uyha reported in the comments of #378, by normalizing the separators in plugin names to match URL separators (i.e. always use `/` instead of Windows' `\`).

@uyha, could you please confirm that this solves the problem? I think it will, but I don't have access to a Windows testing environment, so I'm unsure.

### DIFF
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -131,7 +131,9 @@ plugin_utils.find_missing_plugins = function(plugins, opt_plugins, start_plugins
             -- protocol
             local normalized_remote = remote:gsub("https://", ""):gsub("ssh://git@", "")
             local normalized_plugin_name = plugin.name:gsub("https://", ""):gsub("ssh://git@", "")
-            if (normalized_remote ~= normalized_plugin_name) and (repo_name ~= plugin.name) then
+                                             :gsub("\\", "/")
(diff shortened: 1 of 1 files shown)


uyha:
Yes, this fixes the problem.

wbthomason:
Great, thank you!

Issue #336, opened 2021-05-03
Windows: after_files list is not generated


reiver-dev:
- `nvim --version`:
```
NVIM v0.5.0-dev+1291-gb227cedf8
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
Compilation: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe /DWIN32 /D_WINDOWS /W3 /MD /Zi /O2 /Ob1 /DNDEBUG /W3 -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -DWIN32 -D_WIN32_WINNT=0x0600 -DINCLUDE_GENERATED_DECLARATIONS -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -ID:/a/neovim/neovim/build/config -ID:/a/neovim/neovim/src -ID:/a/neovim/neovim/nvim-deps/usr/include -ID:/a/neovim/neovim/build/src/nvim/auto -ID:/a/neovim/neovim/build/include
Compiled by runneradmin@fv-az152-644

Features: -acl +iconv +tui
```

- `git --version`:
```
bc6feab6b3c1ad5af9aed077857aed8c4e4657f1
```

- Operating system/version:
```
Window 10
Major  Minor  Build  Revision
-----  -----  -----  --------
10     0      19042  0
```

- Terminal name/version:
```
Windows Terminal Preview
Version: 1.8.1032.0
```

### Steps to reproduce
Install nvim-compe to opt. Generate packer_compiled.vim
```
nvim --headless -u NORC --noplugin +"packadd packer.nvim"  +"lua require([[packer]]).startup(function(use) use { [[hrsh7th/nvim-compe]], as = [[compe]], opt = true }; end)" +PackerInstall +PackerCompile  +q
```

### Actual behaviour
after_files not present in packer_compiled.vim

### Expected behaviour
after_files successfully generated and loaded

### Suspected reason
On Windows glob function ignores first directory (plugin) when backslash is used
```
$> nvim --headless -u NORC --noplugin +"echo glob(stdpath('data') . '\site\pack\packer\opt\compe\after\plugin\**\*.vim', v:false, v:true)" +q 
[]
```
```
[...]

PR #384, opened 2021-05-29
Fix recursive glob on Windows, fix filename escape during load

reiver-dev:
Fix issue #336

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -163,7 +163,7 @@ local function make_try_loadstring(item, chunk, name)
   return executable_string, bytecode
 end
 
-local after_plugin_pattern = table.concat({'after', 'plugin', '**', '*.vim'}, util.get_separator())
+local after_plugin_pattern = table.concat({'after', 'plugin', '**/*.vim'}, util.get_separator())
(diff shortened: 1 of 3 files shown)


wbthomason:
@reiver-dev: Thanks for the PR! I'm happy to merge the load change from `silent exe source` to `silent source`; that seems like an oversight. I'm a bit confused about your other changes, though - can you elaborate on the problem they solve? I would expect that hard-coding the forward slash separator in the pattern would actually break things on Windows (hence how the pattern is currently constructed), but I admit I don't have much experience with Vim/Neovim on Windows and could be ignorant of something here.

reiver-dev:
I have checked slash part also with vim8.2.  Vim demonstrates same behavior.

I suppose it is caused by:

https://github.com/neovim/neovim/blob/30befcdce47b4964b4d2a5aeb078ddec8e2fcc0b/src/nvim/path.c#L665
https://github.com/vim/vim/blob/3d9c4eefe656ee8bf58c0496a48bd56bac180056/src/filepath.c#L3595

wbthomason (approved):
Per discussion here and on #336, these seem good. Thank you, and again, sorry for neglecting this issue for so long!

Issue #272, opened 2021-03-31
After key does not trigger loading of dependent plugin if using only cond key


akinsho:
Apologies for the maybe confusing title, also since I think you're planning on a general change to how this works this might be moot but I thought pointing out this specific issue was worth doing anyway. 

Essentially if a plugin needs to be loaded after another, it isn't currently correctly loaded if the plugin it is to be loaded after is loaded after a `cond`. I'd imagin that the condition is run pretty much immediately i.e. during packer compiled's execution, but the after never seems to kick in if the condition is true and that plugin is then loaded

real world eg.

```lua
-- This doesn't work i.e. flutter tools is not loaded
use {"neovim/nvim-lspconfig", cond = function() return vim.env.ENV_VAR ~= nil end}
use  {"akinsho/flutter-tools", after = "nvim-lspconfig", config = function() print('loaded!!') end}
-- this does work
use {"neovim/nvim-lspconfig"}
-- Or
use {"neovim/nvim-lspconfig", event = "VimEnter"}
use  {"akinsho/flutter-tools", after = "nvim-lspconfig", config = function() print('loaded!!') end}
```

akinsho:
On digging in a bit this now makes much more sense and I just didn't understand. The `after` key is *only* relevant when a plugin is being loaded via an autocommand otherwise it does nothing. So it can't be used to control the order of execution for start packages for example 🤔 . Not sure if vim just loads all of these arbitrarily in a random order once installed or if packer can intervene.

Might just be my perspective but I think being able to guarantee ordering for plugins that aren't necessarily lazy loaded is pretty important. I think even if it means maybe making these plugins optional and then loading them during the packer compiled's execution but then doing it sequentially based on the after key etc. Not sure if this deviates from any of what you were thinking, or you have a better idea in mind @wbthomason

wbthomason:
Wow, this slipped my radar. No, `after` should determine sequencing always, not just for autocommand loading. It technically cannot control `start` package order, but only because it implies `opt` - it can absolutely control eagerly-loaded package ordering. This is definitely a bug in the compiler, which might take some redesign to do.

> I think even if it means maybe making these plugins optional and then loading them during the packer compiled's execution but then doing it sequentially based on the after key etc

Good suggestion! This is in fact what we already do :smile: 

The trick is that we currently load `cond` plugins with a direct `packadd`; this should really use the load module function we use for e.g. autocommand lazy-loads to ensure that the chain of sequence-loads gets updated.

[...]

PR #390, opened 2021-06-06
fix: use packer.load for plugins loaded via cond

akinsho:
Attempt to partially address #272 . @wbthomason you mentioned that `cond` plugins should use the load mechanism so the sequencing logic could be taken advantage of. I'm not sure whether just using `packer.load` is enough though since in a very brief test doing this seems to cause the `cond` plugin to load as expected but the `after` plugins do not

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -438,7 +438,9 @@ local function make_loaders(_, plugins, should_profile)
   for condition, names in pairs(condition_ids) do
     local conditional_loads = {}
     for _, name in ipairs(names) do
-      timed_chunk('\tvim.cmd [[packadd ' .. name .. ']]', 'packadd for ' .. name, conditional_loads)
+      timed_chunk(
+        fmt('\trequire("packer.load")({"%s"}, {}, _G.packer_plugins)', name),
+        'packadd for ' .. name, conditional_loads)
       if plugins[name].config then
         local lines = {'-- Config for: ' .. name}
         timed_chunk(plugins[name].executable_config, 'Config for ' .. name, lines)


wbthomason (commented):
This seems like a good step to take anyway. When you say the `after` plugins don't load correctly, you mean that they never get sourced if they're set to load after a `cond` plugin?

akinsho:
@wbthomason yep that was what I meant, sorry missed your comment at the time. Looking at this again I'm not sure if it actually needs anything else

akinsho:
WDYT @wbthomason? does this look alright to merge?

wbthomason:
I'm curious if this will cause problems with `after`, as you noted in June, but any breakage should be restricted to the fairly narrow "using both `cond` and `after`" case. As such, yeah, I think this is good to merge. Thank you!

akinsho:
I dimly recall thinking at the time that the issue was that `cond` wasn't taking into account when resolving `after` plugins so I think (guesstimating from memory) that it's more that the after logic needs to be tweaked rather than something breaking 🤞🏿

shadmansaleh:
`after` is working with `cond` for me .

What I mean is if B plugin has A in `after` option and A is loaded by `cond` then B gets loaded too .

akinsho:
> What I mean is if B plugin has after='A' option and A is loaded by cond then B gets loaded too .
Ah, this is what I thought at the time wasn't working and is the issue I reported in #272 if it works then that's good and unexpected news since I didn't expect that this PR would fix that issue.

shadmansaleh:
I have a similar setup and both plugins are loaded.

```lua
    use {{'neoclide/coc.nvim',
      cond = function()
        local lsp = os.getenv('LSP')
        return lsp == nil or lsp == '1'
      end,
      branch = 'release',
      config = source('$conf/coc_nvim_conf.vim'),
     }, 
    {'shadmansaleh/coc-nvim-lua', after='coc.nvim'},
   }
```

I'd have probably expected it to be solved since after/wants are handled by the plugin loader in packer/loader so that should handle this case too unless there was a bug there .

You can test if after works now .

Issue #201, opened 2021-02-07
Do not put compiled loader in stdpath('config')


clason:
Currently, packer puts the compiled loader script in `stdpath('config')/plugins`. Since I sync `stdpath('config')` across multiple machines, this often leads to conflicts on `:PackerSync`. 

Since the script is generated from user config, I believe it would be more appropriate to put it in `stdpath('cache')` or `stdpath('data')`. In particular, the latter would be easy to handle without new logic by just putting it in, say `.local/share/nvim/site/pack/loader/start/packer/plugin/` instead of `.config/nvim/plugin`. (I tested this.)

(Vim packages allow multiple parallel `pack/**/start` directories, all of which are automatically visible to vim. In fact, I believe the original idea for packages was that each package would have its own `pack/<package>/start` and `pack/<package>/opt` directory, where `start` serves as a replacement for `autoload`.)

tami5:
@clason I believe changing `compile_path` in packer setup function is all you need todo. But I agree the default should be out of user config

wbthomason:
I'm happy to make this change, but I need to think about/figure out the right way to change over from the old default. In particular, we don't want two `packer_compiled` files; we need to either move or delete the old file. This should probably happen via a shim that checks if there's a file in the old location and, if so, prints a warning and deletes the file. We can remove this shim after a month or two.

clason:
Agreed, but I don't see the need to be particularly clever about it -- it's not that the old location is not working. Wouldn't it be enough for `PackerCompile` to check if `packer_compiled` exists in the old location; if so, delete it; and then write it in the new location? (This check can be removed after a while, of course.) Or is that what you meant by "shim"?

wbthomason:
That is exactly what I was suggesting with the "shim", yes. Sorry, that comment was >50% me thinking aloud about the right way to do this.

francisco285:
I think this is a bigger issue now that [#14686](https://github.com/neovim/neovim/pull/14686) is merged, because people will try to use `stdpath('config')/plugin/*.lua` files instead of requiring them in `init.lua` and it will break their configs.

clason:
No, #14686 should have zero (breaking) implication for packer (otherwise we wouldn't have merged it). In fact, I _am_ using `stdpath('config')/plugin/*.lua` with packer plugins and it works just fine.

However, there is one _positive_ implication for packer: it should now be possible to have a `packer_compiled.lua`!

francisco285:
[...]

PR #391, opened 2021-06-07
Avoid using string.dump

shadmansaleh:
Well I've gone a bit off track . This is really cool so I thought lets push it to remote :P

@wbthomason @akinsho with bec6b2ac8b8272390f943a08fa11aa27c1557be7

Packer extracts the functions source code directly from user's configuration and pastes it in packer_compiled . And uses `string.dump` only as last resort.

For now error checks that this pr was mainly about is disabled . But it can be enabled like the 1st commit .

Brefore this:
```lua
-- Setup for: ultisnips
time("Setup for ultisnips", true)
try_loadstring("\27LJ\2\2Ë\1\0\0\2\0\n\0\0176\0\0\0009\0\1\0'\1\3\0=\1\2\0006\0\0\0009\0\1\0'\1\5\0=\1\4\0006\0\0\0009\0\1\0'\1\a\0=\1\6\0006\0\0\0009\0\1\0'\1\t\0=\1\b\0K\0\1\0\ttabo\23UltiSnipsEditSplit\n<C-k>!UltiSnipsJumpBackwardTrigger\n<C-j> UltiSnipsJumpForwardTrigger\n<C-l>\27UltiSnipsExpandTrigger\6g\bvim\0", "setup", "ultisnips")
time("Setup for ultisnips", false)
```

After this :
```lua
-- Setup for: ultisnips
time("Setup for ultisnips", true)
try_loadstring(function()
      vim.g.UltiSnipsExpandTrigger="<C-l>"
      vim.g.UltiSnipsJumpForwardTrigger="<C-j>"
      vim.g.UltiSnipsJumpBackwardTrigger="<C-k>"
      vim.g.UltiSnipsEditSplit="tabo"
    end, "setup", "ultisnips")
time("Setup for ultisnips", false)
```

So much better 😆

Possibly duplicate of #192


lua/packer/compile.lua
```diff
@@ -158,9 +158,19 @@ local function dump_loaders(loaders)
...
 local function make_try_loadstring(item, chunk, name)
-  local bytecode = string.dump(item, true)
-  local executable_string = 'try_loadstring(' .. vim.inspect(bytecode) .. ', "' .. chunk .. '", "'
-                              .. name .. '")'
+  local executable_string, bytecode
+  if type(item) == 'string' then
+    bytecode = item
+    local function_body = fmt(
+[[(function()
+  %s
```
akinsho, line 166: I don't use the string format at all but does this work if my condition was `vim.env.VAR ~= nil` this would just be an expression, it would need to be returned right.

shadmansaleh: Sorry I couldn't understand what you ment there .

> if my condition was vim.env.VAR ~= nil this would just be an expression

Are you talking about condition option here ? If so this is for config & setup option .

Basically if you had
```lua
use{ 'author/plugin',
config = 'vim.env.VAR ~= nil'
}
```

Config portion of It used to get compiled to:
```lua
vim.env.VAR ~= nil
```
Now it compiles to:
```lua
try_loadstring([[(function()
  vim.env.VAR ~= nil
end)()]], "plugin", "config")
```

And we do error check at `try_loadstring`

Either way I can't understand what `vim.env.VAR ~= nil` suppose to do here :P

akinsho: @shadmansaleh what I was alluding to is how is that expression evaluated by packer if it isn't returned? You need to return inside the function otherwise it just returns nil?

shadmansaleh: Packer doesn't care about return value of config/setup . They are just ran for side effects as per my understanding

akinsho: Ah sorry @shadmansaleh I'm being dumb, need more sleep. Yeah was thinking of the `cond` fields and not paying enough attention.

shadmansaleh:
Well I've gone a bit off track .

@wbthomason @akinsho with bec6b2ac8b8272390f943a08fa11aa27c1557be7

Packer extracts the functions source code directly from user's configuration and pastes it in packer_compiled . And uses `string.dump` only as last resort.

For now error checks that this pr was mainly about is disabled . But it can be enabled like the 1st commit .

akinsho:
@shadmansaleh this looks cool and is quite similar to a WIP PR @wbthomason made a while ago #192 I personally am in favour of making the compiled file readable (which is what I said on there) In the PR description @wbthomason mentions
> This is kinda cool, but it's not clear that it's helpful - on the one hand, functions created with loadstring are never JIT'd, but on the other hand, these functions should only run once and are unlikely to benefit from JITing. This approach also creates code duplication by copying the function source to the compiled file.

Is the case the same here i.e. does it provide any particular value other than the improved readability? (again personally in favour of making the compiled file more readable) just wondering if it gives any other advantages or is this less about the string error issue now?

shadmansaleh:
> @shadmansaleh this looks cool and is quite similar to a WIP PR @wbthomason made a while ago #192

Hah I didn't see that pr . Yes they are quite similar.

>  I personally am in favour of making the compiled file readable

Me too . `string.dump` makes it impossible to understand what's going on . If we didn't use lua-haredocs the error locations actually would have been quite correct . (error location is reported from start of haredoc ☹️) 

> Is the case the same here i.e. does it provide any particular value other than the improved readability?
(again personally in favour of making the compiled file more readable) just wondering if it gives any other advantages or is this less about the string error issue now?

I don't know of any other benefit TBH. I was just looking for ways to get rid of `string.dump` To get saner error messages and so compiled files is actuslly readable . This idea seemed interesting . So I was just tinkering with it . After a while when I had a workings setup I thought of sharing the idea and see whay's oyhers opinion are on this . But with #192 it seems this idea has already been explored :)

akinsho:
Tbh @shadmansaleh I think IMO (since I've already established myself as being thoroughly in favour of this idea) it seems worth bringing up. I'm not sure of all the differences in the implementation I think yours caches the files so it doesn't do the reading a bunch of times which is nice, but doesn't cover condition functions.

shadmansaleh:
> or is this less about the string error issue now?

I've split the string error fix to #392 . We can disscuss this issue here .

shadmansaleh:
. I think yours caches the files so it doesn't do the reading a bunch of times which is nice, 

In almost all cases we'll only be reading from a single file again and again so I thought it would be nice to cache instead of reading the file for each plugin :)

> but doesn't cover condition functions.

That's not much of a problem . It'll be just changing 2 lines of code

shadmansaleh:
Ok . I've one more pro for this . Now you can actually find runtime errors in packer_compiled . Basically the line_no in messages are off by one . So whatever line number the error message gives you can open packer_compiled.vim and see the following line of that line_no :) . Also error mesages are better too when saying what's the caise of the error . 

Also now config functions for opt pkugins are stored as funcRef instead of strings too . You guys can give this a try and see if you can find any issues .

lua/packer/compile.lua
```diff
@@ -7,18 +7,57 @@ local luarocks = require('packer.luarocks')
...
+    return table.concat(src, '\n')
+  end
+  if not ok then vim.api.nvim_err_writeln(src) end
+  return vim.inspect(string.dump(func))
+end
 
-try
-]]
+
+local feature_guard = [[if !has('nvim-0.5') | echohl WarningMsg | echom "Invalid Neovim version for packer.nvim!" | echohl None | finish | endif | packadd packer.nvim | try]]
```
akinsho, line 60: nit: is making this one line necessary, it's less readable both in the code and in the compiled file

shadmansaleh: Well the error reports are given based on it possition on where it's in the lua block . For example if the error occurs in line 234 in the compiled file then nvim will report error occured in line 220 . Because current implementation addes 14 lines on top and nvim is coumting lines from `lua << END`. With this nvim will report error in line 220 when error occurs in line line 221 of file . 

This adds way more value IMO .

lua/packer/compile.lua
```diff
@@ -147,20 +191,42 @@ local function timed_condition(condition, body, name, output_table)
   return output_table
 end
 
+local function unwrap_function_src(data)
```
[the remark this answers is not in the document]
shadmansaleh: Yes of cource. I'll add them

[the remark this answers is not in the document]
shadmansaleh: done

shadmansaleh:
https://github.com/neovim/neovim/pull/14686 was merged to core . We should use lua file for packer_compiled . I've added it though I'm not sure how to notify user of the breaking change . 

For anyone trying to try this branch you need to have  neovim master , remove your packer_compiled.vim . Start neovim and run `:PackerCompile` . Also If ylu happen to set your packer compiled location with custom init you'll need to change extension of that filename to lua

wbthomason:
As a note, I plan to take a look at this and #331 tomorrow. As has been noted, this seems like a more robust version of #192 - my initial thinking is that it may be better to just get #331 merged and, as a result, move to having users `require` their configs in their init.lua, therefore directly using the function objects without any serialization or code extraction (which I still worry is brittle). This would keep the number of breaking changes in the move away from manual compilation to 1 (because the remaining compiled file will just be lazy-loaders and precomputed path stuff, which can be hidden from the user) and sets the stage for autorecompilation. I appreciate this work, though, and I could be convinced that my opinion here is wrong.

I want to get the move to auto-recompilation/loading function objects directly done soon, but I also want to be very careful to pick a good design for it and avoid breaking users multiple times.

shadmansaleh:
TBH this has just become my experimental branch. I'd definatly love to see packer use my config directly without compilation . I think moving to plain function object should resolve the error issue . It'll definatly be more robust then this . Actually I found the idea that you could extract source code in lua quite cool. So just shared it here wasn't hoping to get it  merged when function objects are already planned to be used.

If we require users to load packer config at startup wouldn't it make more sense not to have a compiled file at all ? packer could just create those lazy-loaders on the fly.

wbthomason:
See my comment on #201, but: in my early experimentation when first building `packer`, there was still a (slight) startup time advantage to precompiling things like lazy-loaders, `rtp` changes, and some additional paths to source. To my mind, the real annoying complexity of having a compiled file is if you have to rebuild that file yourself. Making users require their config at startup makes it pretty trivial to add auto-recompilation, which removes that annoyance. It would be easy enough to disable compilation in this case too. May also be worth re-running my experiments to see if there's still any benefit to precompilation.

shadmansaleh:
Closing  as this is unnecessary with #402

Issue #396, opened 2021-06-09
Invalid escape sequence near `Sourcing ftdetect script` on windows


EdenEast:
- `nvim --version`: `NVIM v0.5.0-dev+1372-g056c464e8`
- `git --version`: `2.29.2.windows.2`
- Operating system/version: `Windows 10`
- Terminal name/version: `Command prompt`, `Powershell`, and `nvim-qt`

### Steps to reproduce

- Copy `minimal_init.lua`
- Open `cmd`
- Execute `nvim -u minimal_init.lua minimal_init.lua`
  - This should bootstrap packer and install just plugin and execute compile
- Exit neovim
- Re-enter nvim with `nvim -u minimal_init.lua minimal_init.lua`
- Observe error

```lua
-- minimal_init.lua
local fn = vim.fn
local fmt = string.format

-- Setup of temp env
local seperator = jit.os == 'Windows' and '\\' or '/'
local tmp_root = fn.fnamemodify('/tmp/nvim/', ':p')
local pack_path = fn.fnamemodify(tmp_root .. 'site', ':p')
local package_root = fn.fnamemodify(tmp_root .. 'site/pack', ':p')
local compile_path = fn.fnamemodify(tmp_root .. 'plugin/packer_compiled.vim', ':p')
local install_path = fn.fnamemodify(tmp_root .. 'site/pack/packer/start/packer.nvim', ':p')

local function trim(path)
  local r, _ = string.gsub(path, seperator..'$', '', 1)
  return r
end

local function set_env()
  vim.cmd(fmt('set packpath=%s', pack_path))
  vim.o.runtimepath = trim(tmp_root) .. ',' .. vim.o.runtimepath
end

-- Load plugins and configure packer
local function load_plugins()
  local use = require('packer').use

  require("packer").startup(
    {
      function()
        use 'wbthomason/packer.nvim'
        use { 'vmchale/just-vim', ft = { 'just' }}
      end,
      config = {
        package_root = package_root,
        compile_path = compile_path,
      }
    }
  )
end

set_env()
if fn.isdirectory(install_path) == 0 then
  fn.system({'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path})
  load_plugins()
  require('packer').sync()
else
  load_plugins()
end
```

### Actual behaviour

Errors when sourcing `packer_compiled.vim`.

### Expected behaviour

Source `packer_compiled.vim` with no errors.

### packer files

[...]

PR #397, opened 2021-06-09
fix: Ignore escape characters in `timed_chunk` compile sections

EdenEast:
On Windows the `vim.fn.escape` function was not escaping `\` path
characters leaving invalid escape characters in the `ftdetect` section
of the compiled vim file.

This change adds the windows path character as a character to escape.

Fixes: #396

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -120,14 +120,14 @@ end
 
 local function timed_chunk(chunk, name, output_table)
   output_table = output_table or {}
-  output_table[#output_table + 1] = 'time("' .. name .. '", true)'
+  output_table[#output_table + 1] = 'time([[' .. name .. ']], true)'
   if type(chunk) == 'string' then
     output_table[#output_table + 1] = chunk
   else
     vim.list_extend(output_table, chunk)
   end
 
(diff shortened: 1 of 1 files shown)


lua/packer/compile.lua
```diff
@@ -637,9 +637,9 @@ local function make_loaders(_, plugins, should_profile)
   if next(ftdetect_paths) then
     table.insert(result, 'vim.cmd [[augroup filetypedetect]]')
     for _, path in ipairs(ftdetect_paths) do
-      local escaped_path = vim.fn.escape(path, ' ')
+      local escaped_path = vim.fn.escape(path, ' \\')
```
akinsho, line 640: How will this impact non windows machines OOI @EdenEast? haven't tried this myself as I currently don't have the time.

EdenEast: I tested these changes on linux as well and there was no issues. As I went to bed I realized that another solution is to use `[[ ]]` instead of quotes to make the `\` character not be treated as an escape character. I like this solution better for this issue.

akinsho (approved):
LGTM @EdenEast

wbthomason:
This looks good, thank you!

Issue #393, opened 2021-06-08
Abolish command is not available in config


nawordar:
- `nvim --version`: 
```
NVIM v0.5.0-dev+nightly-103-g93f15db5d
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: /usr/bin/cc -O2 -g -Og -g -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wmissing-prototypes -Wimplicit-fallthrough -Wvla -fstack-protector-strong -fno-common -fdiagnostics-color=always -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -DMIN_LOG_LEVEL=3 -I/home/nawordar/clones/neovim/build/config -I/home/nawordar/clones/neovim/src -I/usr/include -I/home/nawordar/clones/neovim/build/src/nvim/auto -I/home/nawordar/clones/neovim/build/include
Compiled by nawordar@arch-pc

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

Run :checkhealth for more info
```
- `git --version`: `git version 2.31.1`
- Operating system/version: Arch Linux
- Terminal name/version: Kitty 0.20.3

### Steps to reproduce

Add the following plugin spec to the config:
``` lua
use {
  'tpope/vim-abolish',
  config = function ()
    vim.cmd 'Abolish cosnt const'
  end,
}

```

### Actual behaviour

An error message pops out:

`packer.nvim: Error running config for vim-abolish: [string "..."]:0: Vim:E492: Not an editor command: Abolish cosnt const`

I am able to run this command manually after dismissing the popup with `:Abolish cosnt const`.

### Expected behaviour

It executes the `:Abolish` command without error

### packer files

<details>
<summary>Plugin specification file(s)</summary>

[MRE repository](https://github.com/nawordar/packer-command-not-available-mre)

</details>

<details>
<summary>packer log file</summary>

```
[INFO  Tue 08 Jun 2021 07:19:00 PM CEST] ./data/site/pack/packer/start/packer.nvim/lua/packer.lua:491: Finished compiling lazy-loaders!
[...]

Issue #307, opened 2021-04-20
Idea: Install all plugins as opt plugin .


shadmansaleh:
### Idea

Install all plugins as opt plugin . Means by default install them in stdpath('data')/site/packer/opt . The plugins that aren't configured to be a opt plugin load them at startup.

### Pros
- setup works for all plugins . It's specially required for configureing vim plugins
- Plugins don't effect different setup if packer_compiled.vim isn't it plugin/ (It's planned to be moved to cache I think). Example `nvim -u config1.lua` and `nvim -u config2.lua` can have different plugins their start plugins won't be loaded for both cases.
- Switching a plugin from opt to non opt and vise-versa can be done with `:PackerCompile` no need for `:PackerSync` or moving arround plugin directories
### Cons
Not one I can think of right now .

akinsho:
Tbh this seems to me to be working around the behaviour of vim packages in general whereas I think if this was a path to explore for the reasons above then moving to runtimepath management as discussed in #71 is probably a better solution used by projects like `vim-plug`, `dein` etc.

Regarding the idea in general I think if using vim packages then packer should honour the way that they work rather than add another layer to their behaviour that isn't clear to users. Also given different use cases I'm not sure this behaviour is what everyone would want.

Tbh other than the first pro, IMO the second two don't really seem like big problems warranting quite a large refactor. `PackerSync` combines install and compile so user only has to run one command to move plugin dirs and I imagine this isn't something people are doing very frequently.

Regarding the second point about different configs packer already has quite a few keys to control which plugins are installed or loaded etc. I run my config on several machines for different usecases by making heavy use of keys like `disable` and `opt` and `cond` etc.

I don't have a strong opinion about moving to another strategy but one thing I don't like about vim packages is the behaviour of `start` plugins in general which is similar to your first pro i.e. that they are quite hard to switch off or disable. A user can't easily do something before they load and cant easily stop them from loading (short of deleting or moving them)

shadmansaleh:
> Tbh this seems to me to be working around the behaviour of vim packages in general whereas I think if this was a path to explore for the reasons above then moving to runtimepath management as discussed in #71 is probably a better solution used by projects like vim-plug, dein etc

Yeah it's a workaround . I thought it was better because this should be less work to achieve the same thing :)

[...]

PR #402, opened 2021-06-13
Big breaking change: Make the compiled file do less and implement autorecompilation

wbthomason:
@shadmansaleh @akinsho Now that #331 is merged, this PR will be the big breaking change to:
1. Require users to load their `packer` config always
2. Reduce the compiled file to lazy-loaders and any precomputed paths we can generate/luarocks setup
3. Change to use the "live" values from the spec for `setup` and `config` as well as `cond` and `after`
4. Implement autorecompilation by storing hashes of the strings that specify lazy-load conditions and regenerating the compiled file if the current values don't match
5. Move the default compiled file location to `stdpath('cache')`
6. Clean up the code while we're in there, try to use `luv` instead of e.g. `vim.fn` wherever possible, in general simplify and tidy things which have grown cruft over the course of `packer`'s existence so far.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -543,6 +543,12 @@ end
 --- Update the compiled lazy-loader code
 --- Takes an optional argument of a path to which to output the resulting compiled code
 packer.compile = function(raw_args)
+  -- TODO: check if the user wants any compilation at all via a new config variable
+  -- TODO: remove a bunch of complexity from the compile module/maybe here once we don't need to
+  -- compile functions or sequencing
+  -- TODO: implement, for the compiled file, logic to check its values against the current config.
+  -- We can do this by computing an efficient hash of the lazy-loader strings and storing this in
+  -- the compiled file
   local compile = require_and_configure('compile')
   local log = require_and_configure('log')
 
@@ -657,6 +663,13 @@ packer.startup = function(spec)
   end
 
   return packer
+end
+
+packer.load = function()
+  -- TODO: This will be responsible for (1) sourcing the new, lighter, compiled file and (2)
+  -- handling things like running setup and config functions, sequencing, etc. 90% of this logic
+  -- already exists in the load module; the main thing to add is loading the new compiled file. This
+  -- function needs to be called at the end of startup()
 end
 
 return packer


wbthomason:
Oh: it's probably out of scope for this PR, but I want to design this change such that it enables adding plugins imperatively, i.e. via a command. Probably a good idea to look at how `paq` added this, specifically w.r.t. how they add the specs for the imperatively added plugins to the overall set of plugin specifications.

wbthomason:
This PR is also going to need some mechanism for migrating users, e.g. by deleting their old `packer_compiled` files if they exist. So I guess that's (7),

shadmansaleh:
> Implement autorecompilation by storing hashes of the strings that specify lazy-load conditions and regenerating the compiled file if the current values don't match

Will auto compilation happen at startup ? when will the check happen actually?

> Move the default compiled file location to stdpath('cache')

Any comment on #307 . Things like #393 will continue to be an issue but more unexpectedly as some plugins will work and some won't based on what filesname packer choses for the loader.

> Require users to load their packer config always

What happens when they are out of sync? Say I wanted to test out someone elses config or a framework more importently trying to isolate a plugin to file a bug report .

wbthomason:
> Will auto compilation happen at startup ? when will the check happen actually?

Right now, I'm thinking the check happens at startup; recompilation happens on `packer` operations or at startup iff a difference is detected.

> Any comment on #307 . Things like #393 will continue to be an issue but more unexpectedly as some plugins will work and some won't based on what filesname packer choses for the loader.

#307 is on my queue to reply to. From what I remember the last time I looked at it, it's somewhat orthogonal to this set of changes (i.e. we can make the move to be fully-opt if we decide to without causing a breaking change), but, regardless, the name of the compiled file should not matter - it won't be somewhere that Neovim sources, but will be explicitly loaded/run by `packer` during startup.

> What happens when they are out of sync? Say I wanted to test out someone elses config or a framework more importently trying to isolate a plugin to file a bug report .

The idea of autocompilation would be that, when you're testing another config, your compiled file automatically gets rebuilt and sourced, and when you go back to your usual config, it gets rebuilt and sourced again for you. So all you see is a slightly slower startup time those two times. I think there is a question here around how we handle using different configs e.g. with cleaning "unused" plugins, etc., but in my mind that's more of an argument in favor of #307.

> If I comment out some plugins will that be an erroe?

No. Without #307, if they're `start`, they'll still get loaded, and regardless if they have lazy-loaders those will get removed. This shouldn't be an error, though.

shadmansaleh:
Ok got it .

> regardless, the name of the compiled file should not matter - it won't be somewhere that Neovim sources, but will be explicitly loaded/run by packer during startup.

I wasn't meaning compiled file but the file in `packer.nvim/plugin` that loads that compiled file.

shadmansaleh:
So we need to make `use` handle creation of `_G.packer_plugins`. The compiled file will just invoke load module and load module executes based config/setup/cond from `_G.packer_plugins` . Does it seem right ?

wbthomason:
> I wasn't meaning compiled file but the file in packer.nvim/plugin that loads that compiled file.

Ah, my bad. Yes, I think that without #307 or similar, which file the user chooses to call `packer`'s loader from could cause issues similar to #393.

> So we need to make use handle creation of _G.packer_plugins. The compiled file will just invoke load module and load module executes based config/setup/cond from _G.packer_plugins . Does it seem right ?

I have a slightly different architecture in mind:
- `use` still just adds plugins to the managed set
- "Automatically" at the end of `packer.startup` or manually if the user doesn't want to use `startup`, a function is called which:
    1. Loads the compiled file, which checks if it's out of date and rebuilds and resources itself if so, then sets up Luarocks paths, lazy-load triggers, precomputed path things, and maybe other things it helps to precompute in the future.
    2. Invokes the load module for all the start functions, handles sequenced loading, and handles conditional loading
- The load module does basically what you say, using the config/setup values from `_G.packer_plugins`.

akinsho:
@wbthomason this all sounds good.

One question, would a user still have to keep all they configs/setup in the main packer file. Not sure anything that's been mentioned implies that but we previously discussed that maybe being required and I'm curious if that might be a new restriction but I don't see why.

I'm wondering how you want to tackle this I appreciate that there's a breaking change and maybe the sense that this can/should be done in one PR but I'm wondering if this work can't be chunked up into bits maybe some code changes that aren't used till a particular point where things finally get switched over.

I'd be happy to help with bits of it as time permits but I guess I'm wondering how we can split this out in a way where it's easier for multiple people to pick up chunks until we reach a stage where we want to flip the switch and move everything over. Or maybe making incremental PRs into this one 🤷🏾 .

wbthomason:
> would a user still have to keep all they configs/setup in the main packer file

They should still be able to split their config across files, but must ensure that any files specifying plugins they want to use are sourced *before* they call the new loader function in their config. This is compatible with all the current use cases I'm aware of; they could always just call the loader function in a file in `~/.config/nvim/plugin/` or something similar.

> how we can split this out in a way where it's easier for multiple people to pick up chunks

Yeah, splitting this up would be smart. I think I prefer to make multiple PRs into this PR branch, but I'm open to suggestions.

I think there are a few separable chunks to this work:
1. Implement the new loader function (including pulling functionality out of the compile module)
2. Implement the autorecompilation mechanism
3. Implement whatever migration logic we need (i.e. printing warnings, deleting old compiled files)
4. (after the rest) cleanup/simplification work

An implementation of the ideas in #307 could also fit into this big set of changes, as part of (1) and (2). That said, I'm not sure if it's wise to change so much all at once.

clason:
To that last point, I currently (with the new Lua runtime PR merged) have
1. a `plugin/packer.lua` that just contains the list of plugins and a `require'packer'.startup {plugins, config = config}` call (no plugin config!)
2. for every plugin, a `plugin/<plugin>.lua` file with the corresponding setup call etc.

This is working very well for me right now, and I would like to keep it this way ;)

wbthomason:
@clason That should still be 100% compatible with both this PR and #307.

shadmansaleh:
Ok if you want to do it that way . Actually I'm not much concerned about #307 ever since I figured I can change XDG env vars to move packers location. Also I can very well set a system like #307 from currently available config options.

> ftdetect files that need to be sourced for opt plugins/other information we can compute at update/install time.

I feel like we don't need to keep track of ftdetect files.
We can just run

```vim
runtime! OPT ftdetect/**/*.vim
runtime! OPT ftdetect/**/*.lua
```

Nvim itself basically runs `runtime` for sourcing rtp stuff don't think it'll be horribly slower . Also I think bfredl is working on some directory caching for source.

weilbith:
Hey, I'm a bit late on the party. So I guess it is too late anyways. My whole NeoVim configuration approach is a different one. And thereby I really like the separation of only specifying plugins when I want to update or compile them. Else I just have the compiled file and that's it. Any chance I can keep this mechanism or would this require me to switch the plugin manager or fork? 

Today I was working on a better stringify function for the compilation. It allows to also have metatables being put into the compilation output. This for example allows to create a `config` as a table that includes all the context (must not be the name and the specification but can be anything) and attach a `__call` meta function to it. After all this is basically a closure implementation that can be stringified and parsed again. I use `load` instead of `loadstring` for this. 
But I guess that isn't the root cause for these changes and after them it doesn't matter anymore.

PR #407, opened 2021-06-15
feat: allow fully Lua `packer_compiled` file

Iron-E:
# Summary

This pull request allows users to specify a Lua file in their `packer.init` configuration, like so:

```lua
local packer = require 'packer'
packer.init {['compile_path']=vim.fn.stdpath 'config'..'/plugin/packer.lua'}
```

This uses new Lua feature gate and error catching templates. It is ~~not~~ enabled by default.

# Relevant PRs

* #385 proposed a similar idea; however this PR allows users with a VimL file terminating their custom `compile_path` to maintain the expected output.
* #402 shouldn't cause any issues but I'm not entirely clear on how these changes are going to be realized.

# Progress

* [x] Detect Lua `compile_path`
* [x] Create Lua templates for `packer_compiled`
* [x] Implement Lua compilation
* [x] Documentation


leiserfg:
Why it's not safe to remove the old `packer_compiled.*` and replace it with the new `packer_compiled.lua` by default? If it's to keep backward comatibility, why not just make a tag the pointing to the current master (so users that don't wanna update can keep using it's old `packer_compiled.vim`)  and the rest of us using nvim as rolling release don't have to change our settings?

clason:
If you don't want to change settings, you shouldn't be using pre-release software :)

I see this so often on social media that I need to comment on this wherever I see it: Neovim is _not_ a "rolling release". What is often blithely -- and incorrectly -- called "neovim 0.5" is the current state of development of what will (very soon) become the 0.5 release. Although this has been remarkably stable (as in: doesn't crash or have nasty regressions, unless you're opting into tree-sitter) for quite a while, it does _not_ mean it's stable as in "the API is frozen"; if this is what you want, stick with released (point) versions. By transitivity, this also applies to all plugins that target the upcoming 0.5 release. 

Instead, you should treat "following HEAD" as "early access" (and report bugs and issues), but be prepared to run into issues and having to make changes to your config to enjoy the latest features.

---

And in this case, it's not necessarily just for backwards compatibility -- it makes sense to have such a major internal change opt-in for a while to allow to revert to the previous functioning behavior in case something doesn't behave as intended while still being able to test (and report issues with) the new code path.

leiserfg:
Packer asks for nvim 0.5  which does not exist yet, if that's not a rolling release dependency, what is it?

clason:
Exactly what I wrote above.

ReneganRonin:
a rolling release does not have versioning. much like why arch linux does not versions. some people usually mistook nightly as rolling release when it isnt i guess

leiserfg:
> a rolling release does not have versioning. much like why arch linux does not have versions. some people usually mistook nightly as rolling release when it isnt i guess

If you use nightly is technically the same.

clason:
Yes, but "nightly" does _not_ mean "rolling release", it means "prerelease".

ReneganRonin:
> 
> 
> 
> 
> 
> 
> 
> 
> > a rolling release does not have versioning. much like why arch linux does not have versions. some people usually mistook nightly as rolling release when it isnt i guess
> 
> 
> 
> If you use nightly is technically the same. 

a nightly is not a rolling release because it is a prerelease which means it will become the next "stable" once released.

ReneganRonin:
a rolling release will never be stable

clason:
> > > a rolling release does not have versioning. much like why arch linux does not have versions. some people usually mistook nightly as rolling release when it isnt i guess
> > 
> > 
> > If you use nightly is technically the same.
> 
> a nightly is not a rolling release because it is a prerelease which means it will become the next "stable" once released.

Hopefully not, since the nightlies can be arbitrarily broken :)

(Rolling releases can be broken as well; they only make sense if you have a _very_ extensive test suite, like Linux distributions. Vim is a bit weird in that it has a sort of rolling release model without such a test suite (it does have tests, but they don't include the ecosystem), but it's a) much older and more battle-tested already and b) much more conservative in making changes. Again, neovim is definitely not  -- and will not become -- a rolling release.)

And with that, might I suggest we move this off-topic discussion (if anyone wants to continue it) to https://gitter.im/neovim/neovim? (@wbthomason feel free to mark these comments -- at least mine -- as "off topic".)

leiserfg:
Sorry for the fuss, feel free to remove the comments regarding the tecnhnisism of nightly vs rolling release. 
My point is not how to call it,  but that in any case the format of the file for the packages loading is just an implementation detail, if it is error prone then ok, make it opt in under a flag, but any one using nvim 0.5 knows it's not the stable version and should be acceptant of certain level of inestavilities. If you make the `packer_compiled.lua` optional and not the default, it will be less tested, as not every one is checking the new options that are available, and I asume we wanna have only one implementation instead of two.  
Take my opinion as a grain of salt and please don't reply trying to teach me what is a rolling release :smile:

Iron-E:
After some consideration I am inclined to agree with @leiserfg. My original concern was that users may be manually sourcing the compiled file, but this is offset by the fact that the language targeted is detected by the path. So any users moving the file out of the "plugin/" directory won't be affected.

And further considering that 0.5 is still pre-release, I think if there ever was a time to make a change like this, it would be better now than later when 0.5 has been broadly pushed into stable channels.

I'm going to rebase onto master and push some new commits which set it to compile by-default to Lua. These commits can be reverted if maintainers request 👍

README.md
```diff
@@ -251,7 +251,7 @@ default configuration values (and structure of the configuration table) are:
 {
   ensure_dependencies   = true, -- Should packer install plugin dependencies?
   package_root   = util.join_paths(vim.fn.stdpath('data'), 'site', 'pack'),
-  compile_path = util.join_paths(vim.fn.stdpath('config'), 'plugin', 'packer_compiled.vim'),
+  compile_path = util.join_paths(vim.fn.stdpath('config'), 'plugin', 'packer_compiled.vim'), -- can also be a ".lua" file
```
wbthomason, line 254: This should be updated to the new default of `packer_compiled.lua`

doc/packer.txt
```diff
@@ -186,7 +186,7 @@ default values: >
   {
     ensure_dependencies   = true, -- Should packer install plugin dependencies?
     package_root   = util.join_paths(vim.fn.stdpath('data'), 'site', 'pack'),
-    compile_path = util.join_paths(vim.fn.stdpath('config'), 'plugin', 'packer_compiled.vim'),
+    compile_path = util.join_paths(vim.fn.stdpath('config'), 'plugin', 'packer_compiled.vim'), -- can also be a ".lua" file
```
wbthomason, line 189: Same as previous comment, should be updated to the new default.

doc/packer.txt
```diff
@@ -280,7 +280,7 @@ installs/updates the given plugin.
 Providing the `installer`/`updater` keys overrides plugin type detection, but
 you still need to provide a location string for the name of the plugin.
 
-POST-UPDATE HOOKS                              *packer-plugin-hooks* 
+POST-UPDATE HOOKS                              *packer-plugin-hooks*
```
wbthomason, line 283: Nitpick: there seem to be a number of "empty" changes to headings in the docs. It's possible that these are whitespace changes (I'm reviewing on my phone, and not sure that the GitHub Android app renders whitespace diffs well), but if they're not deleting dangling whitespace or similar, I'd prefer they be removed to keep the diff clean.

wbthomason (changes requested):
This looks good, thanks! I'm OK with defaulting to the pure-Lua compiled file, and this will serve as a reasonable basis for the reduced compiled file that #402 will introduce. I have a few nitpick requests commented on the diff (just docs tweaks), and one significant request before merging: I think we should, if a user has indicated that they want a Lua compiled file (resp. Vim) but has a Vim compiled file (resp. Lua) currently in place, delete the old compiled file with the wrong file type. I imagine we'll remove the Vim output and this migration mechanism entirely before too long, but I think proactively ensuring that users have only a single compiled file and not duplicated Lua/Vim compiled files will save a lot of headache and issue report debugging.

clason:
> And further considering that 0.5 is still pre-release, I think if there ever was a time to make a change like this, it would be better now than later when 0.5 has been broadly pushed into stable channels.

I'd try to get it in this week, then (for no particular reason 😉 )

README.md
Iron-E: Oops, good catch!

doc/packer.txt
Iron-E: I can confirm these are whitespace diffs.My Neovim has been configured to strip trailing whitespace on save; it seems parts of this file got caught in the crossfire. I didn't mean to update this part of the file— would you like me to revert the changes here?

lua/packer.lua
```diff
@@ -549,21 +549,32 @@ packer.compile = function(raw_args)
...
-  if config.auto_reload_compiled then vim.cmd("source " .. output_path) end
+  if config.auto_reload_compiled then
+    vim.cmd('source ' .. output_path)
+  end
   log.info('Finished compiling lazy-loaders!')
   packer.on_compile_done()
+
+  -- TODO: remove this after migration period (written 2021/06/28)
+  if output_lua then
+    local old_output_path = vim.fn.fnamemodify(output_path, ':r')..'.vim'
```
Iron-E, line 573: I didn't see that the `packer_compiled` file registers its current location on startup (please correct me if I am wrong), so I decided to check for an old `packer_compiled` file by removing the `output_path`'s extension and appending '.vim' to it. This has a gotcha (see the commit message) but will handle the primary case of migrating users attempting to compile from one filetype to another in the same directory.

If the `packer_compiled` does log its location on startup, I can reference that here instead.

clason: Honestly, I never saw the rationale for auto-deleting "inconsistent" `packer_compiled` -- this is early access; just put a big fat warning somewhere that if anything looks fishy, manually delete the `packer_compiled` file and try again. It'll cease to be a problem within days, anyway.

wbthomason: @clason: That's an option; the primary rationale for auto-deleting (from my perspective) is that having multiple `packer_compiled` files which both get sourced has caused a large number of annoying-to-debug problems for users in the past (e.g. issues where the user's spec and compiled file look fine and don't explain a behavior caused by an old compiled file they didn't realize was present). Adding a warning would help with this, but auto-deleting seems cleaner to me.

seblj: Correct me if I'm wrong, but I think the one who implemented the lua runtime made it so only one of them would be sourced if there were both a lua file and a vim file

clason: It's more convenient for sure; I was just thinking since this is likely a one-time scenario (in a pre-release software, if you can get it in this week ;)), that seemed a bit like over-engineering to me.

wbthomason: It's a fair point, yeah.

Iron-E: I think it's worth it, given the aforementioned possibility of an obscure problem occuring for users. However, I think we should raise a warning and tell users to update their Neovim if they haven't already.

That way, we hopefully reduce the number of issue reports for the `packer_compiled` not being sourced and for the `packer_compiled` sourcing twice.

wbthomason:
Just noting that I haven't forgotten about this and plan to review the changes and probably merge sometime tomorrow.

doc/packer.txt
wbthomason: No, that's OK. Stripping whitespace is fine. Thanks!

lua/packer.lua
```diff
@@ -549,21 +549,34 @@ packer.compile = function(raw_args)
   manage_all_plugins()
   local args = parse_args(raw_args)
   local output_path = args.output_path or config.compile_path
+  local output_lua = vim.fn.fnamemodify(output_path, ':e') ~= 'vim'
```
wbthomason, line 552: Nitpick: this condition should be `== 'lua'` rather than `~= 'vim'`.

wbthomason (approved):
This looks good! I'm going to test it out and push a commit tweaking some formatting changes to be consistent, but otherwise I think it's good to merge. Thank you!

wbthomason:
Works like a charm for me! Good to merge; thank you for this!

wbthomason:
One note: this does leave the possibility of someone moving back to a `vim` compiled file (for some reason) and ending up with two compiled files. I'll push a quick fix to this.

Iron-E:
Awesome! Thanks for reviewing, and I'm happy to help!

PR #410, opened 2021-06-16
feat: open packer logs from display

### LINKED ISSUE #409: Open `packer.nvim.log` from display
### Describe the feature

When I have an issue with a packer operation it is helpful to check the log file to read the full error and diagnose the issue. Having a mapping in the display to open the log would aid in debugging issues.

dhruvmanila:
fixes: #409 

This is similar to the diff window.

We could also implement a command like `PackerLog` to open the log file.

Also, should we open it instead in a new tab as the lines could get pretty long?

### DIFF
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -124,6 +124,10 @@ local keymaps = {
     rhs = '<cmd>lua require"packer.display".prompt_revert()<cr>',
     action = 'revert an update',
   },
+  open_log = {
+    rhs = '<cmd>lua require"packer.display".open_log()<cr>',
+    action = 'open packer logs',
+  },
 }
 
 --- The order of the keys in a dict-like table isn't guaranteed, meaning the display window can
@@ -133,6 +137,7 @@ local keymap_display_order = {
   [2] = 'toggle_info',
   [3] = 'diff',
   [4] = 'prompt_revert',
+  [5] = 'open_log',
 }
 
 --- Utility function to prompt a user with a question in a floating window
@@ -690,6 +695,21 @@ local display_mt = {
     end
   end,
 
+  open_log = function(self)
+    if not self:valid_display() then
+      return
+    end
+    if next(self.items) == nil then
+      log.info 'Operations are still running; plugin info is not ready yet'
+      return
+    end
+    log.open()
+    vim.cmd [[wincmd P]]
+    vim.bo.buflisted = false
+    vim.api.nvim_buf_set_keymap(0, 'n', 'q', '<cmd>close!<CR>', { silent = true, noremap = true, nowait = true })
+    vim.bo.filetype = 'log'
+  end,
+
   --- Heuristically find the plugin nearest to the cursor for displaying detailed information
   find_nearest_plugin = function(self)
     if not self:valid_display() then
@@ -854,5 +874,11 @@ end
 
 --- Async prompt_user
 display.ask_user = a.wrap(prompt_user)
+
+display.open_log = function()
+  if display.status.disp then
+    display.status.disp:open_log()
+  end
+end
 
 return display
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -704,7 +704,7 @@ local display_mt = {
       return
     end
     log.open()
-    vim.cmd [[wincmd P]]
+    vim.wo.wrap = false
     vim.bo.buflisted = false
     vim.api.nvim_buf_set_keymap(0, 'n', 'q', '<cmd>close!<CR>', { silent = true, noremap = true, nowait = true })
     vim.bo.filetype = 'log'
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -126,7 +126,7 @@ local keymaps = {
   },
   open_log = {
     rhs = '<cmd>lua require"packer.display".open_log()<cr>',
-    action = 'open packer logs',
+    action = 'open log file',
   },
 }
 
--- a/lua/packer/log.lua
+++ b/lua/packer/log.lua
@@ -169,7 +169,7 @@ log.new(default_config, true)
 function log.open()
   -- This should be consistent with the format mentioned in log.new above.
   local logfile = string.format('%s/%s.log', vim.fn.stdpath('cache'), default_config.plugin)
-  vim.cmd(string.format("pedit %s", logfile))
+  vim.cmd(string.format("tabedit + %s", logfile))
 end
 
 return log
--- a/lua/packer/log.lua
+++ b/lua/packer/log.lua
@@ -166,4 +166,10 @@ end
 log.new(default_config, true)
 -- }}}
 
+function log.open()
+  -- This should be consistent with the format mentioned in log.new above.
+  local logfile = string.format('%s/%s.log', vim.fn.stdpath('cache'), default_config.plugin)
+  vim.cmd(string.format("pedit %s", logfile))
+end
+
 return log
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -231,6 +231,7 @@ default values: >
         toggle_info = '<CR>',
         diff = 'd',
         prompt_revert = 'r',
+        open_log = 'l',
       }
     }
   }
@@ -459,6 +460,8 @@ Once an operation completes, the results are shown in the display window.
 q                    close the display window
 <CR>                 toggle information about a particular plugin
 r                    revert an update
+d                    show the diff for a particular commit
+l                    open the log file in a new tab
 
 They can be configured by changing the value of `config.display.keybindings`
 (see |packer-configuration|). Setting it to `false` will disable all keybindings.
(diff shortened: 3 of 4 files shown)


lua/packer/log.lua
```diff
@@ -150,4 +150,10 @@ end
 log.new(default_config, true)
 -- }}}
 
+function log.open()
+  -- This should be consistent with the format mentioned in log.new above.
+  local logfile = string.format('%s/%s.log', vim.fn.stdpath('cache'), default_config.plugin)
+  vim.cmd(string.format("pedit %s", logfile))
```
akinsho, line 156: I wonder if preview is a good default for the log. As you mentioned in the issue this closes it seems like it will be quite hard to scan, but I haven't tried this locally, so I could be wrong.

dhruvmanila: Yes, it would be. My suggestion would be to open it in a new tab and when the user will quit the tab, they will be redirected to the packer window. There's no additional code to be written here as with `tabedit`, the file will be opened in a new tab next to the current one and when that is closed, the user will be redirected to the previous tab.

nanotee: Tiny suggestion: `pedit`, `tabedit` and other similar commands support a `+cmd` argument that allows you to position the cursor at a specific location. `tabedit + foo` positions the cursor at the end of file `foo`, which can be pretty useful for logs

dhruvmanila: TIL! I always wondered what `+cmd` do. Thanks for the suggestion! :)

lua/packer/display.lua
```diff
@@ -101,6 +101,10 @@ local keymaps = {
   prompt_revert = {
     rhs = '<cmd>lua require"packer.display".prompt_revert()<cr>',
     action = 'revert an update'
+  },
+  open_log = {
+    rhs = '<cmd>lua require"packer.display".open_log()<cr>',
+    action = 'open packer logs'
```
akinsho, line 107: I'd reword this to `open log file` or something, all the commands are in the context of packer so seems redundant for this one command to explicitly mention packer

akinsho (commented):
Just tried this out, and it works really well 👍🏾, I'd be happy to merge this (granted @wbthomason doesn't come back with some feedback) I just had one nit about how the command worded

lua/packer/display.lua
dhruvmanila: Yeah, makes sense. Thanks for the suggestion!

dhruvmanila:
Hi, @akinsho I've made the necessary changes. I've also updated the docs with the new keybindings and also added the documentation for the diff window key. Let me know if there's any other changes to make :)

lua/packer/log.lua
```diff
@@ -150,4 +150,10 @@ end
 log.new(default_config, true)
 -- }}}
 
+function log.open()
+  -- This should be consistent with the format mentioned in log.new above.
+  local logfile = string.format('%s/%s.log', vim.fn.stdpath('cache'), default_config.plugin)
+  vim.cmd(string.format("tabedit + %s", logfile))
```
wbthomason, line 156: Should we make the command for opening the log configurable? Maybe this should be tied to `config.display.open_fn`/`config.display.open_cmd`?

lua/packer/display.lua
```diff
@@ -708,4 +727,8 @@ end
 --- Async prompt_user
 display.ask_user = a.wrap(prompt_user)
 
+display.open_log = function()
+  if display.status.disp then display.status.disp:open_log() end
```
wbthomason, line 731: Perhaps we should separate displaying the log from having an active display? e.g. by opening a new display if there isn't already one, and additionally creating a command to open the log?

lua/packer/display.lua
```diff
@@ -571,6 +576,20 @@ local display_mt = {
     end
   end,
 
+  open_log = function(self)
+    if not self:valid_display() then return end
+    if next(self.items) == nil then
+      log.info('Operations are still running; plugin info is not ready yet')
```
wbthomason, line 582: Should showing the log be tied to plugin operations running?

wbthomason (changes requested):
Thanks! This looks like a useful feature. I left a few comments - mostly, I think that we don't need to tie displaying the logs to having an active display window (i.e. for filing an issue, people may want to just run a command to open the logs), and I think we should make the display for the logs either configurable or, ideally, tied to the existing `open_fn`/`open_cmd` config variables.

lua/packer/log.lua
akinsho: One thought regarding this is that though this would allow floating windows to work it means that the user would have to have their packer update buffer etc. work the same as their log buffer so if you used a tab for logs you'd also get a tab for a status buffer but that isn't really blocking I think it means people will have to stick with splits for logs

akinsho: I'm sure there will be people who disagree (somewhat invalidating this comment 😆 ) but I think opening logs doesn't really make sense inside of a floating window, or a small split.

I appreciate that we're very much in the middle of the floating window trend but surely for readability of log files there isn't that much value to having them open in a bunch of different ways.

Anyway since people are likely to ask for flexibility regardless 🤷🏾

dhruvmanila: No disrespect to people's preferences, but I don't think that should be configurable. The options to open are:
- Split: With this we would have 3 active splits in the current tabpage namely the current buffer, packer window and the log file. This would make viewing the logs uncomfortable with all the scrolling required.
- Floating window: In cases where the user wants to go back to either the packer window or the working buffer, the floating window will have to be closed and reopened again to view the logs.
- Tab: With the log file opened in a new tab, the user has the flexibility to move back and forth and closing the log file is just a single `q` away. It would also be comfortable to view the logs.

Similar examples: `:checkhealth`, `:LspInfo`
Although, the latter opens in a floating window.

---
This is not related to this PR but it seems that floating windows cannot be split further. This means if the user configured packer to be opened in a floating window and if they're trying to access the diff of a commit, it will open in a split in the previous buffer and the floating window will be on top of it. Thus, to access the diff window the user will have to close the floating window, but that would lose access to other diffs which the user would like to view.

lua/packer/display.lua
dhruvmanila: > We could also implement a command like `PackerLog` to open the log file.

Do you mean something like this? That would be a good idea.

> by opening a new display if there isn't already one

I don't quite understand what you mean by this. Are you saying we should open a packer display if there isn't one? What if we just provide a command which calls `log.open()`?

lua/packer/log.lua
nanotee: Alternatively, making it a user command (e.g. `:PackerLog`) would allow supporting modifiers. Split by default, `:vertical PackerLog` to split vertically, `:tab PackerLog` to open a tab (IIRC Neovim is supposed to get a `:float` modifier at some point as well).

It's less configuration for users and it mimics existing Ex commands

wbthomason: Thanks for the discussion; these are reasonable points. I think I favor @nanotee's suggestion to allow supporting modifiers via the `PackerLog` command, *but* to have the keybinding in the results window bound to `:tab PackerLog`, to open in a tab if the results split/float is already open (incidentally, I'm unsure how opening a new tab works when you're in a floating window - I assume it's as normal, but there might be an undesired interaction there?)

lua/packer/display.lua
wbthomason: Yes, sorry - I saw your `PackerLog` comment right after I finished my review.

> I don't quite understand what you mean by this. Are you saying we should open a packer display if there isn't one? What if we just provide a command which calls log.open()?

Unless I misread the code, it currently only opens the log display if there is already a `packer` results window open, correct? What I was proposing was that the command should *always* display the logs whether or not there's already a `packer` window.

lua/packer/log.lua
dhruvmanila: That seems like a good solution. So, correct me if I'm wrong, but:
- `PackerLog` should just open the log file in the _current_ buffer to make the modifiers to work correctly.
- The default behavior in the _packer_ window should be to open in a new tab.

lua/packer/display.lua
dhruvmanila: > Unless I misread the code, it currently only opens the log display if there is already a packer results window open, correct? 

Yes, only in the packer window.

> What I was proposing was that the command should always display the logs whether or not there's already a packer window.

Right, understood.

lua/packer/display.lua
dhruvmanila: If the user has not configured the buffer to be reloaded when the file changes outside of the editor, the logs will be static. (`:h autoread`, `:h checktime`) So, I thought it best to make the command available once packer has completed all the jobs.

dhruvmanila:
Drop the ball on this one, apologies for that. I'm not sure if I'll ever be able to pick it up incase it's needed but it seems that the repository has a notice that it's unmaintained. In that case, I'll close this PR.

Issue #418, opened 2021-06-23
[packer.nvim] [ERROR 08:52:09] async.lua:20: Error i...484: Can't open file C:UsersAdministratorAppDataLocal


yhnu:
- `nvim --version`:

NVIM v0.5.0-dev+1429-gb585f723b

- `git --version`:

git version 2.30.0.windows.2

- Operating system/version:

windows10

- Terminal name/version:

windows terminal

### Steps to reproduce
:PackerInstall

### Actual behaviour
[packer.nvim] [ERROR 08:52:09] async.lua:20: Error i...484: Can't open file C:UsersAdministratorAppDataLocal

### packer files

<details>
<summary>Plugin specification file(s)</summary>

Post or link your plugin specification files here, if you aren't able to provide a minimal
reproducer

</details>

<details>
<summary>packer log file</summary>

Post the contents of ~/.cache/nvim/packer.nvim.log here

</details>

<details>
<summary>packer compiled file</summary>

Post the contents of `packer_compiled.vim` here

</details>

wbthomason:
Presumed closed by #419.

Issue #412, opened 2021-06-17
Windows 10: Errors when running PackerSync


Nazeehe:
- `nvim --version`: NVIM v0.5.0-dev+1417-g6a77def1e
- `git --version`: git version 2.30.0.windows.1
- Operating system/version: Windows 10 21H1
- Terminal name/version: Windows Terminal Preview

Running PackerSync or Install and I sometimes will see:

[ERROR 6/17/2021 11:11:26 AM] ...\site\pack\packer\start\packer.nvim\lua\packer\async.lua:20: Error in coroutine: ...ack\packer\start\packer.nvim\lua\packer\plugin_utils.lua:169: Vim(source):E484: Can't open file C:Users

I added a log.info in the code to see what the "file" name actually is to see if it's coming in wrong, it's not:

[INFO  6/17/2021 11:11:26 AM] ...ack\packer\start\packer.nvim\lua\packer\plugin_utils.lua:168: file: C:\Users\nazee\AppData\Local\nvim-data\site\pack\packer\start\nvim-tree.lua\plugin\tree.vim

Any idea why the file name that's going in is being interpreted wrong by the vim.cmd statement:

            vim.cmd('silent exe "source ' .. file .. '"')

Thoughts?

yhnu:
#418 https://github.com/wbthomason/packer.nvim/pull/419 can slove it

wbthomason:
Presumed closed by #419.

PR #419, opened 2021-06-23
fix windows error #418

yhnu:
#412 #418

### DIFF
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -163,7 +163,7 @@ plugin_utils.load_plugin = function(plugin)
           error(files)
         end
       elseif #files > 0 then
-        for _, file in ipairs(files) do vim.cmd('silent exe "source ' .. file .. '"') end
+        for _, file in ipairs(files) do file = file:gsub("\\", "/") vim.cmd('silent exe "source ' .. file .. '"') end
       end
     end
   end
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -163,7 +163,10 @@ plugin_utils.load_plugin = function(plugin)
           error(files)
         end
       elseif #files > 0 then
-        for _, file in ipairs(files) do file = file:gsub("\\", "/") vim.cmd('silent exe "source ' .. file .. '"') end
+        for _, file in ipairs(files) do
+          file = file:gsub('\\', '/')
+          vim.cmd('silent exe "source ' .. file .. '"')
+        end
       end
     end
   end


Nazeehe (approved):

lua/packer/plugin_utils.lua
```diff
@@ -163,7 +163,7 @@ plugin_utils.load_plugin = function(plugin)
           error(files)
         end
       elseif #files > 0 then
-        for _, file in ipairs(files) do vim.cmd('silent exe "source ' .. file .. '"') end
+        for _, file in ipairs(files) do file = file:gsub("\\", "/") vim.cmd('silent exe "source ' .. file .. '"') end
```
wbthomason, line 166: Nitpick: if this `for` loop is no longer a single statement, then it should be split to multiple lines.

Also: I'm surprised that replacing with `/` works on Windows - is there somewhere in the Vim/Neovim help that indicates why this is OK?

wbthomason (changes requested):
Thanks, Windows errors are very difficult for me to find since I have no Windows system for testing! There's one style tweak I'd like to make, but otherwise this is probably good to go (though I admit I'm surprised that using forward slashes works as expected, and I'd like to document why this is needed).

Nazeehe:
> Thanks, Windows errors are very difficult for me to find since I have no Windows system for testing! There's one style tweak I'd like to make, but otherwise this is probably good to go (though I admit I'm surprised that using forward slashes works as expected, and I'd like to document why this is needed).

Yeah, I run on Windows so I can certainly help bring bugs back here. Forward slashes are acceptable for Windows APIs, that's why they work.

wbthomason:
All Neovim Windows APIs?

I'm ok to merge this once we fix the formatting nitpick I commented on in my review. Thanks again!

bugabinga:
> All Neovim Windows APIs?

Yes, but more specifically all win32-APIs.
On Windows `\` is interchangeable with `/` in paths.

https://stackoverflow.com/questions/38428561/difference-between-forward-slash-and-backslash-in-file-path?rq=1

I would not be surprised if the whole path separator handling in https://github.com/wbthomason/packer.nvim/blob/master/lua/packer/util.lua#L40 can simply be deleted.

Simply use forward slashes everywhere.

akinsho:
@wbthomason I wonder if regarding the comment above we could test out a PR that removes the path handling stuff for windows, get some windows users to test and see if it breaks anything. I guess though that there might be edge cases 🤔 or maybe packer compiled files that currently contain a backlash that might break(although if they are interchangeable they shouldn't)

wbthomason:
Thanks, this is good to know! I thought forward slashes only worked in limited contexts on Windows (e.g. in Powershell, but not in general).

Given that, I like @akinsho's idea to try simplifying the code. In the meanwhile, I'll apply the formatting change I still wanted here and merge this PR.

lua/packer/plugin_utils.lua
wbthomason: Applied the reformat myself in 53b7e1d.

wbthomason (approved):
Seems good to go, thanks!

PR #426, opened 2021-06-26
Call vnew without file arguments

leiserfg:
Calling it with a filename triggers the mechanisms of fileopening in nvim. For instance, if one is using
`lambdalisue/suda.vim` and calls PackerSync on `/`, suda asks for sudo password to open
`/[[packer]]`. As that file is not actually used, I guess is safe to avoid it.


akinsho:
Hi @leiserfg, you can essentially just override the command in your config to not apply a name to the file. I see your point, but this means that a nameless buffer opens instead, which I think is a slightly worse user experience for anyone new since the buffer is then just nameless, whereas something like vim plug does show its name. Although that's probably a pretty minor point. I wonder if this could be solved by opening it as a scratch buffer or nofile or something.

leiserfg:
Then what do you think about using:
`65vsplit |noswapfile hide enew|set buftype=nofile|file \\[packer\\]`
?

akinsho:
That looks kind of dauntingly long, the file is actually already set to `nofile` when I check inside the buffer. Tbh I think maybe some other people might have an opinion on this @nanotee, @wbthomason ? I personally think the case you describe is best solved locally rather than changing this behaviour, since the particular scenario seems uncommon.

nanotee:
To be fair I think having a little more control over buffer creation would be a good thing. In the very unlikely scenario where you have a file named `[packer]`, `:vnew` will open that instead of creating a new buffer.

Since `packer` already sets the `'filetype'` option in [setup_window()](https://github.com/nanotee/packer.nvim/blob/b3a1bb3b62d591ef81e0191ef6ad6eee428d861e/lua/packer/display.lua#L401), maybe adding `api.nvim_buf_set_name(disp.buf, '[packer]')` in that function would simplify things? I don't think users would mind losing the ability to chose the name of the buffer

wbthomason:
I'm fine with @nanotee's suggestion. I don't want to change the default command otherwise, as I agree with @akinsho that this is a niche issue with reasonable mechanisms already for working around, but I would prefer to skip file-opening mechanisms for the `packer` window and doubt that anyone will care that they can't customize the window name (if they really want to, they can do a workaround with an autocmd for `packer` filetype buffers).

wbthomason (approved):
This works well for me, thanks!

wbthomason:
@akinsho unless you're opposed to this fix, I'm good to merge.

akinsho:
@wbthomason this looks good to me 👍🏾

PR #429, opened 2021-06-28
Report line info for debugging incorrect plugin specifications

wbthomason:
This PR is 90% a single variable renaming in `manage`. The important bit is the addition of line info, i.e. the line of a user's plugin specification file at which `use` was called, for reporting errors in plugin specifications (e.g. duplicate aliases).


akinsho (approved):
@wbthomason this looks good to me seems straightforward enough 👍🏾

wbthomason:
Thanks for taking a look!

PR #434, opened 2021-07-01
Add config.git.default_url_format to allow customizing default URLs

### LINKED ISSUE #433: config to specify the default URL format.
### Describe the feature

In vim-plug, there is this nice little config [`g:plug_url_format`](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L776) so that users can specify the URL template used for cloning the repository. Due to certain reasons, I have to set `g:plug_url_format` to `'https://hub.fastgit.org/%s.git'`, which is faster than github.

It would be nice to have such a feature in packer so that when user use `aaa/bbb`, we would clone from `https://hub.fastgit.org/aaa/bbb.git`.

wbthomason:
Closes #433. Introduces a new config variable, `config.git.default_url_format` which is a format string used in `plugin_utils` if a plugin is specified in the `aaa/bbb` form rather than with a full local path or URL. Allows users to customize the default source/URL pattern for plugins, as described in #433.

@jdhao, can you take a look and make sure this does what you wanted?

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ default configuration values (and structure of the configuration table) are:
     },
     depth = 1, -- Git clone depth
     clone_timeout = 60, -- Timeout, in seconds, for git clones
+    default_url_format = 'https://github.com/%s' -- Lua format string used for "aaa/bbb" style plugins
   },
   display = {
     non_interactive = false, -- If true, disable display windows for all operations
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -212,6 +212,7 @@ default values: >
       },
       depth = 1, -- Git clone depth
       clone_timeout = 60, -- Timeout, in seconds, for git clones
+      default_url_format = 'https://github.com/%s' -- Lua format string used for "aaa/bbb" style plugins
     },
     display = {
       non_interactive = false, -- If true, disable display windows for all operations
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -39,7 +39,7 @@ local config_defaults = {
     },
     depth = 1,
     clone_timeout = 60,
-    default_url_format = 'https://github.com/%s'
+    default_url_format = 'https://github.com/%s.git'
   },
   display = {
     non_interactive = false,
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -38,7 +38,8 @@ local config_defaults = {
       revert = '-C %s reset --hard HEAD@{1}'
     },
     depth = 1,
-    clone_timeout = 60
+    clone_timeout = 60,
+    default_url_format = 'https://github.com/%s'
   },
   display = {
     non_interactive = false,
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -26,7 +26,7 @@ plugin_utils.guess_type = function(plugin)
     plugin.type = plugin_utils.git_plugin_type
   else
     local path = table.concat(vim.split(plugin.path, "\\", true), "/")
-    plugin.url = 'https://github.com/' .. path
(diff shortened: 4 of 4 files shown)


README.md
```diff
@@ -278,6 +278,7 @@ default configuration values (and structure of the configuration table) are:
     },
     depth = 1, -- Git clone depth
     clone_timeout = 60, -- Timeout, in seconds, for git clones
+    default_url_format = 'https://github.com/%s' -- Lua format string used for "aaa/bbb" style plugins
```
akinsho, line 281: This is the most minor of nits, but this looks more like changing the `default_url` rather than anything about it's format specifically i.e. is the name `default_url` more applicable. Tbh this is a 0.05 % valid nit so feel free to ignore (PS: I know plug does the same but it feels incorrect there as well, since ultimately what user changes when they change this is the url not the way it's structured 🤷🏾‍♂️ )

jdhao:
@wbthomason  Maybe the suffix `.git` is needed? I remember having met some errors before when I try to clone a repository if `.git` is missing.

README.md
wbthomason: Fair point - I named it this because, as currently used in `plugin_utils`, it does need to be a format string.

wbthomason:
@jdhao This default format worked on my system (and matches what we hard-coded before). Does it fail for you?

jdhao:
@wbthomason  Yes, so I always add `.git` to the repository name when git-cloning.  The default cloning link provided by GitHub also has `.git` suffix.

image

wbthomason:
Interesting. I have never needed that suffix, on any system (despite it being included in the Github-provided checkout URL). Might be a trait of your particular `git` config?

Regardless, it doesn't seem to break anything, so I'm ok including the suffix in the default format string.

jdhao:
I have previously experienced errors when trying to clone from  a private gitlab without `.git` suffix. So I guess it is safer to add `.git` suffix.

wbthomason:
Yep, added it in b04082e. Does this PR now do what you were looking for in #433?

jdhao:
I have tested the pull request and it basically works. What if the user uses `https://github.com/%s.git` initially, and then try to use `https://hub.fastgit.org/%s` later.

IN vim-plug, it will give an error since clearly the remote has changed. The user need to clean all the plugins and re-installl them. But it seems that packer still happily uses the old url format and does not know the url format change.

wbthomason:
`packer` only changes the remote if the *specification* URL changes (because, until now, that's how you would indicate that you wanted to use a different remote). In those cases, we automatically clean and reinstall the plugins. It probably wouldn't be that hard to use the same function for this case. @akinsho, as you wrote that feature, do you think it's a good idea to apply here?

wbthomason:
This would mean adapting `find_missing_plugins` to check against a plugin's URL in the remote update case. Seems straightforward enough?

akinsho:
@wbthomason that functionality works by checking that the git remote URL last 2 segments are different, I guess it could be adapted to check not just the author/project but the full URL. I'm assuming that in this case they would actually be different in case someone went from GitHub to GitLab or something.

IMO this seems like the correct behaviour i.e. adding this change, but I fear this PR will go from a very minor change to one that if there are unforeseen consequences could cause a bunch of issues like the original `find_missing_plugins` change 🤷🏾‍♂️ there are tests though 😆 although that didn't help the first time, it does seem like a small additional change though

jdhao:
Maybe the check for validity of a plugin can be put into a separate PR, where all existing issues related to checking if the plugin repo is valid can be addressed.

wbthomason:
Noting that this leaves a TODO of adding a more robust check for plugin repo correctness in a separate PR, I think this is OK to merge.

PR #303, opened 2021-04-18
refactor!: replace `git -C` with the `cwd` option

nanotee:
This PR replaces the `-C` argument in git commands with the `cwd` option for libuv jobs

I made these changes to address #255

This also partially addresses #165 and #287: I managed to reproduce the bug and while `:PackerInstall` still fails, updating plugins should now work properly. I plan on making a follow-up PR to try and fix the remaining edge cases.

This is a breaking change and could potentially introduce new bugs, so I'd appreciate feedback and testing.

Issue #49, opened 2020-09-09
vim.fn.executable is not reliable


TimUntersberger:
I don't really know why this it the case, but if you don't have `set shell=powershell` on windows, then vim thinks that `mklink` is not executable, even though it is.

I guess `mklink` is not an executable in `cmd`, but in `powershell` it is? I am really confused.

I'll try to find an alternative.

To be honest, I think you can assume that `mklink` exists on Windows right?

Update:

I now tested again with `set shell=powershell` and it still doesn't work. I can execute `mklink` using `:!mklink`, so I think something else is the problem.

wbthomason:
If you run `:lua vim.fn.executable('mklink')` as a manual command, does it output correctly?

TimUntersberger:
I tested using `:echo executable('mklink')`

wbthomason:
And this does not return the expected result?

TimUntersberger:
It returns `0`

wbthomason:
Hmm, ok. Might be an upstream bug?

I'd be pretty comfortable assuming `mklink` exists and then erroring out if that turns out to be incorrect. What do you think?

TimUntersberger:
I think that this would be ok to do 👍

PR #439, opened 2021-07-03
fix!: insert paths separately to avoid splitting

nanotee:
This PR:
- changes the default `git.install` command and how `git.setup()` handles URLs/paths for that command
- uses `vim.fn.delete()` to remove dirty plugins instead of `os.execute()` (as a bonus, this removes the last reference to `is_windows` outside of `util.lua`)

Together with #303, this should address #165 and #287.

The issue with the current approach is that `run_job()` [uses whitespace to split strings](https://github.com/wbthomason/packer.nvim/blob/b76bfba54031ad28f17e98ef555e99cf450d6cb3/lua/packer/jobs.lua#L173-L176), which means any path that contains whitespace will make the command fail.

This PR changes the API so that `git.install` can be split early before inserting the paths in the newly created table with `table.insert()`

### DIFF
--- a/lua/packer/clean.lua
+++ b/lua/packer/clean.lua
@@ -67,10 +67,9 @@ local clean_plugins = function(_, plugins, results)
       if await(display.ask_user('Removing the following directories. OK? (y/N)', lines)) then
         results.removals = dirty_plugins
         log.debug('Removed ' .. vim.inspect(dirty_plugins))
-        if util.is_windows then
-          for _, path in ipairs(dirty_plugins) do os.execute('cmd /C rmdir /S /Q ' .. path) end
-        else
-          os.execute('rm -rf ' .. table.concat(dirty_plugins, ' '))
+        for _, path in ipairs(dirty_plugins) do
+          local result = vim.fn.delete(path, 'rf')
+          if result == -1 then log.warn('Could not remove ' .. path) end
         end
       else
         log.warn('Cleaning cancelled!')
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -83,9 +83,10 @@ end
 git.setup = function(plugin)
   local plugin_name = util.get_plugin_full_name(plugin)
   local install_to = plugin.install_path
-  local install_cmd = config.exec_cmd
-                        .. fmt(config.subcommands.install, plugin.url, install_to,
-                               plugin.commit and 999999 or config.depth)
+  local install_cmd = vim.split(config.exec_cmd
+                        .. fmt(config.subcommands.install,
+                               plugin.commit and 999999 or config.depth), '%s+')
+
   local submodule_cmd = config.exec_cmd .. fmt(config.subcommands.submodules, install_to)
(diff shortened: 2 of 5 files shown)


lua/packer/clean.lua
```diff
@@ -67,10 +67,9 @@ local clean_plugins = function(_, plugins, results)
       if await(display.ask_user('Removing the following directories. OK? (y/N)', lines)) then
         results.removals = dirty_plugins
         log.debug('Removed ' .. vim.inspect(dirty_plugins))
-        if util.is_windows then
-          for _, path in ipairs(dirty_plugins) do os.execute('cmd /C rmdir /S /Q ' .. path) end
-        else
-          os.execute('rm -rf ' .. table.concat(dirty_plugins, ' '))
+        for _, path in ipairs(dirty_plugins) do
+          local result = vim.fn.delete(path, 'rf')
```
wbthomason, line 71: Is `vim.fn.delete` reliable on Windows? We've had trouble with other Vim functions on Windows in the past (e.g. https://github.com/wbthomason/packer.nvim/issues/49) and using `rmdir` was added by a Windows user in https://github.com/wbthomason/packer.nvim/commit/37a19658eeff5710a7ac074deb58f5e67476a675.

I'm fine switching to `vim.fn.delete` if it's known to work, but - as usual with Windows things - I do not personally know this.

lua/packer/plugin_types/git.lua
```diff
@@ -108,9 +109,13 @@ git.setup = function(plugin)
   end
 
   if plugin.branch or plugin.tag then
-    install_cmd = fmt('%s --branch %s', install_cmd, plugin.branch and plugin.branch or plugin.tag)
+    table.insert(install_cmd, '--branch')
```
wbthomason, line 112: Tiny nitpick: I'm gradually moving `packer` to the `tbl[#tbl + 1] = item` style of appending, since I've since been told it's the more "idiomatic"/slightly more performant (though not in a significant way for our use) approach.

wbthomason (approved):
As usual, this seems great and works on my end. I would appreciate if we could get some Windows testing, and I left two nitpick comments, but this seems good to merge to me otherwise.

wbthomason:
Added a commit fixing the merge conflicts my earlier PR merge introduced.

lua/packer/clean.lua
nanotee: I had a look at the implementation and it seems to be a somewhat thin wrapper around `libuv` functions. It looks [fairly well tested](https://github.com/neovim/neovim/blob/master/test/functional/legacy/delete_spec.lua), too.

I tried the function on Windows and it worked fine in most cases. The only time things got weird was when I kept a folder open in the explorer: `delete()` got rid of all the files in the test folder but couldn't remove the folder hierarchy. `rmdir` exhibits the same behavior, though, I think that's just a Windows limitation in general.

I'm happy to switch back to `rmdir` if you think it's safer.

wbthomason: If `rmdir` has the same unusual failure case, I'm ok with relying on `fn.delete` (given the test coverage, etc.). Thanks for testing that out!

Issue #445, opened 2021-07-04
Autoformat packer with Stylua


akinsho:
@wbthomason thought I'd raise an issue to discuss before doing anything. In most of my other lua projects, I've recently added autoformatting on push/commits to the main branch using [stylua](https://github.com/JohnnyMorganz/StyLua). Not sure if you've encountered it, but it's a quite actively maintained, opinionated lua code formatter. I've found that it works quite well producing quite readable code as well as being great in CI since it has a github code action, so it's quite straightforward to setup, and means you'd generally no longer really need to think about how code that is contributed is formatted.

I know you've historically used `luaformatter` which I looked at initially when looking at options, but I've found it to be a bit inactive, harder to setup in CI and probably most subjectively produces quite hard to read code I find what it does to function arguments really hard to follow personally.

Not sure if it means much, but it's being adopted by quite a lot of other plugins and projects in the nvim space, including I think a preliminary draft PR in neovim itself to maybe adopt it as well.

In either case I have set it up on several of my own projects so should be pretty straightforward to do here.

As a side note, I was initially hesitant to adopt it because it seemed strict about the `use {value}` syntax trying to encompass it in parens, but the author added in an option which allows single parameter functions to go without braces.

Anyway, let me know what you think, if you're open to it, I can raise a PR.

wbthomason:
Thanks! I actually was unaware of `stylua`; `lua-format` has served me well so far, so I hadn't looked for alternatives.

I've tried out `stylua` a bit now. I definitely miss some of the more concise formatting preferences of `lua-format`, and don't wholly agree with the particular "opinionated" take `stylua` offers.

That said, I'm far from the only developer on this project anymore, and it would be useful to have a formatter integrated in CI. I'm ok switching over to `stylua` for this benefit, even if I don't love its style.

I added a tentative `stylua` config file in cab2f56. If you'd be willing to raise a PR to get it integrated into CI, that would be much appreciated.

akinsho:
@wbthomason there's definitely choices it makes that I would do differently but since I personally hate to manually format my own code 😅 I spent a bit of time trying various options.

I think `stylua` whilst being opinionated in a way I don't always like is the most actively maintained (I hate the wasted space around the tables 😆), is the least offensive option IMO (subjective as that is).

[...]

PR #446, opened 2021-07-05
ci: add workflow for stylua formatting

akinsho:
@wbthomason this adds a github workflow for stylua.

A few things to note:
* It's currently set to format whenever anything is added to the `master` branch, I had t this way on my projects just to make sure nothing got missed, but it could alternatively be on PRs.
* The author of the formatting commit is the github bot, I've done it this way as I just saw a bunch of examples and copied them, I personally don't care much, but it could be changed
* I've ignored doc files like `.md`

I haven't actually formatted the repo as part of this change

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -1,0 +1,29 @@
+name: formatting
+on:
+  push:
+    paths-ignore:
+      - ".github/**"
+      - "*.md"
+    branches: ["master"]
+
+jobs:
+  stylua:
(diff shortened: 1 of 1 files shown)


akinsho:
fixes #445 

As a side note doing it on master will create `format lua` commits on master which I think is not ideal but since we cant rebase without breakage and adding commits to people WIP PRs will be super annoying it seems like the least problematic option

wbthomason:
Great, thank you for setting this up!

Issue #442, opened 2021-07-04
When lazy loading, packer gives error after running sync or compile [Not an editor command please check your config for correctness] 


rafamadriz:
- `nvim --version`: v0.5.0-dev
- `git --version`: 2.25.1
- Operating system/version: Linux mintdesktop 5.11.0-22-generic
- Terminal name/version: kitty 0.15.0

### Steps to reproduce
minimal init.lua to reproduce:
```lua
-- This file is a minimal `init.lua` example for debuggin plugins.
-- Open with `nvim -u debug.lua`

local fn = vim.fn
local install_path = "/tmp/nvim/site/pack/packer/start/packer.nvim"

vim.cmd [[set packpath=/tmp/nvim/site]]

local function load_plugins()
    local use = require("packer").use
    require("packer").startup(
        {
            function()
                use "wbthomason/packer.nvim"
                -- Put plugins to install here:
                use {"nvim-lua/plenary.nvim"}
                use {"nvim-lua/popup.nvim"}
                use {"nvim-telescope/telescope.nvim", cmd = "Telescope"}
		use {"kyazdani42/nvim-tree.lua", cmd = "NvimTreeToggle"}
            end,
	    
            config = {package_root = "/tmp/nvim/site/pack"}
        }
    )
end

_G.load_config = function()
    -- Put any configuration here:
end

if fn.isdirectory(install_path) == 0 then
    -- install packer
    fn.system({"git", "clone", "https://github.com/wbthomason/packer.nvim", install_path})
    load_plugins()
    require("packer").sync()
    vim.cmd "autocmd User PackerComplete ++once lua load_config()"
else
    load_plugins()
    _G.load_config()
end
```

Steps:
1. Save and open minimal init.lua example.
2. After the initial bootstrap, close and reopen nvim
3. Run Telescope or NvimTreeToggle (it works without issues)
4. Run PackerSync
5. Try to run Telescope or NvimTreeToggle again

### Actual behaviour

Gives `Error in packer_compiled: ...m/site/pack/packer/start/packer.nvim/lua/packer/load.lua:68: Vim:E492: Not an editor command: Telescope Please check your config for correctness`

I have tested this with Telescope and NvimTree. Here a video of the behavior.

https://user-images.githubusercontent.com/67771985/124370148-92efe380-dc63-11eb-91b8-eb130a64274b.mp4

### Expected behaviour

Being able to run all commands without errors after running `PackerSync` or `PackerCompile`

### packer files

</details>

[...]

Issue #408, opened 2021-06-15
Regex cmd for lazy loading


serhez:
Hello,

Could it be possible to have the _cmd_ argument to _use_ be a regex instead of a specific command? This would make a config much cleaner. Examples of plugins that would benefit from this are Telescope and DAP (DebugToggleBreakpoint, DebugContinue, DebugStepInto, DebugStepOver, DebugToggleRepl, DebugStart, etc.).

Thanks!

wbthomason:
This would probably be possible using the `CmdUndefined` event. We'd need to store a list of patterns to match against and add a single autocmd on that event.

I'm not adding new features until after #402 is done, but I'd be happy to help guide a PR adding this; it shouldn't be too involved.

tmccombs:
Even just having wildcards would be good enough for me. I'd be willing to make a PR if you point me in the right direction.

wbthomason:
Sure. This would be a change primarily in `lua/packer/compile.lua` and `lua/packer/load.lua`. The gist of it is that you'd need to change the command lazy-loading - which currently generates "fake" commands in `compile.lua` and deletes these fake commands before sourcing the plugin being loaded and running the resulting "real" command in `load.lua` - to instead generate `CmdUndefined` autocommands with the patterns given in the `cmd` key, and invoke the `load` function in the same way as we currently do when those autocommands trigger.

The one downside to this approach is that we will lose command autocompletion (i.e. when you start typing the command and press `<tab>`) by switching to this approach (see `:h CmdUndefined`).

wbthomason:
One benefit, however, is that we could work around issues like #442 and the issue @gegoune (if I recall correctly) encountered, where using a lazy-loaded command after recompiling breaks unexpectedly because running `packadd` again doesn't re-source runtime files. This isn't the only way to fix those issues, but it would be a pleasant side-effect.

PR #449, opened 2021-07-05
Check that a lazy-load command does not already exist before creating it

wbthomason:
This PR adds a check around each lazy-load command creation to only create the lazy-load command if a command with the same name does not already exist.

This should fix issues like https://github.com/kyazdani42/nvim-tree.lua/issues/471 and #442.

@gegoune: If you have a chance, could you please try this with the `nvim-tree` use case you described on Gitter?

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -671,7 +671,22 @@ packer.compile = function(raw_args)
   output_file:write(compiled_loader)
   output_file:close()
   if config.auto_reload_compiled then
+    local configs_to_run = {}
+    for plugin_name, plugin_info in pairs(_G.packer_plugins) do
+      if plugin_info.loaded and plugin_info.config and plugins[plugin_name].cmd then
+        configs_to_run[plugin_name] = plugin_info.config
+      end
+    end
+
     vim.cmd('source ' .. output_path)
+    for plugin_name, plugin_config in pairs(configs_to_run) do
+      for _, config_line in ipairs(plugin_config) do
+        local success, err = pcall(loadstring(config_line))
+        if not success then
+          vim.notify('Error running config for ' .. plugin_name .. ': ' .. vim.inspect(err), vim.log.levels.ERROR, {})
+        end
+      end
+    end
   end
   log.info 'Finished compiling lazy-loaders!'
   packer.on_compile_done()
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -31,7 +31,6 @@ vim.api.nvim_command('packadd packer.nvim')
 local no_errors = pcall(function()
 ]]
 
-
 local catch_errors = [[
 catch
   echohl ErrorMsg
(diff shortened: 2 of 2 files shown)


wbthomason:
Also note that this feature will not work with #408 as planned.

wbthomason:
Also @rafamadriz, if you have a chance to see if this fixes #442, that would be helpful too!

TheBlob42:
I tested this branch in regards to #442 as I just experienced the same error as @rafamadriz and for me it fixes the issue :slightly_smiling_face:

wbthomason:
> Also note that this feature will not work with #408 as planned (as far as I know, there's no way to use `exists` with a pattern?).

I'm dumb - this is fine with #408, just redundant. If the command already exists, then `CmdUndefined` wouldn't fire, so we'd be fine.

gegoune:
@wbthomason It does seem to work now, nicely done! Have tested with Neogit rather than nvim-tree.

I am not sure if that is the result of this PR but after moving configuration fo neogit from `after/plugin` to `config` function in packer nvim has to be restarted for neogit's configuration changed to take effect.

Also, on unrelated note, I was thinking if it is possible to guard configurations in `after/plugin/*` in case given package was not loaded. Do you know of a way?

wbthomason:
Good point. This will prevent `load` from running the new config after an update...

We could hack around that by, in the reload logic in `packer.lua`, checking if a plugin is currently loaded and running its `setup` and `config` fields if it is.

re: `after/...`: I'm not 100% sure I know what you mean by guarding the configurations in case a given package was not loaded. You can use the `_G.packer_plugins` table to check if a given plugin is currently loaded (there's a boolean field `loaded`); this is how `packer.status` checks plugin load state.

rafamadriz:
> Also @rafamadriz, if you have a chance to see if this fixes #442, that would be helpful too!

@wbthomason Awesome, that is working for me!!! May I ask when you think this will get merged ?

wbthomason:
@rafamadriz I've just now pushed a tentative fix to the issue @gegoune noticed. If that works (I've tested it, but not thoroughly yet), then I think this is good to merge.

wbthomason:
@rafamadriz @gegoune Any chance you've been able to test this branch since the latest commit?

gegoune:
I have since removed any lazy loading. Could test it later today though.

rafamadriz:
@wbthomason Yes I was running this branch for a couple of days, didn't have any problem.

wbthomason:
@gegoune It should be ok, but thanks for the offer!
@rafamadriz Great, thanks! I think, from some additional testing of my own, that this branch is still not running the changed configs correctly, but at least it isn't breaking things further and does fix the problem of commands being deleted. Probably good to merge today.

ram02z:
Would like to also add that I have been running this branch for a week and I have no issues either.

wbthomason:
I'm going to merge this for now - while it doesn't 100% solve the problem (updated configs are still not being sourced correctly), it does fix the main breakage of this issue. I'll try to get to the remaining work in another PR soon, unless someone beats me to it.

PR #453, opened 2021-07-06
fix(git): put the `cwd` option in the right table

nanotee:
I introduced a regression in #303, sorry about this!


akinsho:
This seems to have resolved the issue raised in #454 so think we should merge it sooner rather than later before more reports come in. Will give @wbthomason an hour or so to feedback in case he has any thoughts otherwise will merge to unblock people.

wbthomason:
Thanks for the quick fix. I did test installing plugins before merging the PR that introduced this regression, so I'm curious why I didn't see this error. Maybe something got jumbled while resolving merge conflicts?

PR #458, opened 2021-07-08
fix(compile): blank `packer_compile` error messages

Iron-E:
After the recent PR, error messages occurring in the `packer_compiled.lua` file (such as in configs, or plugins) were left blank. This was my bad— `pcall()` raises its own errors, and they don't interfere with `v:exception`.


wbthomason (approved):
Ah, thanks for catching this. That's my bad too - I also thought/assumed `v:exception` would contain Lua errors.

PR #470, opened 2021-07-12
Use CmdUndefined autocommand for lazy loading on commands.

### LINKED ISSUE #408: Regex cmd for lazy loading

tmccombs:
Fixes #408

The downside to this is that command autocompletion no longer works.

Potentially the old way could be a seperate option?

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ use {
   -- The setup key implies opt = true
   setup = string or function,  -- Specifies code to run before this plugin is loaded.
   -- The following keys all imply lazy-loading and imply opt = true
-  cmd = string or list,        -- Specifies commands which load this plugin.
+  cmd = string or list,        -- Specifies commands which load this plugin. Can have patterns.
   ft = string or list,         -- Specifies filetypes which load this plugin.
   keys = string or list,       -- Specifies maps which load this plugin. See "Keybindings".
   event = string or list,      -- Specifies autocommand events which load this plugin.
@@ -382,6 +382,10 @@ use {
   requiring a string which matches one of these patterns, the plugin will be loaded.
 }
 ```
+
+For the `cmd` option, the command may be a full command, or an autocommand pattern. If the command contains any
+non-alphanumeric characters, it is assumed to be a pattern, and instead of creating a stub command, it creates
+a CmdUndefined autocmd to load the plugin when a command that matches the pattern is invoked.
 
 #### Checking plugin statuses
 You can check whether or not a particular plugin is installed with `packer` as well as if that plugin is loaded.
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ use {
   -- The setup key implies opt = true
   setup = string or function,  -- Specifies code to run before this plugin is loaded.
   -- The following keys all imply lazy-loading and imply opt = true
-  cmd = string or list,        -- Specifies commands which load this plugin. Can have patterns.
+  cmd = string or list,        -- Specifies commands which load this plugin. Can be an autocmd pattern.
   ft = string or list,         -- Specifies filetypes which load this plugin.
   keys = string or list,       -- Specifies maps which load this plugin. See "Keybindings".
   event = string or list,      -- Specifies autocommand events which load this plugin.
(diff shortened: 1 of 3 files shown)


akinsho:
@tmccombs thanks for raising the PR 👍🏾 tbh I think another option will just add more complexity and confusion, regarding how packer handles command lazy loading. I think it should be one thing or the other and taking the tradeoff i.e. if fuzzy matching commands is more valuable then we accept losing completion etc.

That being said, 😆 I wonder if there's a way to indicate that we want one type of behaviour versus another which is kind of what we do with autocommands e.g. if a user specifies `event = 'BufRead'` then packer under the hood does `BufRead *` which means a user can still actually do `BufRead *.md` or something.

I wonder if a similar thing could be done here e.g. if a user specifies something that is a pattern e.g. `NvimTree*` then we check for the pattern character and use `CmdUndefined` autocommand with the pattern as the argument or if they write out whole command e.g. `NvimTreeToggle` then we do what packer always used to do i.e. create the fake command.

I haven't tested if using the pattern is possible but since it's just an `autocmd-pattern` it should be able to work like this maybe?

wbthomason:
I like @akinsho's idea of choosing our lazy-load strategy for commands conditionally based on the pattern provided. We'd need to carefully document how this logic works, though.

tmccombs:
Hmm, I think we could possibly have something where we used the old behavior if the command is entirely alphanumeric, and treat it as a pattern if it has special characters.

akinsho:
That makes sense since special characters aren't allowed in a command name so it's safe to presume that the intention would be for it to be a pattern in that case. We'd need to add it to the docs as clearly as possible to avoid future questions on this behaviour as @.wbthomason mentioned

doc/packer.txt
```diff
@@ -563,4 +563,8 @@ invoked as follows:
   }
 - With a list of plugins specified in either of the above two forms
 
+For the *cmd* option, the command may be a pattern instead of an explicit pattern. If the command contains any
```
akinsho, line 566: I think this is a bit confusing, i.e. `the command maybe a pattern instead of an explicit pattern`. I think maybe saying that it could be the full command or a lua pattern is clear enough 🤷🏿‍♂️

akinsho: Should have clarified the bit about the alphanumeric characters I think is fine just the pattern/explicit pattern comment

akinsho:
@tmccombs sorry I completely forgot about this, is/was it ready for review +/- merge (there's a conflict now but otherwise is this ready?). Haven't tested it out yet though.

tmccombs:
I've fixed the merge conflict, and have done some minimal testing, and afaict it works as expected.

doc/packer.txt
```diff
@@ -555,7 +555,7 @@ invoked as follows:
     config = string or function, -- Specifies code to run after this plugin is loaded.
     rocks = string or list,      -- Specifies Luarocks dependencies for the plugin
     -- The following keys all imply lazy-loading
-    cmd = string or list,        -- Specifies commands which load this plugin.
+    cmd = string or list,        -- Specifies commands which load this plugin.  Can be autocmd pattern.
```
akinsho, line 558:
```suggestion
    cmd = string or list,        -- Specifies commands which load this plugin.  Can be an autocmd pattern.
```

README.md
```diff
@@ -370,7 +370,7 @@ use {
   -- The setup key implies opt = true
   setup = string or function,  -- Specifies code to run before this plugin is loaded.
   -- The following keys all imply lazy-loading and imply opt = true
-  cmd = string or list,        -- Specifies commands which load this plugin.
+  cmd = string or list,        -- Specifies commands which load this plugin. Can have patterns.
```
akinsho, line 373: I think this wording should be the same as the one in `packer.txt` as that is clearer

akinsho (commented):
I noticed a couple of minor documentation things which I've raised, tiny stuff around wording. Sorry I didn't pick this up earlier

akinsho (approved):
Barring the documentation issues I raised, I tested this out and this works well (as I far as I could see) so looks good to me. Happy to merge this after those fixes (granted @wbthomason hasn't got any comments)

akinsho:
Going to go ahead and merge this now @wbthomason 🤞🏿, @tmccombs thanks for your work on this.

Issue #258, opened 2021-03-18
Plugin loaded as a dependency before 'use' gives warning about Plugin being used twice


muchosteechos:
I get the warning `[packer.nvim] [WARN 11:53:25] pack... "nvim-web-devicons" is used twice!` with my configuration on startup. 

I'm kind of aware of this due to the fact that I have this plugin as a dependency and use it after that anyways:
```
return require('packer').startup(function()
  use {'wbthomason/packer.nvim', opt = true}

  use {
    'kyazdani42/nvim-tree.lua',
    requires = {'kyazdani42/nvim-web-devicons'},
  }

  use {'kyazdani42/nvim-web-devicons'}
end)
```

This is intended. I want to specify `nvim-web-devicons` as a dependency of `nvim-tree`, but load it regardless of wether I use `nvim-tree` or not. 

Is there a way to not get this warning at startup? I can only think of inelegant solutions, that either require too much manual work or don't specify the intended behavior I want
- put `nvim-tree` down in the configuration (too much manual work if I manage more plugins in this way),
- something like `cond=--[[ check if nvim-web-devicons loaded --]]` in `nvim-web-devicons` use
- use of `opt`
- suppress warnings in general

I'm also unaware of the implications of this warning. Does `packer` actually load the plugin twice? Can this setup lead to significantly worse startup time?

akinsho:
@muchosteechos I think this might be some confusion in how the dependency specification in packer works. Specifying a plugin as `requires` means that when nvim-tree is installed it will also nvim web devicons, and that if nvim tree is removed it should remove web devicons. If you then go on to specify the plugin in its own use block there is no point having it inside the requires of nvim tree.

akinsho:
`requires` should only really be used if the only reason a plugin is specified is because it is needed by another and should only be there if the other one is there. If you want a plugin to be around regardless of the other ones status then it is not really a dependency anymore

wbthomason:
@akinsho is correct. That said, if you want to specify a plugin as a dependency for multiple plugins (or have it in its own block and also as a dependency), just give its name (not the full path) as the dependency. In your case, this would be `nvim-web-devicons` instead of `kyazdani42/nvim-web-devicons`.

muchosteechos:
Thank you both! This explains things. 

[...]

PR #471, opened 2021-07-12
Allow top level and required plugin usages

akinsho:
This fixes an issue where if a top level spec for a plugin was called
after a required spec for that, a duplicate error was shown.
e.g.
```lua
use {'first/plugin', requires = {'another/plugin'}}
use 'another/plugin' -- <- This would throw a duplicate plugin error
```

This PR instead adds a flag to ignore the duplicate warning if the initial
plugin spec was from a `require`.

This is based off of https://github.com/wbthomason/packer.nvim/issues/258#issuecomment-876568439.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -187,7 +187,7 @@ manage = function(plugin_data)
     return
   end
 
-  if plugins[name] then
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Works like a charm, thank you!

PR #473, opened 2021-07-13
feat(compile): support lua files in after directory

kevinhwang91:
nvim-0.5 has supported sourcing lua files, including `so *.lua` cmd.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -212,8 +212,8 @@ local function detect_after_plugin(name, plugin_path)
 end
 
 local ftdetect_patterns = {
-  table.concat({ 'ftdetect', '**/*.vim' }, util.get_separator()),
-  table.concat({ 'after', 'ftdetect', '**/*.vim' }, util.get_separator()),
+  table.concat({ 'ftdetect', [[**/*.\(vim\|lua\)]] }, util.get_separator()),
+  table.concat({ 'after', 'ftdetect', [[**/*.\(vim\|lua\)]] }, util.get_separator()),
 }
 local function detect_ftdetect(name, plugin_path)
   local paths = {
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -193,7 +193,7 @@ local function make_try_loadstring(item, chunk, name)
   return executable_string, bytecode
 end
 
-local after_plugin_pattern = table.concat({ 'after', 'plugin', '**/*.vim' }, util.get_separator())
+local after_plugin_pattern = table.concat({ 'after', 'plugin', [[**/*.\(vim\|lua\)]] }, util.get_separator())
 local function detect_after_plugin(name, plugin_path)
   local path = plugin_path .. util.get_separator() .. after_plugin_pattern
   local glob_ok, files = pcall(vim.fn.glob, path, false, true)


akinsho:
@kevinhwang91 thanks for the PR 👍🏾 would you be able to add the same treatment to the `ftdetect` patterns just below at line 215, it's a very similar pattern as what you've already changed.

https://github.com/wbthomason/packer.nvim/blob/12207c4d09c7620933a82fb8105eb5047d4a331e/lua/packer/compile.lua#L214-L217

kevinhwang91:
done:)

akinsho:
@kevinhwang91 are there any plugins using `ftdetect` or `after` I can check with although a plugin using after is a real antipattern anyway 😅 sad we have to support it at all

kevinhwang91:
nvim-bqf use `after` directory. It expects to change/override the qf window options such as `nu`, `rnu`, `wrap`, etc.
Users can use their `~/.config/nvim/after` directory which is at the end of `rtp` to override the option changed by nvim-bqf.

vim-markdown.vim use `ftdetect` to set the filetype to enable the scripts located under `ftplugin` directory.
https://github.com/plasticboy/vim-markdown/blob/8e5d86f7b85234d3d1b4207dceebc43a768ed5d4/ftdetect/markdown.vim#L7-L8

akinsho:
@kevinhwang91 does the after in `bqf` use lua i.e. would be a good one to test this PR on since the vim markdown one is in vim so would have worked anyway.

Also OOI (this doesn't relate to this PR) why use the `after` directory instead of the `ftplugin` directory for your plugin? if you use after then if a user has added their own preferences in `nvim/ftplugin` then you will override their settings unless they then have to now use `nvim/after/ftplugin` to then override your own?

kevinhwang91:
> @kevinhwang91 does the after in `bqf` use lua i.e. would be a good one to test this PR on since the vim markdown one is in vim so would have worked anyway.
> 

If packer.nvim doesn't support lua files under `after` directory, I think the lua plugin developer will use `.vim` instead of `.lua`.
That's why I create this PR.

> Also OOI (this doesn't relate to this PR) why use the `after` directory instead of the `ftplugin` directory for your plugin? if you use after then if a user has added their own preferences in `nvim/ftplugin` then you will override their settings unless they then have to now use `nvim/after/ftplugin` to then override your own?

The reason is I want bqf to be `KISS`. I just expect the user to give bqf a try without any configurations and found bqf is neat and powerful. To override options for qf (set by the user or other qf plugins) is the simple way bqf can do. If the user is not satisfied with the default design, `~/.config/nvim/after` is the backdoor if he/she loves bqf but likes customizing its qf style.

> FYI, magic window feature wants the final window height, other qf plugins may change the window height during initialization.

akinsho:
Thanks for explaining re. after, I was just curious why is all. I'll try using this PR with bqf to verify it behaves as expected later on today unless someone else beats me to it 👍🏿

wbthomason:
@akinsho Did you ever have a chance to verify this PR with bqf? I can if you've not yet.

akinsho:
@wbthomason nope never got round to it, pretty short on time over the next couple of weeks.

wbthomason:
No worries (and same re: time); I'll give it a go and then (assuming no weird behavior) go ahead and merge this sometime in the next few days.

akinsho:
@wbthomason @kevinhwang91 I *finally* got round to testing this PR and seems to work fine to me so will merge it 👍🏿 hoping nothing explodes 😅

PR #492, opened 2021-07-17
Fix #491: README dotfile dead links

### LINKED ISSUE #491: README Usage->dotfiles dead links
Just a simple issue to tell there's two dead links in the [Usage](https://github.com/wbthomason/packer.nvim#usage) section. Two examples given in **My dotfiles** end on a 404. 

These are :

- [Loading file](https://github.com/wbthomason/dotfiles/blob/linux/neovim/.config/nvim/plugin/plugins.vim)
- [Generated lazy-loader file](https://github.com/wbthomason/dotfiles/blob/linux/neovim/.config/nvim/plugin/packer_load.vim)

I couldn't find the second one in your dotfiles but the first seems to be [here](https://github.com/wbthomason/dotfiles/blob/linux/neovim/.config/nvim/lua/plugins.lua) now. If I'm correct for the one I found I can make a PR if you want. For the second, if you tell me which file it is I can add it to the PR. 

Not a great issue but I thought fixing this can be quickly done and if we can improve the README a bit in a mater of minutes, why not ! Anyway, hope it will be useful.

n3wborn:


wbthomason (approved):
LGTM, thanks again!

PR #507, opened 2021-07-26
Set non_interactive = true if neovim is headless

### LINKED ISSUE #503: Default to non_interactive mode when neovim is headless
I'm trying to use packer headless. One problem is that with the current autocommand workaround for non-blocking insatlls, packer prompts if it detects any non-tracked packages in package path. Since we can't feedkeys to silence this with headless, the installation hangs.

I would suggest defaulting to non_interactive mode is neovim has no ui attached which can be detected with #vim.api.nvim_list_uis() == 0

image

wbthomason:
Fixes #503 by doing what the title says.

@mjlbach: Does this do what you wanted for #503?

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -109,6 +109,10 @@ packer.init = function(user_config)
   config.pack_dir = join_paths(config.package_root, config.plugin_package)
   config.opt_dir = join_paths(config.pack_dir, 'opt')
   config.start_dir = join_paths(config.pack_dir, 'start')
+  if #vim.api.nvim_list_uis() == 0 then
+    config.display.non_interactive = true
+  end
+
   local plugin_utils = require_and_configure 'plugin_utils'
   plugin_utils.ensure_dirs()
   if not config.disable_commands then


mjlbach:
Yes :)

khjorth:
This breaks interactive mode for people who connect a UI to a headless Neovim instance after it has started, and it's very difficult to figure out why: If you run e.g. `:PackerSync` there is no indication that the command had any effect. It's counter-intuitive that this is caused by non-interactive mode being automatically enabled, even if `non_interactive` is explicitly set to false during init. This is exacerbated by the `false` default value for `non_interactive`, which the user could reasonably expect to mean that non-interactive mode is off by default.

It seems that the only workaround at the moment is to explicitly set `non_interactive` to `false` and reload the packer config *after* the UI is connected.

Could we please have this reverted, or at least have the `#vim.api.nvim_list_uis() == 0` check moved to the initialization of `non_interactive`?

joshuarubin:
I'd like to second @khjorth's request. I run a headless vim session on a remote dev machine. When I connect to it with a local ui over remote tcp, this change makes all packer windows invisible and very hard to reason about.

PR #428, opened 2021-06-28
Refactor: faster operation startup

wbthomason:
This refactoring primarily (1) refactors the main management operations to only read the state of the filesystem (e.g. look for installed/missing plugins) **once** and (2) adds finer-grained debug logging.

These changes are in response to the unusually high "time to display" that @cloggier mentioned on Gitter; for me, this version reduces the average time to display from 0.3s to 0.15s.

This PR also contains some minor renaming/reformatting.

PR #508, opened 2021-07-26
Feature: check that a plugin is actually installed before running its config

wbthomason:
This addresses the concern raised by @florentc in #506 by asynchronously checking that a plugin's install path actually exists before we run its `config` function. Note that we still run `setup` functions as usual because they are designed to run before a plugin is loaded anyway, so they should be fine to run before a plugin is installed. Making this choice helps us avoid needing to think about some race conditions which could otherwise arise.

@florentc: Can you please confirm that this behaves as expected? @akinsho if you have a chance to look at this, that would also be great, but nw if you're busy - it's a fairly simple change.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -164,6 +164,29 @@ local function timed_chunk(chunk, name, output_table)
   return output_table
 end
 
+local function timed_install_check_chunk(chunk, name, path, output_table)
+  output_table = output_table or {}
+  output_table[#output_table + 1] = 'time([[' .. name .. ']], true)'
+  output_table[#output_table + 1] = 'vim.loop.fs_lstat([[' .. path .. ']], function(err, stat)'
+  output_table[#output_table + 1] = '\tif err == nil and stat ~= nil then'
+  output_table[#output_table + 1] = '\t\tvim.schedule(function()'
+  if type(chunk) == 'string' then
+    output_table[#output_table + 1] = '\t\t\t' .. chunk
+  else
+    vim.list_extend(output_table, chunk)
+  end
+
+  output_table[#output_table + 1] = '\t\t\tend)'
+  output_table[#output_table + 1] = '\telse'
+  output_table[#output_table + 1] = '\t\tvim.schedule(function()vim.api.nvim_notify("packer.nvim: Skipping '
+    .. name
+    .. ' because plugin is not installed!", vim.log.levels.WARN, {}) end)'
+  output_table[#output_table + 1] = '\tend'
+  output_table[#output_table + 1] = '\ttime([[' .. name .. ']], false)'
+  output_table[#output_table + 1] = 'end)'
+  return output_table
+end
+
 -- We special-case timing for conditionals because we can't directly wrap the condition test
 local function timed_condition(condition, body, name, output_table)
   output_table = output_table or {}
@@ -496,7 +519,7 @@ local function make_loaders(_, plugins, output_lua, should_profile)
   local config_lines = {}
   for name, plugin_config in pairs(configs) do
     local lines = { '-- Config for: ' .. name }
-    timed_chunk(plugin_config, 'Config for ' .. name, lines)
+    timed_install_check_chunk(plugin_config, 'Config for ' .. name, loaders[name].path, lines)
     vim.list_extend(config_lines, lines)
   end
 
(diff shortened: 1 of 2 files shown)


shadmansaleh:
With this I get a flicker on startup on lualine .
It seems because the plugin check is async first redraw happens before the plugin is loaded and when it loads it changes the statusline hence the flicker . Also the screen seems to load faster which suggests first redraw before plugin load too . I guess more or less all start plugins will have similar effect only this one has a  visual que .

wbthomason:
:+1: to the style comment re: using a `fmt` block.

I made the check async because fs operations can be noticeably slow (e.g. when I was working on https://github.com/wbthomason/packer.nvim/pull/428). The flicker is an issue, though...I'm not sure what a good solution to that might be (unless doing sync fs checks is just fast enough).

re: `opt` plugins: Good point; I'll add logic to `load` to check installation before running any commands also.

shadmansaleh:
> I made the check async because fs operations can be noticeably slow

I'm aware of that . I was just pointing out it can have side affects

> The flicker is an issue, though...I'm not sure what a good solution to that might be

I'm not sure either .

wbthomason:
Fixed style and `opt` comments!

MuhammedZakir:
Can a flag be added (e.g. `needed_at_start`) for that? This way, packer can do:
```lua
if flag then
  -- sync
else
  -- async
end
```

wbthomason:
@MuhammedZakir Yeah, this is possible...I'm just hesitant to add yet more flags, increasing complication. This might still be the best feasible solution, though.

lua/packer/compile.lua
```diff
@@ -164,19 +164,43 @@ local function timed_chunk(chunk, name, output_table)
   return output_table
 end
 
+local timed_install_check_fmt = [=[time([[%s]], true)
+vim.loop.fs_lstat([[%s]], function(err, stat)
+  if err == nil and stat ~= nil then
+    vim.schedule(function()
```
akinsho, line 170: @wbthomason is it worth checking how fast the synchronous version of this is before using the async version to avoid the flickering issues and other bugs that are likely to crop up with this running asynchronously. I think `fs_stat` is relatively quick, alternatively maybe read the contents of the start and opt directory once and get a list of directories and just check the plugin name is in a list of dirs within either dir. That way, the existence check only happens once by just reusing the list of directories that were in `start/opt` dirs?

wbthomason: > is it worth checking how fast the synchronous version of this is before using the async version to avoid the flickering issues and other bugs that are likely to crop up with this running asynchronously

Probably, yeah. I was basing the decision to do async from the start on the cost of scanning for installed plugins that I saw in #428, but if async causes flickering on every startup, it's worth testing the sync alternative.

> alternatively maybe read the contents of the start and opt directory once and get a list of directories and just check the plugin name is in a list of dirs within either dir

This is more expensive than I'd like for something that runs at every start, ideally. That's why e.g. we defer checking for installed plugins until a plugin management operation happens already.

wbthomason:
Ok, I'd love to merge this and still don't have a better solution to the flicker problem than @MuhammedZakir's suggestion to allow plugins like `lualine` to be flagged to skip the installation check. Would that be acceptable to those affected?

wbthomason:
Alternatively: we already `pcall` config functions; we could trigger the installation check only if that `pcall` fails, to report a better cause for the failed config call.

wbthomason:
@akinsho: Yeah, I think anything that needs to be loaded immediately will be affected. That's kind of why I like what I proposed second better, where we only do the "is this plugin installed" check *after* we know it might be relevant, i.e. the user-provided config fails to run as expected. That should avoid the "blink" described with lualine while still giving users protection from configs that fail in a confusing way because of missing plugins (assuming that `pcall` prevents any bad behavior from a config from "leaking", which isn't necessarily true...)

akinsho:
@wbthomason the `pcall` solution sounds like a better alternative 👍🏿 although from the original PR description is it as valuable in this form since it will now only run after the config, you mentioned

> Making this choice helps us avoid needing to think about some race conditions which could otherwise arise.

I'm wondering if this becomes primarily a way of warning users about plugins not being installed (which is still valuable)

wbthomason:
@akinsho Right, the `pcall` form is not quite as useful as the original form of this PR. If a plugin's config has undesirable side-effects (e.g. setting global state, printing error messages, etc.) then I don't believe `pcall` will do anything for us other than letting us notify the user that they have an plugin which needs to be installed. Still less confusing than just showing the error messages, but no more "damage control".

PR #509, opened 2021-07-26
Add the module_pattern key

### LINKED ISSUE #478: LazyLoading by base module path
Great project. Thanks for all the work you have put into it.

### Describe the feature

The following telescope configuration fails
```lua
	use({
		"nvim-telescope/telescope.nvim",
		requires = {
			{ "nvim-lua/popup.nvim", module = "popup" },
			{ "nvim-lua/plenary.nvim", module = "plenary" },
		},
		module = "telescope.builtin",
		config = function()
			require("plugins/telescope").config()
		end,
		setup = function()
			require("plugins/telescope").mappings()
		end,
	})
```
This is because I have `plenary` lazyload using "module" but the telescope does not require plenary directly instead requires submodules like `plenary.job`, `plenary.log`, `plenary.reload` etc. 

Is there some way that I can lazyload `nvim-lua/plenary.nvim` whenever any of `plenary.*` is required?

wbthomason:
Closes #478 by adding a new key, `module_pattern`, which works exactly the same as `module` but does not run `pesc` or prepend a caret to the pattern.

@thezealousfool: Can you please confirm that this does what you wanted in #478?
@akinsho: If you have a chance to try this out, that would be great, but as always nw if you don't have time, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,6 +365,8 @@ use {
   cond = string, function, or list of strings/functions,   -- Specifies a conditional test to load this plugin
   module = string or list      -- Specifies Lua module names for require. When requiring a string which starts
                                -- with one of these module names, the plugin will be loaded.
+  module_pattern = string/list -- Specifies Lua pattern of Lua module names for require. When
+  requiring a string which matches one of these patterns, the plugin will be loaded.
 }
 ```
 
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -562,9 +562,8 @@ invoked as follows:
     setup = string or function,  -- Specifies code to run before this plugin is loaded.
     module = string or list      -- Specifies Lua module names for require. When requiring a string which starts
                                  -- with one of these module names, the plugin will be loaded.
-  module_pattern = string/list   -- Specifies Lua pattern of Lua module names for
-  require. When requiring a string which matches one of these patterns, the
-  plugin will be loaded.
+    module_pattern = string/list -- Specifies Lua pattern of Lua module names for require. When requiring a string
+                                 -- which matches one of these patterns, the plugin will be loaded.
   }
 - With a list of plugins specified in either of the above two forms
 
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -562,6 +562,9 @@ invoked as follows:
     setup = string or function,  -- Specifies code to run before this plugin is loaded.
     module = string or list      -- Specifies Lua module names for require. When requiring a string which starts
                                  -- with one of these module names, the plugin will be loaded.
+  module_pattern = string/list   -- Specifies Lua pattern of Lua module names for
+  require. When requiring a string which matches one of these patterns, the
+  plugin will be loaded.
   }
 - With a list of plugins specified in either of the above two forms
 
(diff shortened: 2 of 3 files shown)


doc/packer.txt
```diff
@@ -562,6 +562,9 @@ invoked as follows:
     setup = string or function,  -- Specifies code to run before this plugin is loaded.
     module = string or list      -- Specifies Lua module names for require. When requiring a string which starts
                                  -- with one of these module names, the plugin will be loaded.
+  module_pattern = string/list   -- Specifies Lua pattern of Lua module names for
```
akinsho, line 565: Indentation seems a little off here. Obv very minor point 😆

lua/packer/compile.lua
```diff
@@ -439,14 +439,26 @@ local function make_loaders(_, plugins, output_lua, should_profile)
...
+
+        if plugin.module then
+          if type(plugin.module) == 'string' then
+            plugin.module = { plugin.module }
+          end
+
+          for _, module_name in ipairs(plugin.module) do
+            module_lazy_loads['^' .. vim.pesc(module_name)] = name
+          end
+        else
```
akinsho, line 454: Another very minor point but this doesn't seem to explicitly also verify that `plugin.module_pattern` exists before operating on it, it only seems to check if module exists, and otherwise it always runs this branch? Seems like the type conditional won't break, and I'm assuming the `ipairs` can handle a nil value, but still seems a little dangerous. Should this not be `elseif plugin.module_pattern`

doc/packer.txt
wbthomason: Whoops, nice catch, thanks.

lua/packer/compile.lua
wbthomason: Yep, good point.

akinsho: Makes sense, missed the parent conditional 👍🏿

wbthomason:
@thezealousfool: Any chance you've had the opportunity to confirm that this closes #478 for you?

README.md
```diff
@@ -365,6 +365,8 @@ use {
   cond = string, function, or list of strings/functions,   -- Specifies a conditional test to load this plugin
   module = string or list      -- Specifies Lua module names for require. When requiring a string which starts
                                -- with one of these module names, the plugin will be loaded.
+  module_pattern = string/list -- Specifies Lua pattern of Lua module names for require. When
+  requiring a string which matches one of these patterns, the plugin will be loaded.
```
thezealousfool, line 369: Apart from the proper formatting, I would also request you to add an example to the README as to how this parameter works

thezealousfool:
@wbthomason I am not sure if I am doing something wrong but I have these -
```lua
	use({ "nvim-lua/popup.nvim", module_pattern = "popup.*" })
	use({ "nvim-lua/plenary.nvim", module_pattern = "plenary.*" })
```
I had these two packages in the `start` directory but adding the `module_pattern` and running `PackerUpdate` does not attempt to move the packages to the `opt` directory

wbthomason:
@thezealousfool Whoops, nice catch. I didn't notice this because my test plugins were already `opt`, but I'd forgotten to mark `module_pattern` as a key that implies `opt`ness. Should be fixed now, thanks!

thezealousfool:
@wbthomason I can confirm this closes #478. Thanks.

wbthomason:
Great, thank you for testing!

PR #511, opened 2021-07-27
Feature: add very simple breaking change detection and marking

wbthomason:
This is a slapdash implementation of the idea @sunjon, @clason, and others proposed on Gitter. It looks at commit headers and bodies for an update and, if it finds the strings "breaking change" or "breaking_change" (case insensitive), marks the commit as potentially breaking, emitting a warning log message and highlighting the commit hash in the status window.

### DIFF
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -35,8 +35,8 @@ end
 
 local breaking_change_pattern = [=[[bB][rR][eE][aA][kK][iI][nN][gG][ _][cC][hH][aA][nN][gG][eE]]=]
 local function mark_breaking_commits(plugin, commit_bodies)
-  local commits = vim.split(table.concat(commit_bodies, '\n'), '===COMMIT_START===')
-  for _, commit in ipairs(commits) do
+  local commits = vim.gsplit(table.concat(commit_bodies, '\n'), '===COMMIT_START===', true)
+  for commit in commits do
     local commit_parts = vim.split(commit, '===BODY_START===')
     local body = commit_parts[2]
     local lines = vim.split(commit_parts[1], '\n')
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -33,6 +33,21 @@ local function ensure_git_env()
   end
 end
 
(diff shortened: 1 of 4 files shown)


clason:
Looks good! One further suggestion: people might have set `show_all_info = false` in their config so they won't see the individual commits and will miss the warning.

In this case (or unconditionally?) you could apply the highlighting to the plugin name (as well).

wbthomason:
Good point; that would help.

wbthomason:
Highlighting plugin names done in b216e2a.

lua/packer/plugin_types/git.lua
```diff
@@ -33,6 +33,21 @@ local function ensure_git_env()
   end
 end
 
+local breaking_change_pattern = [=[[bB][rR][eE][aA][kK][iI][nN][gG][ _][cC][hH][aA][nN][gG][eE]]=]
+local function mark_breaking_commits(plugin, commit_bodies)
+  local commits = vim.split(table.concat(commit_bodies, '\n'), '===COMMIT_START===')
+  for _, commit in ipairs(commits) do
```
nanotee, line 39: Using the `vim.gsplit()` iterator might be a little faster, since you don't have to loop over the string to collect it into a table first (I don't have any benchmarks to back this up though)
```suggestion
  local commits = vim.gsplit(table.concat(commit_bodies, '\n'), '===COMMIT_START===', true)
  for commit in commits do
```

I just checked this branch out locally, works very well! Love this feature already :+1:

wbthomason: Thanks; this seems like a reasonable suggestion. Even if it's not faster, I don't think it will be slower, and it's a tad more idiomatic.

wbthomason:
This has worked for several other people and seems unlikely to break anything (it hasn't in my testing), so I'm going to go ahead and merge!

clason:
Just a brief suggestion that doesn't merit a new issue: I just had a plugin update failure (someone force-pushed to master; not a packer issue) and saw that update failures also get highlighted as `WarningMsg`. I don't think they should use the same highlight group, though: an update failure is an error, so I'd prefer `ErrorMsg` for actual errors (leaving `WarningMsg` for breaking changes as now).

wbthomason:
@clason it is done.

Issue #510, opened 2021-07-26
Way to get or add luarocks package bin path


s1n7ax:
I installed `luaformatter` but when running `:Format` command, it cannot find `lua-format`.
```lua
use {
    'mhartington/formatter.nvim',
    config = function()
        require("formatter").setup({
            logging = false,
            filetype = {
                lua = {
                    function()
                        return {exe = "lua-format", args = {}, stdin = true}
                    end
                },
            }
        })
    end
}
```

However I noticed the `lua-format` bin is available under following directory.

```bash
 ~/.cache/nvim/packer_hererocks/2.0.5/bin
```

Is there a way to get the path to bin grammatically so I can set `env`. I guess I'm expecting something like below.

```lua
    rocks {
        'luaformatter',
        server = 'https://luarocks.org/dev',
        env = {
            PATH = string.format("$PATH:%s", packer.luarocks.path)
        }
    }
```

gegoune:
Wrong project? :)

wbthomason:
Did you install `lua-format` via `packer`'s Luarocks integration? That's a somewhat unusual use of it - it's mostly meant for installing libraries as modules.

I wouldn't really recommend installing binaries this way, but you could look at the code in `packer.luarocks` to see what paths we do/could possibly expose for this kind of use.

s1n7ax:
I can't think of a reason why I shouldn't do this. `formatter` depends on one of the packages.  In this case it happened to a binary.

I have no clue how to add `bin` to the `PATH`. But I created a pull request to add an API to expose the path to `bin`. 
#515

PR #515, opened 2021-07-27
Add feat to expose get_bin_path API

s1n7ax:
Packages like `mhartington/formatter.nvim` may depend on executable files installed through `luarocks`. At the moment there is no any API to get the `hererocks/bin` path. This exposes `packer.luarocks.get_bin_path()` API that returns the absolute path to bin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,28 @@ return require('packer').startup(function()
     rocks = {'lpeg', {'lua-cjson', version = '2.1.0'}}
   }
 
+  -- Plugint can get path to rocks binaries
+  use {
+    'mhartington/formatter.nvim',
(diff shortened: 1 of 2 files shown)


akinsho:
@s1n7ax will leave this to @wbthomason to decide, but I think the point of the luarocks functionality is for adding *libraries*, not binaries. I think those make most sense installed as you would do other binaries on your system and using the path to that. Seems like we'd be inviting people to try and use packer to manage binaries by adding this sort of API

wbthomason:
Sorry for taking so long to get to this. I'm inclined to agree with @akinsho, as I mentioned in #510. If we added this, it would need to be explicitly unsupported, as allowing arbitrary Luarocks binaries to be managed by `packer` feels to me like scope creep that can open up a whole new set of issues to worry about.

Issue #519, opened 2021-07-29
Messy documentation for floating window config


n3wborn:
Hi !  I was about to change packer's config to have floating windows and simple border when I found that config documentation is sometimes messy, so I decided to fill an issue :-)

From README :

```lua
packer.startup({function()
  -- Your plugins here
end,
config = {
  display = {
    open_fn = require('packer.util').float,
  }
}})
```

**function() is between brackets and `config = {}` is present**

From packer.txt :

```lua
  packer.startup(function()
    -- Your plugins here
  end, {
    display = {
      open_fn = require('packer.util').float,
    }
  })
```

**function() is no more between brackets and `config = {}` is missing**

We can see the same about floating window with simple border. As this could lead to confusion and useless issues, it could be better to fix these sections from packer.txt. I'll make a PR, not really used to but I'll do my best. Hope this will help.

Thanks for your work on packer !

akinsho:
Fixed by #520

PR #520, opened 2021-07-29
Fix: doc-floating-window (#519)

n3wborn:


akinsho (approved):
@n3wborn thanks for the fix. It looks good to me 👍🏿

wbthomason:
Nice catch, @n3wborn, and thanks for the PR! Sorry for the lag; I missed this in my notifications until @akinsho's comment bumped it to the top.

PR #528, opened 2021-08-01
Ensure rocks even if there are no plugins to install

### LINKED ISSUE #524: `:PackerInstall` doesn't install `use_rocks` packages
- `nvim --version`: v0.6.0-dev+1570-g15698eb5a
- `git --version`: 2.32.0
- Operating system/version: Solus 4.3 Fortitude
- Terminal name/version: Konsole 21.04.2

### Steps to reproduce

Create this plugin specification:
```lua
require('packer').startup(function(use, use_rocks)
    use 'wbthomason/packer.nvim'
    use_rocks 'plenary.nvim'
end)
```

Restart Neovim and run `:PackerInstall`

### Actual behaviour

The display windows never opens and the rock is not installed

### Expected behaviour

`:PackerInstall` should install the missing rock

### Additional info

Here's some more debugging info (might not be 100% accurate because I'm not super familiar with that part of the code)

I think the problem comes from there:
https://github.com/wbthomason/packer.nvim/blob/74452f3a48cc351a5444adc4ad56678d479ff849/lua/packer.lua#L383-L391

The code called by `get_fs_state()` only checks `start` and `opt` plugins, which means standalone rocks don't appear in `fs_state.missing` and the function returns early.

An additional issue I found is that rocks are only added to the list of tasks if there are already tasks scheduled for normal plugins:
https://github.com/wbthomason/packer.nvim/blob/74452f3a48cc351a5444adc4ad56678d479ff849/lua/packer.lua#L398-L402

If there are no new regular plugins to install, the `tasks` table is empty and Luarocks tasks will never start

wbthomason:
Closes #524 by always checking if we need to ensure the installation of any rocks.

@nanotee I have not tested this at all yet, but I think this is (modulo some small bugfixes) what needs to happen to fix #524.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -384,25 +384,28 @@ packer.install = function(...)
     if not install_plugins then
       install_plugins = vim.tbl_keys(fs_state.missing)
     end
-    if #install_plugins == 0 then
-      log.info 'All configured plugins are installed'
-      packer.on_complete()
-      return
-    end
 
     await(a.main)
     local start_time = vim.fn.reltime()
     local results = {}
     await(clean(plugins, fs_state, results))
+
+    if #install_plugins == 0 and #rocks == 0 then
+      log.info 'All configured plugins and rocks are installed'
+      packer.on_complete()
+      return
+    end
+
     await(a.main)
     log.debug 'Gathering install tasks'
     local tasks, display_win = install(plugins, install_plugins, results)
+    log.debug 'Gathering Luarocks tasks'
+    local luarocks_ensure_task = luarocks.ensure(rocks, results, display_win)
+    if luarocks_ensure_task ~= nil then
+      table.insert(tasks, luarocks_ensure_task)
+    end
+
     if next(tasks) then
-      log.debug 'Gathering Luarocks tasks'
-      local luarocks_ensure_task = luarocks.ensure(rocks, results, display_win)
-      if luarocks_ensure_task ~= nil then
-        table.insert(tasks, luarocks_ensure_task)
-      end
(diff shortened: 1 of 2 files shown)


nanotee:
Sorry for taking so long to get back to this. Gave this PR a go, still can't install rocks unfortunately. I think this is because the `rocks` table is a dict-like table, so getting the length with `#` always returns `0`.

An idea I had to simplify code reuse is to make `rocks` a plugin type (like `git` and `local`). Do you think that would make sense as a long term goal?

wbthomason:
Ah, you're right. I can fix that check to work properly.

re: better code reuse: I like that idea in principle, but it would require re-engineering a bit to change how we check if a given plugin is installed/not. Right now we batch the filesystem work for this for plugins, but if we treated rocks as a different kind of plugin we'd need to think about how to restructure.  Otherwise, it fits nicely with the existing architecture for defining custom installers/updaters for plugins.

wbthomason:
2eaff76 hypothetically fixes the check for rocks to install.

nanotee:
Thanks for the quick fix. `packer` now detects that there are `rocks` to install, but it can't open the display window:
```
[packer.nvim] [ERROR 11:41:30] async.lua:20: Error in coroutine: ...e/nvim/site/pack/packer/start/packer.nvim/lua/packer.lua:414: attempt to index local 'display_win' (a nil value)
```

wbthomason:
Whoops. This comes from a bad design decision in how `packer` opens its display. I'll see if I can figure out a workaround, or just bite the bullet and refactor.

PR #532, opened 2021-08-03
fix: source remote plugins after updating

### LINKED ISSUE #464: Remote plugin with conditional loading is not handled well
- `nvim --version`: Latest
- `git --version`: 2.21.0
- Operating system/version: macOS
- Terminal name/version: kitty

### Steps to reproduce

```
lua << EOF
require('packer').startup(function()
    use {'numirias/semshi', run = ':UpdateRemotePlugins', ft = 'python'}
end)
EOF
```

1. Use the above config to start nvim: `nvim -u init.vim`.
2. Use PakcerCompile to generate packer_compiled.lua.
3. Run `PackerUpdate` to update the plugin.
4. Open a Python source file.

### Actual behaviour

We see error that command `Semshi enable` is not available.

### Expected behaviour

No error occurs.

### Analysis of the error

If Semshi is set to opt plugin, when we run `:PackerUpdate`, the command `:
UpdateRemotePlugins` will be run. At this time, since Semshi is only loaded
when we are editing a Python source file, it is not in the runtimepath. So its
manifest will not be generated in rplugin.vim (located under
~/.local/share/nvim/rplugin.vim). Thus, when we open a Python file afterwards,
although Semshi is in runtimepath, but its manifest is missing, we see error
that command `Semshi enable` is not found.

After loading this plugin, i.e., when we open a Python source file,
we need to run `:UpdateRemotePlugins` manually so that semshi can work correctly.

Maybe the user should be warned if they try to conditionaly load a nvim remote plugin.

akinsho:
The approach in this I think is perhaps not what you had in mind @wbthomason, this PR currently works by loading remote plugins during update so that the `UpdateRemotePlugin` works but I guess this forces them to load not at the point a user might have specified for that to happen, so doing it after packadd is probably better.

Will fix #464


akinsho:
I've essentially refactored my initial approach in my subsequent one, so you can have a look at how I was originally doing it.

akinsho:
Actually not sure if either of them *actually* work... 🤔

akinsho:
This seems to be what dein does

https://github.com/Shougo/dein.vim/blob/bc6fa9611736707a85e955bb02d2ae265c71f983/autoload/dein/autoload.vim#L330-L339

But frankly I have no idea what that achieves

lua/packer/plugin_utils.lua
```diff
@@ -208,9 +208,21 @@ plugin_utils.load_plugin = function(plugin)
   end
 end
 
+plugin_utils.is_remote_plugin = function(plugin)
```
wbthomason, line 211: Do we use this anywhere?

akinsho: @wbthomason nope it was from the previous commit, I left it in place incase it might be useful in the future/I wasn't really sure what direction to take this PR in since neither approach resolve the issue 100% for me. I think I'm still trying to get clear on exactly what is happening.

It seems like if I run update whilst the rplugin is still not loaded, then the manifest is created without it's content. So when I next go to load the plugin (having restarted nvim since the original session has sourced the file before it changed) then I see the errors.

Somehow adding it just after packadd doesn't seem to always work though. I need to test this again.

akinsho:
@jdhao or @wbthomason can either of you test out this PR and see if it fixes #464. I was having some local issues (with my dev version of packer not being used) so I thought this wasn't working but managed to test it more clearly, and it seems to resolve the issue for me

lua/packer/util.lua
```diff
@@ -9,6 +9,19 @@ util.map = function(func, seq)
   return result
 end
 
+---Find an item in a list based on a compare function
+---@generic T
+---@param compare fun(item: T): boolean
+---@param list `T`
+---@return `T`
+util.find = function(list, compare)
```
akinsho, line 17: I've left this is since I assumed it might come in handy at some point 🤷🏿‍♂️

amenonsen:
Doesn't seem to fix #464 for me.

akinsho:
@amenonsen can you describe how exactly you are testing the plugin? i.e. what bug with what plugin exactly are you hitting and what steps did you take to try this PR out, just to make sure we're describing the same bug/fix.

In my case, the specific issue I'm testing is the ability to lazy load wilder.nvim on `CmdLineEnter` which repeatedly fails before this PR because `UpdateRemotePlugins` can be called by packer sync et al. before it is loaded, so its data isn't in the remote plugin manifest.

If you add the update portion of this PR to the same file where your local packer is installed in my case anyway this seems to fix the bug which I've just tried again today, and it still seems to work 🤷🏿

wbthomason:
This also works in my testing. @amenonsen, do you have replication steps for your `vim-ultest` case? Otherwise, I think this is OK to merge.

Down the road, I wonder if we could check for `rplugin/` at install/update time, statically gather a list of the relevant files into the compiled file, and just directly source those on load rather than relying on `UpdateRemotePlugins`. This might be slightly faster (though honestly probably not in a meaningful way).

akinsho:
@wbthomason having messed around with this for a little bit I'm not sure this would work because running `UpdateRemotePlugins` updates the `rplugin.vim` manifest files with the functions that are then actually used by remote plugins, I think it needs to be populated in order for remote plugins to work correctly and so just trying to source the files in the `rplugin` directory wouldn't work especially as the files in that directory from the few remote plugins I use/have looked at are in the language of the plugin e.g. in my case `wilder.nvim`'s rplugin dir only contains python (last I checked/noticed).

It would be nice to do this more granularly though and only run the update command if it was relevant, so  compiling a list at install or update is an interesting idea. In the first commit of this PR I actually wrote a function that used `glob` to check through the packer start and opt dir for all rplugin containing folders so could actually work that into a variable that gets stored in the compiled file and use that list to only run the update where necessary but yeah maybe best as a future optimisation.

wbthomason:
That's super useful to know, thanks!

amenonsen:
My original configuration (which I had posted on #464) looked like this:

```lua
-- Integrates with vim-test and nvim-dap to run tests.
    use {
        { 'vim-test/vim-test', ft = { 'python' } },
        {
            "rcarriga/vim-ultest", ft = { 'python' },
            after = { "vim-test" },
            run = function ()
                -- vim.cmd[[packadd vim-test vim-ultest]]
                vim.cmd[[PackerLoad vim-test vim-ultest]]
                vim.cmd[[UpdateRemotePlugins]]
            end, …
        }
    }
```

I removed the `run` function and applied the load.lua patch locally again just now, and it does seem to work correctly. Sorry for the false alarm, I must have botched my earlier test somehow.

amenonsen:
That said, however, running :UpdateRemotePlugins after packadd the way this patch does means that if I change the condition from `ft = { 'python' }` to, say, `cmd = "Ultest"`, then type `:Ultest`, I see the error twice:

```
Error: vim-ultest remote function not detected, try running :UpdateRemotePlugins on install/update
Error: vim-ultest remote function not detected, try running :UpdateRemotePlugins on install/update
```

But it does work fine after that. I'm not sure what's going on there. (Does UpdateRemotePlugins run async?)

As a workaround, I was doing this instead, which also works:

```
    autocmd User PackerComplete exe "PackerLoad vim-test vim-ultest" | silent UpdateRemotePlugins
```

akinsho:
Hmm will look into it, I'm not sure why using a command would be any different since it goes through the same code path (presumably)

akinsho:
Oh wait @amenonsen do you mean that you changed the condition and the ran packer compile again or something in the same neovim session you see the error or if you change to use `cmd` run compile and reload nvim specifically using `cmd` is what doesn't work?

amenonsen:
No, not the same session. I make the config changes in plugins.lua, run PackerSync, exit Neovim, then restart and open a Python file and run 'Ultest', and then I get those errors.

akinsho:
@amenonsen can you try the most recent version and see if that works, it was more aimed around being a little more conservative with when we call `UpdateRemotePlugins` but it also looks like it might have stopped the Ultest error on my end. If it did, I'm not really sure why tbh

amenonsen:
It doesn't seem to make any difference here.

amenonsen:
I confess that I don't understand this rplugin stuff very well, but the `vim-ultest remote function not detected` error is from vim-ultest's `autoload/ultest/handler.vim`. Wouldn't that code get run in between packadd and UpdateRemotePlugins?

wbthomason:
@amenonsen I'm not very familiar with `vim-ultest`, but what you describe would only happen if something in one of the runtime directories (e.g. `plugin/`) within `vim-ultest` calls a function in `autoload/ultest/handler.vim` (which is possible). Unless this is the case, the `UpdateRemotePlugins` call *should* happen before the `Ultest` command is run, triggering the code in `autoload/ultest/handler.vim`.

akinsho:
Hmm, looking at Ultest's plugin directory there is this line 👉🏿 https://github.com/rcarriga/vim-ultest/blob/416c58d00280c452f4c8c75866393394031fcb6b/plugin/ultest.vim#L8
which creates a function reference of one of the handlers in the autoload dir, so that could be the cause. Although I'm not sure why the error happens though, it seems to come from a generic helper function which is called by most of Ultests test mechanisms, but I don't know why any of them are running yet since the `Ultest` command is not actually called later after the load and remote plugin has been loaded.

I can't follow the full flow but dein doesn't seem to do anything massively different that I can tell
https://github.com/Shougo/dein.vim/blob/8a6aa83bd8/autoload/dein/install.vim#L546-L575

rcarriga:
Hey coincidentally came across this thread today, thought I'd make myself available in case I can be of help :smile: 

> Hmm, looking at Ultest's plugin directory there is this line 👉🏿 https://github.com/rcarriga/vim-ultest/blob/416c58d00280c452f4c8c75866393394031fcb6b/plugin/ultest.vim#L8

Just to point out that the offending line which calls the remote functions is actually here https://github.com/rcarriga/vim-ultest/blob/master/plugin/ultest.vim#L421. You'll see a linked issue in pynvim that is the reason the line is there. However even if it wasn't it would still make remote calls if there is an open test file. 

I don't have time right now to fully catch up on the nitty gritty details of this issue but please feel free to ask questions, and I'll try make time to help.

akinsho:
Thanks for pointing that line out @rcarriga 🙏🏿, I'm not sure, but I think maybe this issue is more about how packer handles conditionally loaded remote plugins though, rather than any specific thing the Ultest is doing that is problematic. I think things are happening out of order here because calling `packadd` causes the plugin to be loaded, presumably also sourcing its `plugin/*.vim` file which causes the line you pointed out to load and not find the remote plugin because `UpdateRemotePlugin` on the next line hasn't run yet.

I just tried a change to this logic @wbthomason which uses `vim.opt.runtimepath` to add rplugins to the runtime path and then call update remote plugins which seems to work to update the remote plugin manifest without sourcing the plugin till the packadd which happens last.

@amenonsen can you give this latest change a go, I think it solved things on my end but could just be random/flaky 😅

amenonsen:
The latest patch doesn't seem to have made any difference here. I run `:PackerSync`, check that my `rplugin.vim` is empty, open a Python test file, and run `:Ultest`. I see the two "remote function not detected" errors, and although Ultest is loaded (e.g., I can see all the commands), nothing works. If I quit and start nvim again and run Ultest, everything works fine (because rplugin.vim is fine by then). I haven't changed anything else in the configuration, just applied the patch to lua/packer/load.lua.

akinsho:
Hmm I just repeated the steps you described here in a random python and lua file, and it works without issue. I haven't seen the two error messages since the most recent commit here 🤔. Are you sure that you've definitely added the last commit, tbh rather than manually applying the diff you can just switch to my fork and branch to test it and packer will pull the changes down.

rcarriga:
Can confirm that I'm having the same issue as @amenonsen. 
1. `:PackerSync`
2. Close neovim
3. Open neovim
4. `:UltestSummary`
5. 
The error message occurs but restart neovim and it works fine.

 Interestingly if I run an ultest command after a sync within the same session, the next time I open it works fine.
1. `:PackerSync`
2. `:UltestSummary`
3. Close neovim
4. Open neovim
5. `:UltestSummary`

No error message, everything works

akinsho:
Well in which case I'm thoroughly stumped then. The changes here add Ultests files to the runtime path first so nothing is sourced yet, it then runs update remote plugins which populates the remote plugin manifest and then lastly it calls packadd which then sources the actual plugin. I have no idea how it's possible for the file in the plugin directory to be being called and not find the contents of the `rplugin.vim` maybe an error is occurring during `UpdateRemotePlugins` on either of your ends, can you try removing the `silent` from the command and see if any errors are printed.

amenonsen:
Well, that shows this:
```
remote/host: python3 host registered plugins ['ultest']
remote/host: generated rplugin manifest: /home/ams/.local/share/nvim/rplugin.vim
Error: vim-ultest remote function not detected, try running :UpdateRemotePlugins on install/update
Error: vim-ultest remote function not detected, try running :UpdateRemotePlugins on install/update
Press ENTER or type command to continue
```

Could it be that (the empty) rplugin.vim is loaded once earlier during startup, the new one is generated by your patch but not (re)loaded, and that's why ultest doesn't work until you restart?

amenonsen:
Indeed, adding one line to your patch after the call to UpdateRemotePlugins fixes it:

```
diff --git a/lua/packer/load.lua b/lua/packer/load.lua
index 07b52ee..0c0cbee 100644
--- a/lua/packer/load.lua
+++ b/lua/packer/load.lua
@@ -37,6 +37,16 @@ packer_load = function(names, cause, plugins)
         end
       end

+      local is_rplugin = vim.loop.fs_stat(require('packer.util').join_paths(plugin.path, 'rplugin'))
+
+      if is_rplugin and vim.fn.exists ':UpdateRemotePlugins' == 2 then
+        -- Add the plugin to the runtime path so that it's remote plugin directory can be found
+        -- so that the remote plugin manifest is correctly generated but the plugin is not sourced yet
+        vim.opt.runtimepath:append(plugin.path)
+        cmd 'silent UpdateRemotePlugins'
+        cmd('source ' .. vim.fn.stdpath('data') .. '/rplugin.vim')
+      end
+
       cmd('packadd ' .. names[i])
       if plugin.after_files then
         for _, file in ipairs(plugin.after_files) do
```

akinsho:
@amenonsen 🙌🏿 thanks for digging in, tbh I played with that idea on one iteration but I hadn't swapped the order of the `packadd` yet so it made no difference at that point. I had just assumed that `UpdateRemotePlugins` would source the file it created. Will add that change to this PR 👍🏿

lua/packer/load.lua
```diff
@@ -37,6 +37,15 @@ packer_load = function(names, cause, plugins)
         end
       end
 
+      local is_rplugin = vim.loop.fs_stat(require('packer.util').join_paths(plugin.path, 'rplugin'))
+
+      if is_rplugin and vim.fn.exists ':UpdateRemotePlugins' == 2 then
+        -- Add the plugin to the runtime path so that it's remote plugin directory can be found
```
amenonsen, line 43:
```suggestion
        -- Add the plugin to the runtime path so that its remote plugin directory can be found
```

akinsho:
Ok added in your suggestion @amenonsen if either you or @rcarriga could give it a go to confirm it works I'd appreciate it (presuming it will given you already tried it @amenonsen).

Granted it all works this is ready for a review/merge @wbthomason (can remove the find util if you'd rather not have it we might never need it 🤷🏿 )

akinsho:
As an update on this (would be good to get some feedback on the current version, in case it somehow automagically works) but I've made a few discoveries actually having read through dein and neovim's remote plugin code now. So dein actually doesn't offer event or command lazy loading for remote plugins but rather offers to lazy load all remote plugins at once or not.

The main reason being that each remote plugin after `UpdateRemotePlugins` gets run adds a register function to the remote plugin manifest which ends up looking like
```vim
" python3
call remote#host#RegisterPlugin('wilder.nvim', ...)
call remote#host#RegisterPlugin('vim-ultest', ...)
```
This file needs to be sourced at some point before the plugins can be used, it essentially can only be sourced once because the `RegisterPlugin` function can only be called once without erroring so imagine `wilder.nvim` is loaded and the manifest it updated, the second line for `vim-ultest` won't have been added yet since it is only written when ultests rplugin files are in the runtimepath and `UpdateRemotePlugins` is called so sourcing `rplugin.vim` at this point will prevent it being called again to correctly register `vim-ultest` when the `Ultest` command is eventually used since the `RegisterPlugin('wilder.nvim')` will error because it was already called.

Would be possible if upstream would just swallow the error and let the rest of the file continue to run but it doesn't (maybe one for a future upstream PR).

In terms of a solution on packer's end, I think one not too expensive thing that packer could do is register an event on something like `VimEnter` which globs the `packer/opt` dir for all `rplugin` files then adds them all to the runtimepath and calls `UpdateRemotePlugins` at this point the remote plugin manifest should be fully filled out and so the actually packadd for the plugins can occur whenever the lazy load trigger occurs.

I guess its kind of a halfway house between dein and how packer works for other plugins. I think just adding files to the runtime path should cost very little and the actual possible expensive sourcing of the plugins files will still be delayed till the packadd point

Thoughts @wbthomason?

lua/packer.lua
```diff
@@ -20,6 +20,7 @@ local config_defaults = {
   transitive_opt = true,
   transitive_disable = true,
   auto_reload_compiled = true,
+  rplugin_load_event = 'BufReadPre',
```
akinsho, line 23: This is different to how `dein` handles loading remote plugins since a user doesn't specify an event instead, as far as I can tell, remote plugins are set up when the first lazy loaded plugin is loaded and then a `g:_dein_rplugins_loaded` var is set and used a guard against future runs. Not sure if this is better or worse, it seems more automatic 🤷🏿‍♂️ so can change it to work like that

lua/packer/rplugin.lua
```diff
@@ -0,0 +1,31 @@
...
+  ]]
+end
+
+---@param path string
+function M.get(path)
+  return vim.fn.globpath(util.join_paths(path, '*'), 'rplugin', false, true)
+end
+
+---@param opt_dir string
+function M.has_remote_plugins(opt_dir)
```
akinsho, line 26: Would be nice if config could be reached from anywhere by having it in its own module e.g.
```lua
local M = {}

local config = {}

function M.get()
  return config
end

function M.set()
  return config
end

-- In rplugin.lua
local config = require('packer.config').get()
```

ATM config is only passed in by calling a function on a callable class, which each module must return which seems a little more involved.

amenonsen:
> In terms of a solution on packer's end, I think one not too expensive thing that packer could do is register an event on something like VimEnter which globs the packer/opt dir for all rplugin files then adds them all to the runtimepath and calls UpdateRemotePlugins at this point the remote plugin manifest should be fully filled out and so the actually packadd for the plugins can occur whenever the lazy load trigger occurs.

I can't help but think that it might be better to do this at `Packer{Install,Sync,Update}` time.

akinsho:
I think whilst that would be in a way arguably tidier than from packer's perspective, it means that rather than ensuring that the plugin manifest is accurate for each vim session it will be possible for the user to break things unintentionally and very confusingly for them since if they were to hit `UpdateRemotePlugins` for whatever reason without having called a packer management command already then they would override the plugin manifest with an incomplete one since the plugins won't be in the runtimepath and so their next run of neovim will have errors unless they happened to run another management command beforehand.

I could be that this is worth documenting and doing anyway from a tidiness perspective but I've erred more towards something like what dein is doing (as far as I can tell) since packer management commands might be called very infrequently by a user especially if they don't change plugins much.

amenonsen:
> it will be possible for the user to break things unintentionally and very confusingly for them since if they were to hit UpdateRemotePlugins for whatever reason without having called a packer management command already then they would override the plugin manifest with an incomplete one

That's fair enough, but isn't it already that way? If I run `UpdateRemotePlugins` by hand without loading vim-ultest in advance, I get an empty manifest. Doing a glob search for all installed remote plugins at `PackerXXX` time seems much more reasonable than doing that search at `packadd onething` time.

akinsho:
@amenonsen just to be clear the current implementation, similar to dein's removes this potential pitfall more or less entirely since just after vim is started all the rplugins are globbed for added to the rtp and then the manifest is generated, so subsequent manual updates won't cause any trouble.

The only advantage of moving to doing this at `PackerX` time is getting round having an autocommand at startup. Which I think isn't a particularly big deal in any event, whereas switching to `PackerX` allows this old issue to still occur, i.e. seems like a net negative to me 🤷🏿‍♂️.

In any event, I think the big question about if this is how we want to tackle this in packer at all still needs @wbthomason's input, so I don't want to finesse the details of that till we're actually agreed on this general direction anyhow.

PR #543, opened 2021-08-10
feat(keys) adds function so easier specify many key combinations

AckslD:
I wrote some utility function to make it easier to specify many key combinations and thought I would share it here if you find it useful.

Previously for example I wrote:
```lua
use {
    'michaeljsmith/vim-indent-object',
    keys = {
        {'o', 'ii'},
        {'o', 'ai'},
        {'o', 'iI'},
        {'o', 'aI'},
        {'v', 'ii'},
        {'v', 'ai'},
        {'v', 'iI'},
        {'v', 'aI'},
    },
    wants = 'vim-textobj-user',
}
```
but with this I can write:
```lua
use {
    'michaeljsmith/vim-indent-object',
    keys = get_keys({'o', 'v'}, {{'a', 'i'}, {'i', 'I'}}),
    wants = 'vim-textobj-user',
}
```
which expands to the product of the individual keys in the specified modes.

Some examples:
```lua
local get_keys = require('packer').get_keys
print(vim.inspect(get_keys('n', 'a')))
print(vim.inspect(get_keys('n', {'a', 'b'})))
print(vim.inspect(get_keys('n', {{'a', 'b'}, {'c', 'd'}})))
print(vim.inspect(get_keys({'n', 'o'}, {{'a', 'b'}, {'c', 'd'}})))
```
which gives
```
{ { "n", "a" } }
{ { "n", "a" }, { "n", "b" } }
{ { "n", "ac" }, { "n", "ad" }, { "n", "bc" }, { "n", "bd" } }
{ { "n", "ac" }, { "n", "ad" }, { "n", "bc" }, { "n", "bd" }, { "o", "ac" }, { "o", "ad" }, { "o", "bc" }, { "o", "bd" } }
```

I considered trying to have some way to automatically parse these structures so that one does not need to even do the function call but not sure it's possible/worth it.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -1,6 +1,7 @@
 -- TODO: Performance analysis/tuning
 -- TODO: Merge start plugins?
 local util = require 'packer.util'
+local keys = require 'packer.keys'
 
 local join_paths = util.join_paths
 local stdpath = vim.fn.stdpath
@@ -822,5 +823,7 @@ packer.startup = function(spec)
 
   return packer
 end
+
+packer.get_keys = keys.get_keys
 
 return packer
--- a/lua/packer/keys.lua
+++ b/lua/packer/keys.lua
@@ -1,0 +1,47 @@
+local M = {}
+
+local function string_product(keys)
+    if #keys == 1 then
+        return keys[1]
+    end
+    local keys_copy = {}
+    for _, key in ipairs(keys) do
+        table.insert(keys_copy, key)
+    end
+    local product = {}
+    local first = table.remove(keys_copy, 1)
+    for _, a in ipairs(first) do
+        if type(a) ~= 'string' then
+            vim.cmd('echoerr "Not a valid keys table"')
+        end
+        for _, b in ipairs(string_product(keys_copy)) do
+            table.insert(product, a .. b)
+        end
+    end
+    return product
+end
+
+local function expand_keys(keys)
+    if type(keys) == 'string' then
+        return {keys}
+    elseif type(keys[1]) == 'string' then
+        return keys
+    else
+        return string_product(keys)
+    end
+end
+
+M.get_keys = function(modes, keys)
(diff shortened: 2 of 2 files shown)


wbthomason:
Thanks for this, and sorry for taking a full month to get to looking at it! I like this functionality, but I'm not sure it needs to be built into `packer` itself. Perhaps this would make more sense as a snippet we could provide in the documentation, or as part of a utility library like `plenary`?

AckslD:
No problem @wbthomason! :) Yeah I also wasn't sure where this would best live but thought I would share it here if you had some ideas. Maybe as a snippet in the documentation indeed makes sense.

wbthomason:
How would you feel about adding this to a new "Tips and Tricks" page in the (newly activated) Wiki, or something along those lines?

AckslD:
Sounds good @wbthomason!

wbthomason:
Yes, I believe so - if you don't have permission to, please let me know and I'll see what the options are.

AckslD:
I'm not sure I have permission, it looks as follows for me:
gscreenshot_2021-09-17-142056

akinsho:
@AckslD I just added a tips and tricks page can you try clicking it and see if it allows you to edit the page, could be that @wbthomason needs to allow anyone to have access to the wiki

AckslD:
@akinsho Doesn't look like, unless I'm missing something:
gscreenshot_2021-09-17-152012

akinsho:
Think it's a permissions issue then

wbthomason:
Yep - I just enabled non-collaborators to edit wiki pages. @AckslD, do you mind trying again?

AckslD:
@wbthomason @akinsho I now added this to the tips and tricks :) Feel free to edit anything if you feel
https://github.com/wbthomason/packer.nvim/wiki/Tips-and-Tricks

PR #567, opened 2021-08-23
docs/README: avoid old API

justinmk:


wbthomason (approved):
Looks good, thank you for catching this!

PR #571, opened 2021-08-25
Fix: Add `--depth 1` flag to clone faster

NarutoXY:
This will result in a blazing fast clone speed even on slow 3g networks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ To get started, first clone this repository to somewhere on your `packpath`, e.g
 > Unix, Linux Installation
 
 ```shell
-git clone https://github.com/wbthomason/packer.nvim\
+git clone --depth 1 https://github.com/wbthomason/packer.nvim\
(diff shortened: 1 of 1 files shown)


akinsho:
@wbthomason this is a small README change WDYT? I think the speed improvement would be good but I was wondering how it would behave if a user tried to change their packer commit or change branches etc. will this end up meaning people who use the docs like this will have to deepen things first so might introduce a new class of issues?

akinsho:
Fair point 👍🏿

bnjmnt4n:
What about adding `--depth 1` to https://github.com/wbthomason/packer.nvim#bootstrapping as well?

wbthomason:
Sorry for the lag; still catching up on issues etc. after my absence. Thanks for handling this @akinsho. I'll add a commit now to resolve @bnjmnt4n's point.

PR #577, opened 2021-08-27
Fix #551: failing to run command post hooks on Windows

### LINKED ISSUE #551: The "run" command doesn't work on Windows.
- `nvim --version`: 0.5.0
- `git --version`: NA
- Operating system/version: Windows 10
- Terminal name/version: NA

### Steps to reproduce

Add a plugin with a run command. For example

    use {
        "baysmith/nvim-rs-hello",
        run = "cargo build --release",
    }

### Actual behaviour

`cmd.exe -c "cargo build --release"` is executed, but cargo doesn't run.

### Expected behaviour

`cmd.exe /c "cargo build --release"` should be executed and cargo should run. Note the difference of `-c` versus `/c`. I think that may be the problem.

spindensity:
Fix #551 

The default value of `vim.o.shell` on Windows is `cmd.exe` or `C:\WINDOWS\system32\cmd.exe`, and `cmd` shell rejects the parameter `-c`, use `/c` instead when the shell is `cmd`.

By the way, you can not just check the OS since the `powershell` on Windows should use `-c` parameter.

### DIFF
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -240,7 +240,13 @@ plugin_utils.post_update_hook = function(plugin, disp)
               stderr = jobs.logging_callback(hook_output.err, hook_output.output, nil, disp, plugin_name),
               stdout = jobs.logging_callback(hook_output.err, hook_output.output, nil, disp, plugin_name),
             }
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Thank you, this looks good to me! I don't use Windows, so my knowledge of nuances like this is lacking.

PR #581, opened 2021-09-01
Wrap the example PackerCompile autocmd in an augroup

amenonsen:
Applying the earlier example verbatim without thinking about it would lead to the autocommand being re-created each time plugins.lua was reloaded, and could make PackerCompile "feel slow" because it was run multiple times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,7 +203,7 @@ You can configure Neovim to automatically run `:PackerCompile` whenever `plugins
 [an autocommand](https://neovim.io/doc/user/autocmd.html#:autocmd):
 
 ```
-augroup packer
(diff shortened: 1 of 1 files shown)


amenonsen:
Unfortunately, there isn't an `<a name=…>` closer to the augroup/autocmds_loaded explanation in the docs for a more relevant link. I considered adding an extra sentence here about the need for augroup, but I think it would be overkill.

wbthomason (approved):
This looks good, thanks! I'm going to tweak the recommended augroup name, since there's a chance that `packer` itself might use a name like that in the future, but otherwise I think this is good to merge. Thanks again!

PR #605, opened 2021-09-17
fix: plugins loaded with cond not marked as loaded

shadmansaleh:
Currently plugins loaded with `cond` shows as not loaded in `PackerStatus`. This fixes that issue by marking them loaded.

TBH I'm not sure why it's directly loading with `:packadd` instead of calling loader from packer/load .

@wbthomason Is there a reason why it's like that ? Using single loader sounds like a better idea and that'll also solve this issue .

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -526,7 +526,8 @@ local function make_loaders(_, plugins, output_lua, should_profile)
   for condition, names in pairs(condition_ids) do
     local conditional_loads = {}
     for _, name in ipairs(names) do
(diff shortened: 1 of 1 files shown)


akinsho:
@shadmansaleh pretty sure this wasn't really intentional i.e. the naked packadd, I had a draft PR open for it in https://github.com/wbthomason/packer.nvim/pull/390 which I think we both forgot about

shadmansaleh:
> pretty sure this wasn't really intentional i.e. the naked packadd

I thought so too.

> I had a draft PR open for it in #390 

Great . Looks like a better solution . I also noticed that after plugins weren't loaded .

closing in favor of #390

PR #606, opened 2021-09-17
feat: enable :PackerUpdate/Install/Sync to operate on specific plugins

shadmansaleh:
- Now :PackerUpdate/Install/Sync can be used to operate on specific plugins
- PackerLoad now uses <f-args>

### DIFF
--- a/doc/packer.txt
+++ b/doc/packer.txt
@@ -174,7 +174,7 @@ load `packer` and perform common package management operations. The following
 commands work well for this purpose: >
 
   command! PackerInstall packadd packer.nvim | lua require('plugins').install()
-  command! PackerUpdate packadd packer.nvim | lua require('plugins').update()
+  command! -nargs=* -complete=customlist,v:lua.require'packer'.plugin_complete PackerUpdate lua require('packer').update(<f-args>)
   command! PackerSync packadd packer.nvim | lua require('plugins').sync()
   command! PackerClean packadd packer.nvim | lua require('plugins').clean()
   command! PackerCompile packadd packer.nvim | lua require('plugins').compile('~/.config/nvim/plugin/packer_load.vim')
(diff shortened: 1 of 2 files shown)


wbthomason:
Thanks, this looks good - I particularly like the `plugin_complete` inclusion.

While we're extending the update command, it might make sense to apply the same changes to the install and sync commands, the functions for which also optionally take a plugin argument.

shadmansaleh:
> it might make sense to apply the same changes to the install and sync commands

Indeed . Is there a marker for which plugins are installed or not installed ?

wbthomason:
I'm not sure I understand - what do you mean by "a marker"?

shadmansaleh:
> Not currently, but I think the plugin_complete function as it is will be fine to start. Installing an already-installed plugin is a nullop.

Right plugin_complete will be enough . Checking which plugins aren't installed will require lots of fs operations.

shadmansaleh:
Huh didn't think there were tests to fail :P . The runner had a stroke while grabing neovim archive .

Why are you guys still running tests against nightly instead of v0.5 ?

wbthomason:
Honestly, because the test infrastructure for `packer` doesn't get the attention it needs. I'll see about updating the tests to use both 0.5 and nightly in a separate PR; we can ignore the failing tests for now, as we don't test anything that this PR should affect anyway.

wbthomason (approved):
LGTM, thanks!

shadmansaleh:
> I'll see about updating the tests to use both 0.5 and nightly in a separate PR

You can checkout [shadmansaleh/lualine.nvim](https://github.com/shadmansaleh/lualine.nvim/blob/master/.github/workflows/ci.yml) I had recently updated it to use both nightly and 0.5 . But later disabled nightly as nightly had disappeared :P

PR #607, opened 2021-09-17
enhance: sort lines in PackerStatus

shadmansaleh:
Plugins appearing at random positions in `PackerStatus` is just annoying


wbthomason (approved):
Looks good, thanks!

PR #609, opened 2021-09-18
feat: allow using other lazy loaders along with cond

shadmansaleh:
@wbthomason the `cond` behaviour we were talking about yesterday.


shadmansaleh:
with https://github.com/wbthomason/packer.nvim/pull/609/commits/9b705315226410153e4dd2a1133b32358726099a you can add `cond=false` to a plugin and it'll become a opt plugin , won't be loaded by packer, won't be uninstalled, will be updated . Plus if you want you can manually load it with `PackerLoad!`.

Great for plugin testing IMO . exactly what I wanted from `disable` :D

PR #610, opened 2021-09-18
enhance: move plugins on :PackerCompile

shadmansaleh:
With this start <-> opt plugins will be moved to correct location on `:PackerCompile` . Preferably this would be implemented without code duplication but there are some duplications in packer.sync & packer.compile mainly because with packer.sync result matters as it'll show which plugins where moved in the display window.


lua/packer.lua
```diff
@@ -663,71 +663,84 @@ end
 
 --- Update the compiled lazy-loader code
 --- Takes an optional argument of a path to which to output the resulting compiled code
-packer.compile = function(raw_args)
+packer.compile = function(raw_args, donot_move_plugins)
```
wbthomason, line 666: Nitpick: I'd prefer this be `move_plugins` (which will also require changing - but imo simplifying - the conditionals that use this parameter).

lua/packer.lua
```diff
@@ -663,71 +663,84 @@ end
...
-
-  -- TODO: remove this after migration period (written 2021/06/28)
-  local old_output_path
-  if output_lua then
-    old_output_path = vim.fn.fnamemodify(output_path, ':r') .. '.vim'
-  else
-    old_output_path = vim.fn.fnamemodify(output_path, ':r') .. '.lua'
-  end
+    log.info 'Finished compiling lazy-loaders!'
+    print 'PackerCompile: Done'
```
wbthomason, line 720: We should remove this print statement before merging.

lua/packer.lua
```diff
@@ -663,71 +663,84 @@ end
...
+    packer.on_compile_done()
+
+    -- TODO: remove this after migration period (written 2021/06/28)
```
wbthomason, line 723: Doesn't need to be part of this PR, but we're out of the migration period and should remove this shim.

wbthomason (changes requested):
Thanks! I have two nitpicky requests (a variable name change and removing a debug print statement), but I think this looks good otherwise.

lua/packer.lua
shadmansaleh: This isn't for debugging but to let user know compilation is done . Since compilation is async now . :PackerCompile exits immediately and there's no way to know when it's done . I'm open for better way to let user know but there should be a way at lest :)

akinsho: @shadmansaleh how about using `vim.notify`?

shadmansaleh: > how about using vim.notify?

Works for me.

akinsho: Is there a similar thing for if compilation errors? this has always been a thing that I've hit quite often historically using packer and I currently have the call to `packer.sync` wrapped in a `pcall` to avoid explosions from typos or syntax errors, haven't properly reviewed this but is there a point where the errors can be reported to the user as well or how will that come through now?

lua/packer.lua
shadmansaleh: done

lua/packer.lua
shadmansaleh: Done

lua/packer.lua
akinsho: Oh nevermind, there's already vim.notify calls for errors

shadmansaleh: You'll be getting messages about errors . But yeah you can't catch them with pcall 😅

Can you catch errors in `packer.sync` ? it's aleady async isn't it ?

akinsho: Sorry that was a typo I meant `compile`

shadmansaleh: > Oh nevermind, there's already vim.notify calls for errors

Ohh right . Now I remember had added it for functions in #343 . The handling errors in sting is still hanging #392 😐

PR #616, opened 2021-09-25
docs(bootstrap): run sync function automatically after cloning packer

khuedoan:
An improvement on https://github.com/wbthomason/packer.nvim/commit/6391cbf414d0aa6e9dabcd4cc9bbadb8204c89b5 so we don't have to run `:PackerSync` manually on first run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,22 +223,27 @@ vim.cmd([[
 
 ## Bootstrapping
 
-If you want to automatically ensure that `packer.nvim` is installed on any machine you clone your
-configuration to, add the following snippet (which is due to @Iron-E) somewhere in your config **before** your first usage of
-`packer`:
+If you want to automatically install and set up `packer.nvim` on any machine you clone your configuration to,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -223,20 +223,32 @@ vim.cmd([[
...
-`packer`:
+If you want to automatically install and set up `packer.nvim` on any machine you clone your configuration to,
+add the following snippet (which is due to @Iron-E and @khuedoan) somewhere in your config:
+
 ```lua
 local fn = vim.fn
 local install_path = fn.stdpath('data')..'/site/pack/packer/start/packer.nvim'
 if fn.empty(fn.glob(install_path)) > 0 then
-  fn.system({'git', 'clone', '--depth', '1', 'https://github.com/wbthomason/packer.nvim', install_path})
-  vim.cmd 'packadd packer.nvim'
```
khuedoan, line 234: Look like this is only needed if we have packer configured as `opt`

PR #636, opened 2021-10-06
fix(compile) schedule loading to after event

AckslD:
As reported in https://github.com/AckslD/nvim-neoclip.lua/issues/26 it seems that autocommands defined in `config` are called twice if plugin is loaded by an `event`. To reproduce one can try the following small example:
```lua
    use {
        '~/some_path',
        event = 'TextYankPost',
        config = function()
            local count = 0
            _G.test_handler = function()
                count = count + 1
                print('called', count, 'times')
            end
            vim.cmd([[autocmd TextYankPost * lua _G.test_handler()]])
        end,
    }
```
(where `~/some_path` is some existing path)

If you compile this and yank something you will see `called 2 times`. Unless I'm mistaken this happens since the package (and `config`) is loaded while handling the event so the autocommand will be added and still executed. But also `doautocmd` is called which triggers the handler again.

This PR puts the loading in `vim.schedule` and will thus be handled after the event is "finished".

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -487,7 +487,7 @@ local function make_loaders(_, plugins, output_lua, should_profile)
     table.insert(
       event_aucmds,
       fmt(
-        'vim.cmd [[au %s ++once lua require("packer.load")({%s}, { event = "%s" }, _G.packer_plugins)]]',
+        'vim.cmd [[au %s ++once lua vim.schedule(function() require("packer.load")({%s}, { event = "%s" }, _G.packer_plugins) end)]]',
         event,
         table.concat(names, ', '),
         event:gsub([[\]], [[\\]])


AckslD:
@wbthomason Yeah that's cleaner indeed, however it does not seem to work at least in my setup and breaks other things. I guess because for other loading triggers (`ft`, `keys` etc) there is no corresponding `doautocmd`?

We could add another function somewhere (not sure where exactly) which is
```lua
schedule_load = vim.schedule_wrap(require("packer.load"))
```
which is only used when `event` is used?

Btw I have to give credit to @tjdevries and his [video](https://youtu.be/GMS0JvS7W1Y) without which it would have taken me much longer to understand what was going on here :)

wbthomason:
Hmm. `schedule_wrap`-ing `packer.load` *does* work in my testing, and doesn't appear to break things. To make sure we're on the same page, I'm modifying only this line: https://github.com/wbthomason/packer.nvim/blob/5972ba5352ace1bf7bbd221d946c3daca6c67ddc/lua/packer/load.lua#L127
to read
```lua
return vim.schedule_wrap(load_wrapper)
```
which, with your example spec, prints "called 1 times" as expected (I can reproduce the issue if I don't use `schedule_wrap`).

Do you have an example of what this breaks for you? I didn't notice any breakage, but it's possible I just didn't test sufficiently extensively.

AckslD:
@wbthomason Doing the change you mention above seems to break for me to lazy load by module. Does that work for you?

AckslD:
@wbthomason I think what breaks for me is that I have some plugins which gets loaded by `module` which are in turn called in the `config` of another plugin.

For example this is my lsp setup:
```lua
    use { -- lspconfig {{{
        'neovim/nvim-lspconfig',
        requires = {
            {'ray-x/lsp_signature.nvim', module = 'lsp_signature'},
            { -- {{{ null-ls
                'jose-elias-alvarez/null-ls.nvim',
                module = 'null-ls',
                config = 'require("plugin_settings.null_ls").config()',
            }, -- }}}
        },
        config = 'require("plugin_settings.lspconfig").config()',
    } -- }}}
```
which inside `pluging_settings.lspconfig` I call `require('null-ls')` which triggers the loading. Before and also with the submitted fix this works fine. However adding `schedule_wrap(load_wrapper)` gives the error:
```
Error in packer_compiled: /home/axel/.config/nvim/lua/plugin_settings/lspconfig.lua:143: module 'null-ls' not found:
```

AckslD:
@wbthomason any more thoughts on this? :)

PR #642, opened 2021-10-08
fix(load): handle nil mods properly

AckslD:
I hadn't update my packer in quite a while but after doing so I had some issues, for example when running a command that is set to lazy load a package I got the error:
```
no such command: nil Command
```
(or something like that)
which seems to be caused by `cause.mods` being `nil`. This change just replaces it with `''` in that case.


akinsho:
actually @AckslD I can't reproduce this anymore 🤔 is this happening for you for any command you use to lazy load?

akinsho:
Thanks for the repro steps @nanotee will merge this as is then since it's a minor change that could unblock some people 🤞🏿

AckslD:
I see, thanks @nanotee!

PR #647, opened 2021-10-11
Feature: add packer.use_with_context

wbthomason:
This PR adds a minor feature, the function `packer.use_with_context(ctx, ...)`. This is a convenience function intended to make it easier to specify a group of plugins which share some configuration keys.

Basic usage is:
```lua
use_with_context(
  {after = 'some-plugin', requires = 'some-common-library'}, 
  'my-plugin-1', 
  {'my-plugin-2', cmd = 'SomeCmd'}
)
```

This PR still needs to add documentation, and has not been thoroughly tested.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -309,8 +309,12 @@ packer.use = function(plugin_spec)
   }
 end
 
-packer.use_with_context = function(ctx, plugin_specs)
-  for _, plugin_spec in ipairs(plugin_specs) do
+packer.use_with_context = function(ctx, ...)
+  for _, plugin_spec in ipairs { ... } do
+    if type(plugin_spec) == 'string' then
+      plugin_spec = { plugin_spec }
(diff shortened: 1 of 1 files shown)


wbthomason:
@ndav1d, is this what you had in mind from https://github.com/hrsh7th/vim-vsnip/issues/219#issuecomment-940357882?

wbthomason:
The syntax is just that you pass a table with `packer` configuration keys as the first argument, then a list of plugin specifications (like normal) as the rest of the arguments. The keys from the first argument will be applied to the plugin specifications.

If you can give an example of what you mean by a group of plugins, that would be helpful. Feel free to imagine syntax, etc. as long as it's valid Lua.

wbthomason:
Precisely, yep!

ndavd:
hi @wbthomason any news on this? :D

Issue #198, opened 2021-02-06
Feature Request: Blocking Install


jzelinskie:
I'm trying to be clever while bootstrapping the first run of neovim to install packer. I'd actually like to block during install so that I don't get any errors when the rest of my config is loaded after kicking off the `packer.install()`.

cwebster2:
Seconding This.  What i'd love to do is be able to run `nvim --headless +PackerSync +qa` in my env installer script.

wbthomason:
Making a blocking install is more involved than it seems on its face, because the jobs we rely on are inherently non-blocking. We could use busy-waiting (e.g. a loop until the job done callback gets invoked), but that seems undesirable. I could be wrong, but I don't believe we have access to any of the usual synchronization primitives to block without spinning.

I'll need to think about the best way to do this - contributions or suggestions are welcome!

wbthomason:
Note to myself/anyone who wants to take this issue on: `vim.wait` might provide a good way to make things block without spinning.

nanotee:
I believe these use cases have been addressed at least in part with #259

<details>
<summary>Bootstrapping packer + plugins in init.lua without errors</summary>

```lua
local fn = vim.fn

-- Auto install packer.nvim
local install_path = fn.stdpath('data') .. '/site/pack/packer/start/packer.nvim'

if fn.isdirectory(install_path) == 0 then
  fn.system({'git', 'clone', 'https://github.com/wbthomason/packer.nvim', install_path})
  require('plugin_specification')
  vim.cmd 'autocmd User PackerComplete ++once lua require("my_config")'
  require('packer').sync()
else
  require('my_config')
end
```

</details>

<details>
<summary>Bootstrapping packer + plugins in a shell script</summary>

```sh
#!/usr/bin/env sh

install_packer() {
  PACKER_DIRECTORY="${XDG_DATA_HOME:-$HOME/.local/share}/nvim/site/pack/packer/start/packer.nvim"

  if ! [ -d "$PACKER_DIRECTORY" ]; then
    git clone "https://github.com/wbthomason/packer.nvim" "$PACKER_DIRECTORY"
  fi
}

bootstrap_plugins() {
  nvim -u NONE \
    +'autocmd User PackerComplete quitall' \
    +'lua require("plugin_specification")' \
    +'lua require("packer").sync()'
}

install_packer
bootstrap_plugins
```

</details>

I can open a PR to add these snippets to the README if you find them useful.

2deth:
[...]

PR #674, opened 2021-10-28
feat: add sync_then to accept a callback when sync is complete

michaelbeaumont:
With this function, we can solve https://github.com/wbthomason/packer.nvim/issues/198 a little nicer:

plugins.lua
```
return function(continue)
	local install_path = vim.fn.stdpath("data") .. "/site/pack/packer/start/packer.nvim"
	local packer_bootstrap

	if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
		packer_bootstrap = vim.fn.system({
			"git",
			"clone",
			"--depth",
			"1",
			"https://github.com/wbthomason/packer.nvim",
			install_path,
		})
		vim.cmd("packadd packer.nvim")
	end

	return require("packer").startup(function(use)
		-- plugins here
                -- this could also be abstracted out
		if packer_bootstrap then
			require("packer").sync_then(continue)
		else
			continue()
		end
	end)
end
```
init.lua
```
local init = function()
	-- setup everything else
end

require("plugins")(init)
```

Might make sense to add a callback to every async function on the `packer` module.
What do people think?

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -566,7 +566,7 @@ end
 --  - Clean stale plugins
 --  - Install missing plugins and update installed plugins
 --  - Update helptags and rplugins
-packer.sync = function(...)
+local function sync_then(callback, ...)
   local log = require_and_configure 'log'
   log.debug 'packer.sync: requiring modules'
   local plugin_utils = require_and_configure 'plugin_utils'
@@ -648,8 +648,12 @@ packer.sync = function(...)
     local delta = string.gsub(vim.fn.reltimestr(vim.fn.reltime(start_time)), ' ', '')
     display_win:final_results(results, delta, opts)
     packer.on_complete()
+    callback()
   end)()
 end
+
+packer.sync = function(...) return sync_then(function() end, ...) end
+packer.sync_then = sync_then
 
 packer.status = function()
   local async = require('packer.async').sync


wbthomason:
Thanks! I suppose the intent here is to avoid needing to make an autocommand on `PackerComplete`?

michaelbeaumont:
Yeah exactly, just a more Lua-first/cleaner way of configuring things I suppose.

wbthomason:
Here's where I wish Neovim had a single standardized `async` library/abstraction, so the user could just `await` here/similar. In lieu of that, I think we could maybe add a convenience function that creates a `PackerComplete` autocommand for you, to avoid duplication of functionality. Your example would be slightly longer, but I think still fairly clean:
```lua
return function(continue)
	local install_path = vim.fn.stdpath("data") .. "/site/pack/packer/start/packer.nvim"
	local packer_bootstrap

	if vim.fn.empty(vim.fn.glob(install_path)) > 0 then
		packer_bootstrap = vim.fn.system({
			"git",
			"clone",
			"--depth",
			"1",
			"https://github.com/wbthomason/packer.nvim",
			install_path,
		})
		vim.cmd("packadd packer.nvim")
	end

	return require("packer").startup(function(use)
		-- plugins here
                -- this could also be abstracted out
		if packer_bootstrap then
                        require('packer').run_on_complete(continue)
			require("packer").sync()
		else
			continue()
		end
	end)
end
```

Another more personalized option is to override the existing `packer.on_complete` function, wrapping it like:
```lua
local old_on_complete = packer.on_complete
local function continue_on_complete()
  old_on_complete()
  continue()
end

packer.on_complete = continue_on_complete
```

This snippet is a little more verbose, but it's at least Lua-only.

PR #658, opened 2021-10-18
Show full URLs for PackerSync and PackerStatus. Fixes #542

dsully:

PR #682, opened 2021-11-03
Fix: Don't duplicate URL in plugin info display on toggle

### LINKED ISSUE #681: URL line being repeated if hitting enter on it
- `nvim --version`: 

NVIM v0.6.0-dev+553-g7899c4099
Build type: Release
LuaJIT 2.1.0-beta3

- `git --version`:

git version 2.26.3

- Operating system/version:

Fedora 32

- Terminal name/version:

xfce4-terminal 0.8.10 (Xfce 4.14)

### Steps to reproduce

1 - `:PackerUpdate`
2 - Move cursor  on URL line
3 - Hit `Enter`

### Actual behaviour

URL line is repeated, like `yyp` would have done. If done 5 times, **display.lua:210: Index out of bounds**  error message is shown

### Expected behaviour

Just do nothing (not sure about this, I don't know if this line has to do something instead of showing the url used)

image

image

wbthomason:
After #658, we display plugin URLs in the status window after an update. Unfortunately, this broke the simple heuristic used for finding the nearest plugin name to determine which set of info to toggle, because plugin names now appear in two places. This PR fixes that bug by checking for the ' URL:' prefix to a line, and additionally fixes a minor bug wherein we were not restoring cursor position after toggling.

@n3wborn: Can you please confirm that this closes #681 for you?

### DIFF
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -606,6 +606,7 @@ local display_mt = {
       return
     end
 
+    local current_cursor_pos = api.nvim_win_get_cursor(0)
     local plugin_name, cursor_pos = self:find_nearest_plugin()
     if plugin_name == nil then
       log.warn 'No plugin selected!'
@@ -622,6 +623,8 @@ local display_mt = {
     else
       log.info('No further information for ' .. plugin_name)
     end
+
+    api.nvim_win_set_cursor(0, current_cursor_pos)
   end,
 
   diff = function(self)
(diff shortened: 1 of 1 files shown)


n3wborn:
> After #658, we display plugin URLs in the status window after an update. Unfortunately, this broke the simple heuristic used for finding the nearest plugin name to determine which set of info to toggle, because plugin names now appear in two places. This PR fixes that bug by checking for the ' URL:' prefix to a line, and additionally fixes a minor bug wherein we were not restoring cursor position after toggling.
> 
> @n3wborn: Can you please confirm that this closes #681 for you?

Ok for me :)

Thanks 👍🏽

wbthomason:
Great, thanks for checking it out!

PR #711, opened 2021-11-28
Fix #704: respect log level config

### LINKED ISSUE #704: Log level configuration
I'm very new to packer, and was trying to debug some behavior I wasn't sure about (and still am not...)
Since there's a nice logger builtin I naturally wanted to use it, but quickly realized it doesn't respect the log = { level } configuration. It stays on the default "warn" setting.
I could be totally wrong here, as I'm new to both packer and Lua, and still learning and trying to understand how it works.
However, it seems to me that since #331, log configuration takes place in 'startup' before 'init', and thus doesn't respect the log configuration.
If needed, I can provide more information, but for now it seems pretty straight forward so I'm posting this as is.

gadkadosh:
A quick fix to this issue. Push using require_and_configure for 'log' in packer.startup to after loading config, so that log level is respected.
This means requiring 'log' twice though - there might be a better way...

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -789,9 +789,7 @@ end
 packer.plugin_complete = function(lead, _, _)
   local completion_list = vim.tbl_filter(function(name)
     return vim.startswith(name, lead)
-  end, vim.tbl_keys(
-    _G.packer_plugins
-  ))
+  end, vim.tbl_keys(_G.packer_plugins))
   table.sort(completion_list)
   return completion_list
 end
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Wow, sorry for letting this slip for so long! Thanks - I think that, given that Lua caches `require`s (so this is basically just configuring `log` on the second require) this is a fairly clean solution for `startup`. We'll still want to think about what to do for non-`startup` configs, but this is a good start.

PR #769, opened 2022-01-11
feat(compile) adds possibility to manually set if a plugin needs BufRead

AckslD:
I had an issue with a plugin that was lazy loaded in insert mode and triggered a BufRead event which caused some other issues. For the plugin in question I actually don't need it BufRead to be triggered. So this PR adds the option to manually set if BufRead is needed instead of relying on the detection of folders (default)


wbthomason (approved):
LGTM, thanks!

PR #779, opened 2022-01-16
feat(load) delay BufRead if inside packer_compile

### LINKED ISSUE #778: Execute BufRead later
When `packer` sees that a plugin requires `BufRead` (ie by detecting eg a `ftdetect` folder) it does `doautocmd BufRead`.

This happens sometimes too early. For example I had an issue where LSPs didn't start automatically when opening a file using nvim (by doing `nvim some_file`). After a very long time of debugging and trying do understand what was going on, I realised that the following was happening:

In `packer_compiled` when calling all the `setup`/`config` for various plugins the `config` for `nvim-gps` was called before `lspconfig`. `nvim-gps` in turn `require` `nvim-treesitter` by module and I have `nvim-treesitter` lazy loaded by module. So `packer` starts loading `nvim-treesitter` which has a `ftdetect` folder which makes `packer` trigger `BufRead`. However all this happens before `lspconfig` is loaded and the `FileType` autocmds that `lspconfig` adds are therefore never executed.

I think that the better approach would maybe be to:
* keep track of when `packer_compiled` is entered/exited.
* if `needs_bufread` is true while executing `packer_compiled`, don't trigger `BufRead`.
* if by the end of `packer_compiled` at least one of the plugins `needs_bufread` then trigger it.

AckslD:
Closes #778

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -35,16 +35,17 @@ local no_errors, error_msg = pcall(function()
 
 
 local enter_packer_compile = [[
-vim.g.packer_inside_compile = true
+_G._packer = _G._packer or {}
+_G._packer.inside_compile = true
 ]]
 
 local exit_packer_compile = [[
 
-vim.g.packer_inside_compile = false
-if vim.g.packer_needs_bufread == true then
+_G._packer.inside_compile = false
+if _G._packer.needs_bufread == true then
   vim.cmd("doautocmd BufRead")
 end
-vim.g.packer_needs_bufread = false
+_G._packer.needs_bufread = false
 ]]
 
 local catch_errors = [[
@@ -105,7 +106,6 @@ local function save_profiles(threshold)
     end
   end
 
-  _G._packer = _G._packer or {}
   _G._packer.profile_output = results
 end
 ]]
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -33,6 +33,20 @@ vim.api.nvim_command('packadd packer.nvim')
 local no_errors, error_msg = pcall(function()
 ]]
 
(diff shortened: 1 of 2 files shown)


lua/packer/compile.lua
```diff
@@ -33,6 +33,20 @@ vim.api.nvim_command('packadd packer.nvim')
 local no_errors, error_msg = pcall(function()
 ]]
 
+
+local enter_packer_compile = [[
+vim.g.packer_inside_compile = true
```
wbthomason, line 38: Nitpick: I'd prefer that this/the other `vim.g` variables currently used be switched to the `_G.packer` table for consistency.

wbthomason (changes requested):
LGTM, except for requesting a change in global scope convention.

lua/packer/compile.lua
AckslD: Sure thing, I tried to grep for this but only found `_G.packer_plugins`, is this the one you mean? Or did you mean a new table?

wbthomason: We create a `_G._packer` in e.g. line 76 of packer.lua or line 94 of compile.lua - this is meant to be the global "packer metadata" table. It's just slightly nicer than relying on variables with specific names in Vim's global scope, imo.

AckslD: Ah I see, I missed the `_` prefix of `packer`. Done in 5effb94da662e5052cd1ec3d958853efb57bcd90 :+1:

wbthomason (approved):
LGTM, thank you!

wbthomason:
Check failed with an unrelated error: `gzip: stdin: not in gzip format`

PR #780, opened 2022-01-16
Fix find_nearest_plugin in display buffer

FredericHamel:
I added a config to hold the item symbol use to list the plugins. I also improve `find_nearest_plugin` so it only returns line that contains a valid plugin item. This fixes the collapsing and expanding of plugin item in the plugins buffer.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -52,6 +52,7 @@ local config_defaults = {
     done_sym = '✓',
     removed_sym = '-',
     moved_sym = '→',
+    item_sym = '•',
     header_sym = '━',
     header_lines = 2,
     title = 'packer.nvim',
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -356,7 +356,7 @@ local display_mt = {
       local load_state = plug_conf.loaded and ''
         or vim.tbl_contains(rtps, plug_conf.path) and ' (manually loaded)'
         or ' (not loaded)'
-      local header_lines = { fmt(' • %s', plug_name) .. load_state }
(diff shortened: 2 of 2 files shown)


wbthomason:
Thanks for the PR - unfortunately, this seems to break collapsing and expanding items in the plugin buffer for me. Do you have a test case that this fixes for you?

FredericHamel:
1. Open the plugin buffer with `:PackerStatus`
2. Move cursor on a line that contains a plugin.
3. Press enter to expand the plugin
4. Move my cursor on any line under the expanded plugin. This patch will make sure that `find_nearest_plugin` will return a valid line.
5. Press enter to collapse the plugin.

The function `find_nearest_plugin` will search the first plugin above the cursor by decreasing line. It searches for the plugin from the current line to line 1. It won't  expand the first plugin if the cursor in header lines.

FredericHamel:
I improved `find_nearest_plugin` so it always find the line of the nearest plugin. The cursor is positioned on the plugin item being collapsed or expanded. This should not break the expanding or collapsing in the plugin buffer.

If it breaks the expanding and collapsing of plugin items, what exactly happen?

wbthomason:
@FredericHamel Thanks again! I've confirmed that this does fix the status display, as expected. However, it's still breaking expanding/collapsing of plugin items for me in the post-update/install status display. Try updating a plugin (you can force an update by checking out an old commit of one of your installed plugins) and, after everything completes, toggling the display of the update info. You should see warning messages saying that no plugin is selected, which is a regression.

The real issue seems to be that neither toggling function works for both use cases. We either need to make a single function that works for both, or use your fix for the status display and the original function for the post-update status.

wbthomason (approved):
LGTM, thank you!

PR #782, opened 2022-01-17
Specify `-u` option for bootstraping requested plugins.

grundic:
By providing `plugins.lua` as a startup nvim configuration, would let us to isolate plugin requirement from the rest of the configuration. As was shown in #777, some options could lead to failure of Lua initialization, therefore preventing the whole bootstrap process.


wbthomason:
Thanks! I'm unfortunately not quite sure that this does what you want - `start` plugins will still be loaded, and the `packer_compiled` file will still be sourced, no?

PR #785, opened 2022-01-19
Feature: Add option to specify use_rocks-style Luarocks in the table form of startup

wbthomason:
Fixes an oversight wherein there was no way to add rocks separate from a plugin (i.e. the `use_rocks` style) if you use the table-only form of `packer.startup`.


wbthomason:
Thanks to @clason for testing!

PR #795, opened 2022-01-22
Handle unreachable git repos without nil exception

smhc:
If a git repo is not reachable a nil exception is thrown (neovim v7 nightly)

These small changes handle this scenario without crashing out.


wbthomason (approved):
This looks good, thank you!

Issue #529, opened 2021-08-03
Add support for wildcards in tags and make contextual information available to hooks


oblitum:
As per [gitter discussion](https://gitter.im/packer-nvim/community?at=6104790ff4d1dc69f31396c7):

- vim-plug offers [wildcard support in tags](https://github.com/junegunn/vim-plug/pull/476), it would be nice to have the same in packer.nvim
- `run`, `setup` and `config` hooks could benefit of information such as current tag, commit, repository url, etc.

These features could _indirectly_ allow an user to use packer.nvim to track for repository changes as a trigger to use hooks to download and install relevant binaries from github release pages for example.

wbthomason:
Thanks! It's worth noting that `run` functions are already passed the corresponding plugin table; it should be easy to add information on the latest update.

Also, from Gitter, for notes for me when I have time/anyone else who takes it up first:

> oblitum: I can point though how vim-plug did it junegunn/vim-plug#476, and also warn regarding an issue they have with their implementation, which packer could look to avoid falling in the same trap: junegunn/vim-plug#814

PR #809, opened 2022-02-06
Tag Wildcards support

danymat:
Hello, 

Providing support for wildcards, as discussed in #529, in the effort of providing plugin managers ways to follow plugin versioning.

For example, if the plugin has these tags:
```
1.0.0
0.1.0
0.0.11
0.0.10
```

And you add `tag = "0.0.*"`, it'll follow `0.0.11`. Doing `tag = "*"` will follow latest tag. 
If the wildcard expansion did not succeed, will default to latest commit with a warning log.

Some TODO:

- [x] Installing a plugin with wildcard tag
- [x] Updating a plugin with wildcard tag
- [x] Proper error handling if wildcard did not catch a tag
- [x] Verifying consistency for plugins without tags
- [x] Verifying consistency for plugins with tags but without wildcards
- [ ] Update help file + readme

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -38,6 +38,7 @@ local config_defaults = {
       get_bodies = 'log --color=never --pretty=format:"===COMMIT_START===%h%n%s===BODY_START===%b" --no-show-signature HEAD@{1}...HEAD',
       submodules = 'submodule update --init --recursive --progress',
       revert = 'reset --hard HEAD@{1}',
+      tags_expand_fmt = 'tag -l %s --sort -version:refname',
     },
     depth = 1,
     clone_timeout = 60,
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -136,6 +136,13 @@ git.setup = function(plugin)
     '%s+'
   )
 
+  local has_wildcard = function()
+    if not plugin.tag then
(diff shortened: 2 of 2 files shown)


danymat:
This seems to work after some testing. 
Some considerations: 
- Packer seems to not know when a tag got deleted (try `git tag`, delete a tag in remote, `PackerSync`, `git tag` again)
- I don't know how to throw an error if a wildcard has not found anything (or maybe we should fallback to latest commit branch)

wbthomason:
If the wildcard finds no tags, then I think it's reasonable to print a warning and choose the latest commit instead. I'll try to give this a test and see if I can figure out why packer doesn't handle deleted tags appropriately, but I think this is ok to merge without that fix, once we have the error case handled.

Thanks for this!

danymat:
Thanks for your input ! 
Do you have any particular function I can call to print the warning ?

wbthomason:
Yup - if you use the `warn` function from `packer.log`, it should be handled appropriately.

danymat:
Hello, you can review and try the PR !

wbthomason (approved):
LGTM and works like a charm in testing. Thank you!

danymat:
Thanks for the merge ! 
Maybe we should write in the readme that the wildcard support is here @wbthomason

Issue #750, opened 2021-12-29
Bootstrapping doesn't work


SimonMandlik:
- `nvim --version`: v0.6.0
- `git --version`: 2.32.0 (Apple Git-132)
- Operating system/version: Mac OS Monterey 12.1
- Terminal name/version: iterm2 3.4.15

Bootstrapping suggested in README throws an error on a fresh installation.

### Steps to reproduce

`~/.config/nvim/init.vim`:
```vim
lua require('plugins')
```

`~/.config/nvim/lua/plugins.lua`:
```lua
local fn = vim.fn
local install_path = fn.stdpath('data')..'/site/pack/packer/start/packer.nvim'
if fn.empty(fn.glob(install_path)) > 0 then
  packer_bootstrap = fn.system({'git', 'clone', '--depth', '1', 'https://github.com/wbthomason/packer.nvim', install_path})
end

require('packer').startup(function(use)
  use 'wbthomason/packer.nvim'

  -- Automatically set up your configuration after cloning packer.nvim
  -- Put this at the end after all plugins
  if packer_bootstrap then
    require('packer').sync()
  end
end)
```

`~/.local/share/nvim` and `~/.config/nvim/pluginn` are both empty

### Actual behaviour

```
Error detected while processing /Users/simon.mandlik/.config/nvim/init.vim:
line   10:
E5108: Error executing lua /Users/simon.mandlik/.config/nvim/lua/plugins.lua:7: module 'packer' not found:
        no field package.preload['packer']
        no file './packer.lua'
        no file '/usr/local/Cellar/luajit-openresty/2.1-20210510/share/luajit-2.1.0-beta3/packer.lua'
        no file '/usr/local/share/lua/5.1/packer.lua'
        no file '/usr/local/share/lua/5.1/packer/init.lua'
        no file '/usr/local/Cellar/luajit-openresty/2.1-20210510/share/lua/5.1/packer.lua'
        no file '/usr/local/Cellar/luajit-openresty/2.1-20210510/share/lua/5.1/packer/init.lua'
        no file './packer.so'
        no file '/usr/local/lib/lua/5.1/packer.so'
        no file '/usr/local/Cellar/luajit-openresty/2.1-20210510/lib/lua/5.1/packer.so'
        no file '/usr/local/lib/lua/5.1/loadall.so'
stack traceback:
        [C]: in function 'require'
        /Users/simon.mandlik/.config/nvim/lua/plugins.lua:7: in main chunk
[...]

PR #814, opened 2022-02-10
Add packer itself to bootstrapping in README.md

kad-tijhuo:
When you copy/paste the bootstrapping code from the README packer will install but try to remove itself again right away. No matter if you answer yes or no to the question you get a 'display_win' error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,7 @@ if fn.empty(fn.glob(install_path)) > 0 then
 end
 
 return require('packer').startup(function(use)
+  use 'wbthomason/packer.nvim'
   -- My plugins here
   -- use 'foo1/bar1.nvim'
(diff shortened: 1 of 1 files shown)


wbthomason:
Are you referring to #750? The bootstrap example should work as is, but it seems there may have been a regression in Neovim core since v0.6.1.

craigmac:
Still happens on neovim HEAD 0.8.0-dev-917-gbaaaa1f2b, the README.md should include `use { 'wbthomason/packer.nvim' }` in order to work without error. I think it's because if you have no packages defined in the `startup` call, clean tries to remove the `pack/packer/start/packer.nvim` directory as an 'unused' package, which it isn't, in this case.

PR #818, opened 2022-02-13
feat: add option for quiet compilation

mikatpt:
Somewhat trivial case, but I run `packer.compile` at startup and wish I could turn off the notification. Thanks for all your hard work!

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -333,8 +333,10 @@ end)
 
 --- Hook to fire events after packer compilation
 packer.on_compile_done = function()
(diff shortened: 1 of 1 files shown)


wbthomason:
Thanks! I agree that this would be nice to disable - I might prefer changing the level of the `vim.notify` call to `DEBUG`, to avoid adding this new argument. Would that work for you?

mikatpt:
Good suggestion - that works perfectly for me, made the change!

wbthomason (approved):
LGTM, thanks.

PR #846, opened 2022-03-01
fix: move snapshot configuring to packer.init()

### LINKED ISSUE #845: `packer.snapshot` not configured when not using `packer.startup()`
- `nvim --version`: NVIM v0.6.0
- `git --version`: git version 2.35.0
- Operating system/version: macOS Monterey
- Terminal name/version: iTerm2

### Steps to reproduce
Config:
```lua
local packer = require("packer")
packer.init({ ... anything here ...})
packer.use("wbthomason/packer.nvim")
-- Other plugins etc...
```

Execute command like `PackerRollback <a snapshot name>`.

### Actual behaviour

Error that `config.git` is null.

### Expected behaviour

Snapshot is rolled back

### Investigation

So, I started poking around. I've noticed that for one, the `packer.snapshot()` function and the `packer.rollback()` functions, the snapshot module is required by using `require 'packer.snapshot'` rather than `require_and_configure 'snapshot'`. This means that they never receive the user or default configuration resulting in the null reference. This would be an easy fix.

However, `PackerDelete` is executed directly from the binding as `lua require('packer.snapshot').delete(<f-args>)`, not providing an opportunity to configure. This could probably be refactored to be similar to snapshot and rollback.

This is not an issue with `packer.startup()` because that function calls `require_and_configure 'snapshot'` already.

So an alternative fix could be to add `require_and_configure 'snapshot'` to `packer.init()` and probably removing it from `packer.startup()` because that calls `packer.init()` already.

I'm happy to make the patch, but I wasn't sure what approach would be preferred as I haven't contributed to this project yet.

ViViDboarder:
Fixes #845

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -123,6 +123,9 @@ packer.init = function(user_config)
 
   local plugin_utils = require_and_configure 'plugin_utils'
   plugin_utils.ensure_dirs()
+
+  require_and_configure 'snapshot'
+
   if not config.disable_commands then
     packer.make_commands()
   end
@@ -992,7 +995,6 @@ packer.startup = function(spec)
     end
   end
 
-  require_and_configure 'snapshot' -- initialize snapshot config
   if config.snapshot ~= nil then
     packer.rollback(config.snapshot)
   end


smhc:
Will this also fix the fact :PackerInstall doesn't honour the snapshot file?

Or is the expectation that PackerInstall always ignores the snapshot file and you need to perform an explicit rollback after installing?

ViViDboarder:
This would have no impact on that and I’m not sure if that’s intended or not, so I don’t plan on including that in this patch.

ViViDboarder:
@wbthomason I see there is work being done on a v2 branch, would it be better for me to rebase this there? It would be good to get this bugfix merged into the current major release as well though.

wbthomason:
Sorry for the delay. Probably best to merge this into the current `master`, then bring that into the v2 work separately. I'll try to take a review in the next few days.

ViViDboarder:
Hey @wbthomason. Did you get a chance to take a look at this diff?

ViViDboarder:
Hello. Friendly reminder any this bug fix.

ViViDboarder:
Bump. This is a two line change and should be a quick review. Let me know if there is anything I can do to make it easier like providing test output or something, or tagging another contributor for review.

ViViDboarder:
It’s a bug fix and not a breaking change. I have all the investigative detail on the issue ticket.

PR #855, opened 2022-03-13
feat: add autoremove option

dundargoc:
Automatically removes unused or disabled plugins without asking the user if set to true.


wbthomason (approved):
This is a good addition, thank you!

PR #857, opened 2022-03-14
fix(luarocks): use lower string as package name

kevinhwang91:
If package name contains any uppercase chars, PackerUpdate action always displays `✓ Installed packerName`.

reference:
- https://github.com/luarocks/luarocks/blob/8f3ce333e7a1abfc6bf0488817d9eff4a2fd443f/src/luarocks/rockspecs.lua#L121

### DIFF
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -486,10 +486,9 @@ local function ensure_rocks(rocks, results, disp)
     local to_install = {}
     for _, rock in pairs(rocks) do
       if type(rock) == 'table' then
-        to_install[rock[1]] = rock
-      else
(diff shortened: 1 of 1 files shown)


wbthomason (approved):
Thanks! I was a bit concerned that this would cause problems if the installed rock didn't have the lowercased name (i.e. when we check to see if a rock is already installed), but some testing indicates that this doesn't happen - probably because of the luarocks logic you linked.

akinsho:
So months later I've only just realised that this change breaks installing any luarocks that require arguments such as an `env` or a `server` since it no longer passes on the rest of the metadata for the installation. I'll raise a PR shortly to fix this

PR #861, opened 2022-03-23
Expose functionality to get a plugin's short name

runiq:
I'm preprocessing plugin specs at the moment (until the config merging PR lands) and, in that context, it's sometimes useful to refer to a plugin's canonical short name outside of packer. I figured why not expose the functionality in general?

If you think this isn't necessary, feel free to close this, of course. :)


wbthomason:
Thank you!

PR #872, opened 2022-04-09
patch: Use Lua autocmd API

schra:
Make use of the new API implemented in
https://github.com/neovim/neovim/pull/14661.

This API is not yet released, but will probably be released with Neovim
0.7 (see https://github.com/neovim/neovim/milestone/29). So I guess you
can merge my PR when Neovim 0.7 is released?

I tried out this diff on the Git master of Neovim.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,23 +201,15 @@ You can configure Neovim to automatically run `:PackerCompile` whenever `plugins
 [an autocommand](https://neovim.io/doc/user/autocmd.html#:autocmd):
 
 ```
-augroup packer_user_config
-  autocmd!
-  autocmd BufWritePost plugins.lua source <afile> | PackerCompile
-augroup end
+local group = vim.api.nvim_create_augroup("packer_user_config", { clear = true })
+vim.api.nvim_create_autocmd("BufWritePost", {
+  command = "PackerCompile",
+  pattern = "plugins.lua",
+  group = group,
+})
 ```
 
-This autocommand can be placed in your `init.vim`, or any other startup file as per your setup.
-Placing this in `plugins.lua` could look like this:
-
-```lua
-vim.cmd([[
-  augroup packer_user_config
-    autocmd!
-    autocmd BufWritePost plugins.lua source <afile> | PackerCompile
-  augroup end
-]])
-```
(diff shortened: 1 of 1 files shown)


clason:
This API will definitely be released with 0.7, which is scheduled for April 15. Be aware that until then, the API may still change!

I'd recommend waiting a bit before merging this to allow people (and distros) to update (this being the Easter weekend, after all).

README.md
```diff
@@ -201,23 +201,15 @@ You can configure Neovim to automatically run `:PackerCompile` whenever `plugins
...
-
-```lua
-vim.cmd([[
-  augroup packer_user_config
-    autocmd!
-    autocmd BufWritePost plugins.lua source <afile> | PackerCompile
-  augroup end
-]])
-```
+This autocommand can be placed in your `init.vim`, in `plugins.lua`, or any other startup file as per your setup.
```
clason, line 212: Should be `init.lua`, as this is Lua code.

schra: Good catch. However, I see that there are other references to `init.vim` in the readme. I guess it would be inconsistent if I used `init.lua` in this one place and the rest of the readme uses `init.vim`. I dropped the reference to `init.vim` entirely for now, I think that's probably the easiest solution.

README.md
```diff
@@ -201,23 +201,15 @@ You can configure Neovim to automatically run `:PackerCompile` whenever `plugins
 [an autocommand](https://neovim.io/doc/user/autocmd.html#:autocmd):
 
 ```
-augroup packer_user_config
-  autocmd!
-  autocmd BufWritePost plugins.lua source <afile> | PackerCompile
```
vheon, line 206: is `source <afile>` not needed anymore?

schra: I'm not familiar with what `source <afile>` does. If I remember correctly I tried to find the equivalent in the lua api and there was nothing equivalent (but maybe I remember that wrong). I have the lua code in my init.lua as I wrote in this PR and it works.

clason: There is -- Lua callbacks get passed an `args` table, with `args.file` being equivalent to `<afile>`.

See `:h nvim_create_autocmd()`.

PR #884, opened 2022-04-21
feat: print total time taken to load all plugins

melkster:


lua/packer/compile.lua
```diff
@@ -85,12 +85,20 @@ local function save_profiles(threshold)
...
       results[i] = elem[1] .. ' took ' .. elem[2] .. 'ms'
     end
+    total_time = total_time + elem[2]
   end
 
+  if threshold then
+    table.insert(results, '(Only showing plugins that took longer than ' .. threshold .. ' ms ' .. 'to load)')
+  end
+  table.insert(results, '')
+  table.insert(results, 'Total time: ' .. total_time .. 'ms')
```
wbthomason, line 100: Nitpick: can we use the `results[#results + 1]` style here?

wbthomason (changes requested):
Thanks! I think this will be a good addition; I'd just like some small changes to keep things clean and consistent before we merge.

melkster:
Thanks for your feedback! I've updated the PR accordingly :)

PR #888, opened 2022-04-25
Add `nosuf = true` param to `vim.fn.expand()` call

deathlyfrantic:
This change ensures that `output_path` doesn't get transformed into "" if `packer_compiled.lua` is in `&wildignore`.


deathlyfrantic:
I ran into a problem where Packer was throwing an error when trying to compile, and tracked it down to having added `packer_compiled.lua` to my `&wildignore` settings. Apparently Vim's `expand()` function takes `&wildingnore` into account for some reason. 🤷‍♂️

wbthomason (approved):
Wow, nice find. Thanks!

PR #893, opened 2022-04-28
chore: move default compile_path under $XDG_DATA_HOME

TheLudd:
The old location is usually under source control, which was not optimal
for generated files. The new location is more suitable for automated
output and is still in the runtimepath.


mcepl:
I don’t think this is a good idea without the automagic migration from the old place to the new one.

TheLudd:
But this is a generated file for performance purposes? If it does not exist, it will just be re-generated. Will users even notice?

mcepl:
Ah, I see. You are talking about just about `packer_compiled.lua`? Yes, that should really not be in `~/.config/nvim` tree.

mcepl (approved):
LGTM

TheLudd:
Any thoughts on this? If it looks good, can there be a merge?

mcepl (approved):

TheLudd:
Once again, this thumbed up, approved, LGTM PR is rebased and ready for merge :smile:

thetic:
https://github.com/wbthomason/packer.nvim/blob/master/README.md?plain=1#L3

PR #895, opened 2022-04-29
Added note about nvim 0.7.0 lua autocmds

FredeHoey:
neovim introduced native lua autocmds, so I though they'd be worth a mention.

I was considering adding an example with the `callback` field instead of `command` but couldn't find a lua variant of source, and doing it with `vim.cmd` would yield the same result as using command

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,8 +222,10 @@ vim.cmd([[
 Since neovim 0.7.0 you can do this in lua too
 
 ```lua
+vim.api.nvim_create_augroup('packer_user_config', { clear = true })
 vim.api.nvim_create_autocmd('BufWritePost', {
-    pattern = 'plugins.lua'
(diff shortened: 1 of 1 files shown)


yehorb:
Consider creating `augroup` with `{ clear = true }` before creating the `autocmd`. This will ensure that plugin reload autocmmand will be cleared before creation with each source. Duplicating autocommands can lead to unexpected results.

FredeEB:
@yehorb - TIL! Thanks. I updated the PR

yehorb:
> @yehorb - any further comments on this?

Hi. Looks good to me. Thank you for updating the PR.

FredeEB:
@wbthomason - Sorry for pinging you, but I cannot add any reviewers to this PR, but maybe you could give it a read?

PR #898, opened 2022-05-01
Feat: add PackerSnapshotDone autocmd

younger-1:
### Description

Add a new `PackerSnapshotDone` autocmd which will be fired after `PackerSnapshot`  command.
Why? As https://github.com/wbthomason/packer.nvim/pull/370#issuecomment-1051643004 said, the snapshot file generated is not ordered or stable and is not ideal for checking into source control and easily seeing what changed. Before `packer` could expand and sort the generated  snapshot file, this is needed for more smooth experience of packer's snapshoot feature.

usage example:

```vim
autocmd User PackerSnapshotDone lua require('utils').sort_snapshot_file()
```

### Test

I have test the commits of the pull request in Archlinux-WSL and Windows 10 and in both environments the autocmd works well.

see usecase here <https://github.com/younger-1/nvim/blob/d0ce631b41644e841a25bc7eac0b053a288ce0fe/lua/young/plugin-loader.lua#L131-L132>

### DIFF
--- a/README.md
+++ b/README.md
@@ -562,6 +562,7 @@ require knowing when the operations are complete, you can use the following `Use
 
 - `PackerComplete`: Fires after install, update, clean, and sync asynchronous operations finish.
 - `PackerCompileDone`: Fires after compiling (see [the section on compilation](#compiling-lazy-loaders))
+- `PackerSnapshotDone`: Fires after snapshot done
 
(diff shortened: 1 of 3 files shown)


younger-1:
@wbthomason Do you have any feedback for this feat? Thanks!

PR #902, opened 2022-05-03
Don't close Neovim if there's only packer's window

melkster:
`display.close` previously closed Neovim if there were no other windows

### DIFF
--- a/lua/packer/display.lua
+++ b/lua/packer/display.lua
@@ -876,7 +876,10 @@ display.status = { running = false, disp = nil }
 --- Close a display window and signal that any running operations should terminate
 display.quit = function()
   display.status.running = false
-  vim.fn.execute('q!', 'silent')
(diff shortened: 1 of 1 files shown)


wbthomason:
Thanks, and sorry for taking so long to get to this! Do you have an easy reproduction example? I haven't encountered this behavior before.

Separately, though I think it's probably wise to move to `api.nvim_win_close` regardless, I would feel more comfortable if we stored and used the `display` window ID rather than relying on the display window being the current window (i.e. by passing `0`)

melkster:
> Do you have an easy reproduction example?

1. Open a Packer window for instance using `:PackerStatus`
2. Close all other windows using `<C-w><C-o>`
3. Close Packer window by pressing `q`

melkster:
> I would feel more comfortable if we stored and used the `display` window ID

I'm not sure exactly how you want this to be implemented, but feel free to add a commit that fixes this and/or elaborate :)

PR #904, opened 2022-05-05
Fix code duplication in major operations

messier102:
There is a ton of code duplication between major Packer operations (install, update, sync, etc). This is an effort to refactor it into something maintainable.


messier102:
Oops, I meant to open this in my fork. My bad. :sweat_smile:

PR #905, opened 2022-05-07
fix(luarocks): update hererocks url

kevinhwang91:
latest branch look like deprecated


wbthomason (approved):
Good catch, thank you! Probably, we should move away from `hererocks` if possible or at least use a tagged version....

PR #907, opened 2022-05-08
Readme: fix code syntax in src block

j-steinbach:
Added a missing `--` before a comment. This has been annoying me for a long time.

(The `string/list` in line 400 should probably also be turned into a `string or list`, but that would break the indent of the comment cascade..)

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,7 +398,7 @@ use {
   module = string or list      -- Specifies Lua module names for require. When requiring a string which starts
                                -- with one of these module names, the plugin will be loaded.
   module_pattern = string/list -- Specifies Lua pattern of Lua module names for require. When
-  requiring a string which matches one of these patterns, the plugin will be loaded.
+                               -- requiring a string which matches one of these patterns, the plugin will be loaded.
 }
 ```
 


wbthomason:
Thanks! This is also fixed in #961 - as I commented there, do you think just removing the newline would be acceptable? It seems like a simpler fix to me if it doesn't cause unpleasant scrolling on the rendered README.

j-steinbach:
I used the style that was already in the document (line 399), but choose whichever you prefer :)

This PR:

```lua
                               -- with one of these module names, the plugin will be loaded.
  module_pattern = string/list -- Specifies Lua pattern of Lua module names for require. When
                               -- requiring a string which matches one of these patterns, the plugin will be loaded.
}
```

No newline:

```lua
                               -- with one of these module names, the plugin will be loaded.
  module_pattern = string/list -- Specifies Lua pattern of Lua module names for require. When requiring a string which matches one of these patterns, the plugin will be loaded.
}
```

Having to scroll to the side is a _(minor)_ annoyance, though. And there is a [reason](https://www.researchgate.net/publication/234578707_Optimal_Line_Length_in_Reading--A_Literature_Review) why newspapers limit their paragraph width to a limited amount of characters, even though they have way more space available.

lewis6991 (approved):

PR #922, opened 2022-05-28
Add packadd command to bootstrapping example

Supreeeme:
Without the `packadd`, on first download packer won't be sourced, throwing a bunch of errors when trying to use the packer commands.


wbthomason:
This was fixed in #953, but thank you!

PR #957, opened 2022-07-03
(feat)luarocks: make rocks path configurable; fixes #672

### LINKED ISSUE #672: Allow the customization of rocks_path
It's possible to customize the `package_root` and the `compile_path`, but the `rocks_path` is [hard coded](https://github.com/wbthomason/packer.nvim/blob/797f15afd80dcfe213d421e969f9f5f62af3a728/lua/packer/luarocks.lua#L44).

Feature: Add an option that allows the user to customize the `rocks_path`.

msva:
Well, actually, it makes all paths configurable (just in case), but main reason was to make rocks path configurable, as I have `~/.cache` on tmpfs (and it is not persistent across reboots. Moreover, actually, `cache_dir` by itself, wherever it placed, is not guaranteed to be persistent). So, it was impossible for me to properly use rocks functionality with hardcoded rocks path...

### DIFF
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -1,3 +1,5 @@
+-- luacheck: globals vim
+
 -- Add support for installing and cleaning Luarocks dependencies
 -- Based off of plenary/neorocks/init.lua in https://github.com/nvim-lua/plenary.nvim
 local a = require 'packer.async'
@@ -10,15 +12,12 @@ local fmt = string.format
 local async = a.sync
 local await = a.wait
 
-local config = nil
-local function cfg(_config)
-  config = _config.luarocks
-end
 local function warn_need_luajit()
   log.error 'LuaJIT is required for Luarocks functionality!'
 end
 
-local lua_version = nil
+local lua_version
+
 if jit then
   local jit_version = string.gsub(jit.version, 'LuaJIT ', '')
   lua_version = { lua = string.gsub(_VERSION, 'Lua ', ''), jit = jit_version, dir = jit_version }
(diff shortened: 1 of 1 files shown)


lua/packer/luarocks.lua
```diff
@@ -40,11 +36,34 @@ else
   }
 end
 
-local cache_path = vim.fn.stdpath 'cache'
-local rocks_path = util.join_paths(cache_path, 'packer_hererocks')
-local hererocks_file = util.join_paths(rocks_path, 'hererocks.py')
-local hererocks_install_dir = util.join_paths(rocks_path, lua_version.dir)
-local shell_hererocks_dir = vim.fn.shellescape(hererocks_install_dir)
+local config = nil
+local cache_path = nil
```
wbthomason, line 40: Slight nitpick: if we're making these externally configurable, let's group them together in a table to make them more manageable within this file. Really, I suppose they should just be fields in the `config` table, and we should use `vim.tbl_extend` with the appropriate merge behavior.

wbthomason (changes requested):
Thanks for this; it's a good improvement. I would like to request a small change: the configuration logic can be cleaned up a bit by making the paths fields in the `packer.luarocks` `config` table, and using `vim.tbl_extend`.

wbthomason:
Also sorry for taking so long to get to this! I haven't had much time/energy for `packer` lately.

lua/packer/luarocks.lua
msva: Well, I try to invent something here, but it wouldn't be so easy: most parameters requires values of previously-defined value.

Current behaviour is just to check if it is defined in `config`, and use defined value if it is, and some default value if don't. And, as it is executed line-by-line, if some "early" parameter was defined in config, it will affect all next-defined things.

And if we'll just use a table, this magic won't work, as we cant call neither other fields in the table of "defaults", nor values from not-yet-merged resulting table.

So, I'll still need to check values from _config, or invent some other way.

// by the way, I found that `cache_path` here is never used in anu way except being a part of `rocks_path`, so I removed it as useless.

Hope, you'll not object that change.

msva:
Well, I try to invent something here, but it wouldn't be so easy: most parameters requires values of previously-defined value.

Current behaviour is just to check if it is defined in config, and use defined value if it is, and some default value if don't. And, as it is executed line-by-line, if some "early" parameter was defined in config, it will affect all next-defined things.

And if we'll just use a table, this magic won't work, as we cant call neither other fields in the table of "defaults", nor values from not-yet-merged resulting table.

So, I'll still need to check values from _config, or invent some other way.

// by the way, I found that cache_path here is never used in anu way except being a part of rocks_path, so I removed it as useless.

Hope, you'll not object that change.

msva:
@wbthomason ping? (direct ping in case you're overwhelmed by notifications)

PR #961, opened 2022-07-06
Fix wrong format

CharlotteFallices:
It added a newline at the end but not used a multiline comment :(


wbthomason:
Thanks for catching this! Do you think it would work to instead fix this by removing the newline? That seems a bit cleaner to me.

wbthomason (commented):
LGTM, thanks! (whoops, only selected comment, not approve...)

wbthomason (approved):
LGTM, thanks!

PR #963, opened 2022-07-07
Rename variable on bootstraping script

eduardoarandah:
`packer_bootstrap` sounds very "oficial", at first I thought it had special properties.

- renamed to `packer_just_installed`
- made it local (lsp complains that global vars should be uppercase)
- added a comment, so future me don't hate me


kuator:
Doesn't sound that "official" to me.

wbthomason:
Thank you for the suggestion! I think I'd prefer to leave the name as it is, since (1) I don't think the current name is too "official" and (2) this is just example code, which users can modify as they will.

PR #974, opened 2022-07-16
fix(luarocks): rock metadata not passed in install

akinsho:
PR #857 introduced a change which prevents the full data from a rock
from being used in its installation, e.g.

```lua
{
  "lyaml", server = "http://server", env = {YAML_DIR = "/usr/local"}
}
-- Is changed to
{"lyaml"}
```

Was no longer being passed as each rock was being transformed to just
it's name with not metadata passed on during the case change

@kevinhwang91 can you please check if your usecase is still satisfied with the change in case of the rocks. If so I'll merge this unless @wbthomason steps in

### DIFF
--- a/lua/packer/luarocks.lua
+++ b/lua/packer/luarocks.lua
@@ -486,9 +486,10 @@ local function ensure_rocks(rocks, results, disp)
     local to_install = {}
     for _, rock in pairs(rocks) do
       if type(rock) == 'table' then
-        rock = rock[1]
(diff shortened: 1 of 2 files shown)


kevinhwang91:
Run `PackerUpdate` with `use_rocks {'lrexlib-PCRE2'}`  will reinstall `lrexlib-pcre2` and  show:
```log
 ✓ Installed lrexlib-pcre2
 ✓ Removed lrexlib-pcre2
```

PR #979, opened 2022-07-18
add `use` in quickstart

princejoogie:
adds the variable `use` in quickstart
lsp errors show up when this is not specified


wbthomason:
Noting for posterity that this is just to fix an erroneous lint complaint. It'll probably help avoid confusion, though, so it's a good idea. Thanks!

PR #989, opened 2022-07-31
fix: use escaped mappings for compiled keymaps

### LINKED ISSUE #988: Incorrect escape sign handling
- `nvim --version`: NVIM v0.7.2
- `git --version`: git version 2.37.1
- Operating system/version: macOS 12.5
- Terminal name/version: Alacritty 0.10.1

### Steps to reproduce
Run `:PackerSync` after install the latest packer (commit 213ed031490ab5318e7996a594b52131e5c1c06f)

### Actual behaviour
Mappings with `\` sign (for example `vim.keymap.set("n", "\\[", ":NvimTreeToggle<CR>", { silent = true })`) cause the following error:

```
[packer.nvim] [ERROR 09:06:46] async.lua:20: Error in coroutine: ...e/nvim/site/pack/packer/start/packer.nvim/lua/packer.lua:725: Vim(source):E5112: Error while creating lua chunk: /Users/user/.config/nvim/plugin/packer_compiled.lua:477: invalid escape sequence near '"'
```

### Expected behaviour
No error

Commit 213ed031490ab5318e7996a594b52131e5c1c06f introduced this error I believe (installing previous versions and rerunning `:PackerSync` fixes the error.

The latest version produces
```
vim.api.nvim_set_keymap("", "\[", "<cmd>lua require('packer.load')({'nvim-tree.lua'}, { keys = '\\[', prefix = '' }, _G.packer_plugins)<cr>", { noremap = true, silent = true })
```
note the second argument to `nvim_set_keymap`

akinsho:
@konart can you please test this branch to see if it fixes your issue

fixes #988

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -572,14 +572,14 @@ local function make_loaders(_, plugins, output_lua, should_profile)
     if keymap[1] ~= 'i' then
       prefix = ''
     end
-    local escaped_map_lt = string.gsub(keymap[2], '<', '<lt>')
-    local escaped_map = string.gsub(escaped_map_lt, '([\\"])', '\\%1')
+    local escaped_map = string.gsub(keymap[2], '([\\"])', '\\%1')
+    local escaped_map_lt = string.gsub(escaped_map, '<', '<lt>')
     local keymap_line = fmt(
       [[vim.api.nvim_set_keymap("%s", "%s", "<cmd>lua require('packer.load')({%s}, { keys = '%s'%s }, _G.packer_plugins)<cr>", { noremap = true, silent = true%s })]],
       keymap[1],
-      keymap[2],
-      table.concat(names, ', '),
       escaped_map,
+      table.concat(names, ', '),
+      escaped_map_lt,
       prefix == nil and '' or (", prefix = '" .. prefix .. "'"),
       keymap[3] == nil and '' or (", desc = '" .. keymap[3] .. "'")
     )
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -575,7 +575,7 @@ local function make_loaders(_, plugins, output_lua, should_profile)
     local escaped_map = string.gsub(keymap[2], '([\\"])', '\\%1')
     local escaped_map_lt = string.gsub(escaped_map, '<', '<lt>')
     local keymap_line = fmt(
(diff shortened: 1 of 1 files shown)


konart:
@akinsho It does (the mapping actually will toggle `nvim-tree`). There is however another problem (can't really tell where is the problematic symbol or whatever is).

Here is the error I'm getting if I try to use mapping `\[` to open `nvim-tree` right after nvim startup:
```
E5107: Error loading lua [string ":lua"]:1: invalid escape sequence near '''
```

The reason is lazy loading (?)
Here is how I load it:
```
	use({
		"kyazdani42/nvim-tree.lua",
		keys = { "\\[", "\\g" },
		cmd = "NvimTreeFindFileToggle",
		config = function()
			require("settings.nvim-tree")
		end,
	})
```
So we waiting for either one of two key combinations or the command.
With this fix I can run the command and then the mappings work fine. But keys defined in `use` result in the error mentioned above.

akinsho:
@konart thanks for trying this out, can you try again and see if it's resolved now, I've made some changes

konart:
@akinsho 

Okay, keymaps starting with `\` now do work.
The ones starting with `<Leader>` do not though. 

For example if I load Trouble like this:
```
	use({
		"folke/trouble.nvim",
		requires = "kyazdani42/nvim-web-devicons",
		config = function()
			require("settings.trouble")
		end,
		module = "telescope",
		cmd = { "TroubleToggle", "TroubleRefresh" },
		keys = { "<Leader>tt" },
	})
```

akinsho:
@konart are you sure it's the mapping that isn't working, since unless you have mapped trouble to also open using `<leader>tt` it will be loaded, but trouble will not open using the setup you have above?

I just tested my localleader and leader mappings, and they seem to work fine.

captainko:
I believe this PR has bugs, now I cannot lazy load any packages. Please check again.

captainko:
```lua
		use({
			"numToStr/Comment.nvim",
			keys = { { "n", "<C-_>" }, { "n", "gcc" }, { "n", "gc" }, { "v", "<C-_>" } },
			config = config("Comment"),
		})

```
this is my config

akinsho:
Ok I think I give up, I'm going to just revert all the changes

konart:
@akinsho sorry for the late reply. In my case the problem was with this part: `module = "telescope",` not with the changes to packer. I guess at some point I just forgot about the changes I've made...

I've tried @captainko keys for `Comment` too - and at least `gcc` worked fine for me.

akinsho:
No worries @konart for now I'm just going to hold off any such changes till when packer is refactored hopefully we won't be doing any more compilation so adding descriptions to lazyload maps will be much easier

Issue #1011, opened 2022-08-12
package name that contains boolean value is not displayed correctly.


xarthurx:
- `nvim --version`: 0.7.2
- Operating system/version: Windows 11

### Steps to reproduce
1. Install `true-zen`
2. Run `PackerSync`

### Actual behaviour
The name `true-zen` was not displayed correctly

### Expected behaviour
The name `true-zen` is  displayed correctly

### packer files
Please see img below. I believe this is a trivial case that can be easily fixed.
![image](https://user-images.githubusercontent.com/1921878/184306212-62a75a01-7d78-443e-81c9-749cdae14d1e.png)

EdenEast:
Issue is that `packerBool` was getting checked after `packerPackageName`.

PR #1012, opened 2022-08-12
fix(display): Apply package name after bool check

EdenEast:
Screen Shot 2022-08-12 at 11 31 47 AM

Resolves: #1011


akinsho (approved):
LGTM 👍🏿 thanks

Issue #944, opened 2022-06-21
moving a local plugin to another folder doesn't work, packersync wont update the symlink to point at the new folder


teto:
- `nvim --version`: v0.8.0-dev
- Operating system/version: nixos
- Terminal name/version: kitty

### Steps to reproduce

Configure packer to use a local plugin, e.g.,:
`use "/home/teto/rest.nvim"`
run PackerSync and all.
Change the path to for instance `use "/home/teto/neovim/rest.nvim"` (in my case I renamed /home/teto/rest.nvim to /home/teto/meovim/rest.nvim) . Restart rerun packersync and now you get
` ✗ Failed to install /home/teto/neovim/rest.nvim`
in the site folder I have a dangling symlink to  /home/teto/rest.nvim.

### Expected behaviour
I would expect packer to update the symlink ?

My solution so far has been to remove the symlink in .../pack/start/rest.nvim manually. PackerSync then works.

morphykuffour:
How did you get packer to work on nixos

teto:
@morphykuffour my setup is complex but here you go https://github.com/teto/home/blob/49f0e7a4e5e7770ba3619968ce6e3f7cf234de46/hm/profiles/neovim.nix#L334 (i use home manager)

EdenEast:
This is because local plugins have a [stub] update function. The update function should check if the current to links `to` the desired `from`.

[stub]: https://github.com/wbthomason/packer.nvim/blob/851c62c/lua/packer/plugin_types/local.lua#L47-L51

PR #1015, opened 2022-08-15
fix(update): Validate local symlinks on update

EdenEast:
If a local plugin is installed and then the path is moved this was not
checked for in the local update function. This change checks if the
current dest link points to the desired src. If not then the link is
re-created.

Resolves: #944

### DIFF
--- a/lua/packer/plugin_types/local.lua
+++ b/lua/packer/plugin_types/local.lua
@@ -15,7 +15,7 @@ end
 -- Due to #679, we know that fs_symlink requires admin privileges on Windows. This is a workaround,
 -- as suggested by @nonsleepr.
 
-local symlink_fn unlink_fn
+local symlink_fn
 if util.is_windows then
   symlink_fn = function(path, new_path, flags, callback)
     flags = flags or {}
--- a/lua/packer/plugin_types/local.lua
+++ b/lua/packer/plugin_types/local.lua
@@ -26,10 +26,16 @@ else
 end
 
 local symlink = a.wrap(symlink_fn)
+local unlink = a.wrap(vim.loop.fs_unlink)
+
+local function strip_trailing_sep(path)
+  local res, _ = string.gsub(path, util.get_separator() .. '$', '', 1)
+  return res
+end
 
 local function setup_local(plugin)
-  local from = plugin.path
-  local to = plugin.install_path
+  local from = strip_trailing_sep(plugin.path)
+  local to = strip_trailing_sep(plugin.install_path)
 
   local plugin_name = util.get_plugin_full_name(plugin)
   plugin.installer = function(disp)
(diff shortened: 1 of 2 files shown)


lua/packer/plugin_types/local.lua
```diff
@@ -44,8 +50,17 @@ local function setup_local(plugin)
     end)
   end
 
-  plugin.updater = function(_)
+  plugin.updater = function(disp)
     return async(function()
+      disp:task_update(plugin_name, 'checking symlink...')
+      local resolved_path = vim.fn.resolve(to)
+      vim.pretty_print { to = to, from = from, res = resolved_path }
```
akinsho, line 57: Is this supposed to be here or left in by accident?

EdenEast: Oh good catch that was there for validation while working on the change.

EdenEast: Has been removed and the branch has been rebased

lua/packer/plugin_types/local.lua
```diff
@@ -44,8 +50,16 @@ local function setup_local(plugin)
...
 
-  plugin.updater = function(_)
+  plugin.updater = function(disp)
     return async(function()
+      disp:task_update(plugin_name, 'checking symlink...')
+      local resolved_path = vim.fn.resolve(to)
+      if resolved_path ~= from then
+        disp:task_update(plugin_name, 'updating symlink...')
+        local r = await(unlink(to)):and_then(symlink(from, to, { dir = true }))
+        return r
```
wbthomason, line 60: Nitpick: we should either define `r` as `result.ok()` outside of this conditional or just straight return the `await()` call here.

wbthomason (changes requested):
Thanks for this! I have a few small style changes to request, but otherwise LGTM.

lua/packer/plugin_types/local.lua
```diff
@@ -26,10 +26,16 @@ else
 end
 
 local symlink = a.wrap(symlink_fn)
+local unlink = a.wrap(vim.loop.fs_unlink)
```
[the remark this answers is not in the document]
EdenEast: Ya I am not sure on windows. Looking at the [luv](https://github.com/luvit/luv/blob/master/docs.md#uvfs_unlinkpath-callback), I am not sure if the workaround used for symlink would apply here.

lua/packer/plugin_types/local.lua
```diff
@@ -44,8 +50,16 @@ local function setup_local(plugin)
...
+  plugin.updater = function(disp)
     return async(function()
+      disp:task_update(plugin_name, 'checking symlink...')
+      local resolved_path = vim.fn.resolve(to)
+      if resolved_path ~= from then
+        disp:task_update(plugin_name, 'updating symlink...')
+        local r = await(unlink(to)):and_then(symlink(from, to, { dir = true }))
+        return r
+      end
+
```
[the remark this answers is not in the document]
EdenEast: Is this not handled by the install / update commands?

lua/packer/plugin_types/local.lua
EdenEast: This can be cleaned up

lua/packer/plugin_types/local.lua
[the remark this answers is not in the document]
wbthomason: Derp, good point. You're right!

wbthomason (approved):
LGTM modulo a small syntax error (the `local symlink_fn unlink_fn` line) that I just fixed. Thanks!

wbthomason:
Noting for posterity that this might cause issues on Windows, but we'll need Windows testers to be sure...

PR #1016, opened 2022-08-16
fix(ci): resolve CI failures

EdenEast:


akinsho:
@EdenEast thanks for getting this working. I'd been looking at https://github.com/rhysd/action-setup-vim for a while as quite a clean way of setting up neovim in CI, maybe one to revisit at some point, but this looks good. I think the workflow that was originally copied was quite outdated.

Issue #1017, opened 2022-08-18
Dependencies cannot use aliases


svvac:
Referencing a dependency by its alias currently doesn't work.

```lua

use {
  'some/dep',
  requires = { 'other_dep' },
}

use {
  'other/dep',
  as = 'other_dep',
}

```

Looking around, it appears the conflict checks on a new dependency are too strict. When the actual aliased module gets registered afterwards, the name conflict check against the alias prevents the setup to properly execute.

https://github.com/wbthomason/packer.nvim/blob/90b323bccc04ad9b23c971a85813a1405c7725a8/lua/packer.lua#L206-L222

The second call to `use` above warns that the alias is already used because fails check on L211. As opposed to L206, this check doesn't guard against `from_requires` to deduplicate.

In any case, it seems that the second check is redundant, as `name` always equals `plugin_spec.as`

https://github.com/wbthomason/packer.nvim/blob/90b323bccc04ad9b23c971a85813a1405c7725a8/lua/packer.lua#L199

https://github.com/wbthomason/packer.nvim/blob/90b323bccc04ad9b23c971a85813a1405c7725a8/lua/packer/util.lua#L66

I suggest dropping it entirely.

PR #1018, opened 2022-08-19
Remove redundant name check when registering a plugin

svvac:
`name` is equal to `plugin_spec.as` when the latter is set by `get_plugin_short_name()`. The conflict check against the alias is thus redundant.

Dropping this allows plugins to require aliased plugins, see #1017.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -208,19 +208,6 @@ manage = function(plugin_data)
     return
   end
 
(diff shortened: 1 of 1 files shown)


svvac:
Or, if you prefer, guarding the second check as demonstrated in eae47319af7f3807390b801e3d5d7d8dba6503b2.

Issue #1022, opened 2022-08-21
Url of plugins are not correctly displayed


glyh:
- `nvim --version`:`NVIM v0.8.0-dev-699-gcbfae548e`
- `git --version`: `git version 2.37.2`
- Operating system/version: `Linux archlinux 5.19.2-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Wed, 17 Aug 2022 14:28:00 +0000 x86_64 GNU/Linux`
- Terminal name/version: `alacritty 0.10.1 ()`

### Steps to reproduce

1. Use the given test.lua
2. PackerSync
3. Check status of plugins, specifically the URL field

### Actual behaviour
Got urls like `https://wwwhub.com/Olical/conjure`.

### Expected behaviour

The correct URL.

### packer files

<details>
<summary>Plugin specification file(s)</summary>

Post or link your plugin specification files here, if you aren't able to provide a minimal
reproducer
```lua
-- this template is borrowed from nvim-lspconfig
local on_windows = vim.loop.os_uname().version:match("Windows")

local function join_paths(...)
    local path_sep = on_windows and "\\" or "/"
    local result = table.concat({ ... }, path_sep)
    return result
end

vim.cmd([[set runtimepath=$VIMRUNTIME]])

local temp_dir
if on_windows then
    temp_dir = vim.loop.os_getenv("TEMP")
else
    temp_dir = "/tmp"
end

vim.cmd("set packpath=" .. join_paths(temp_dir, "nvim", "site"))

local package_root = join_paths(temp_dir, "nvim", "site", "pack")
local install_path = join_paths(package_root, "packer", "start", "packer.nvim")
local compile_path = join_paths(install_path, "plugin", "packer_compiled.lua")

local function load_plugins()
    -- only add other plugins if they are necessary to reproduce the issue
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            {"Olical/conjure", branch = 'develop'},


        },
        config = {
            package_root = package_root,
            compile_path = compile_path,
            git = { default_url_format = 'https://www.github.com/%s' }
        },
    })
end

if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system({ "git", "clone", 'https://www.github.com/wbthomason/packer.nvim', install_path })
end
load_plugins()
-- require("packer").sync()

vim.cmd([[ set completeopt=menu,menuone,noselect]])

```

</details>

<details>
<summary>packer log file</summary>

[...]

PR #1023, opened 2022-08-21
fix(git): remove `.git` from end url only

EdenEast:
Removes `.git` from url only if it ends with `.git`.

Resolves: #1022

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -264,7 +264,7 @@ manage = function(plugin_data)
   end
 
   -- Add the git URL for displaying in PackerStatus and PackerSync.
-  plugins[plugin_spec.short_name].url = plugin_spec.url:gsub('%.git', '')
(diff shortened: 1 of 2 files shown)


akinsho (approved):
LGTM thanks for the fix

axieax:
@akinsho @EdenEast don't think the Lua pattern is entirely correct. `%s.git` matches a space character, then any character, then "git". The expected pattern should be `%.git`. I have corrected this in #1027.

This breaks the urls detected by my plugin [urlview.nvim](https://github.com/axieax/urlview.nvim).

akinsho:
@axieax thanks for raising this and the PR, didn't notice the issue 🤦🏿‍♂️

EdenEast:
Ah thanks @axieax for the catch sorry about that, my mistake 🤦 .

axieax:
No worries :))

PR #1028, opened 2022-08-22
fix(git): use vim.endswith to check git suffix

axieax:
image

@akinsho apologies but it seems like there's still a dangling period character left behind from the previous commits. This should hopefully address it!

### DIFF
--- a/lua/packer/util.lua
+++ b/lua/packer/util.lua
@@ -86,7 +86,7 @@ util.get_plugin_full_name = function(plugin)
 end
 
 util.remove_ending_git_url = function(url)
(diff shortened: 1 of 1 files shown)


EdenEast (approved):

EdenEast (approved):
Did not know that vim has an `endswith` keeping that in mind.

axieax:
> Did not know that vim has an `endswith` keeping that in mind.

ahaha I've recently found quite a number of gems hidden in the Neovim source code, particularly in [this file](https://github.com/neovim/neovim/blob/master/runtime/lua/vim/shared.lua) :D

PR #1035, opened 2022-08-27
feat(display): allow not showing info by default when displaying results

AckslD:
This allows to opt-in for having all the infos closed by defaults in the final update results. They can still be opened as normal with `<CR>`.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -49,6 +49,7 @@ local config_defaults = {
   },
   display = {
     non_interactive = false,
+    compact = false,
     open_fn = nil,
     open_cmd = '65vnew',
     working_sym = '⟳',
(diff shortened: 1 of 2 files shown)


lua/packer.lua
```diff
@@ -49,6 +49,7 @@ local config_defaults = {
   },
   display = {
     non_interactive = false,
+    oneline_results = true,
```
akinsho, line 52: Not sure that this is the best option name, it seems very specific to the current UI implementation, but also doesn't exactly clarify what exactly might be expected to be 1 line there are lots of results in packer. I think something along the lines of compact or default_closed etc. it might be one day that we can't guarantee 1 line, and it just becomes a matter of trying to make things more concise than full output

AckslD: Good point! Changed to `compact`

wbthomason (approved):
LGTM and I like the new `compact` name. Thanks!

PR #1041, opened 2022-09-01
Fix: check if a plugin given as a requirement is a single-specification table

### LINKED ISSUE #1038: plugin with opt option install start dir
- `nvim --version`: latest
- `git --version`: latest
- Operating system/version: macos
- Terminal name/version: kitty

### Steps to reproduce

```lua
use({
  'kristijanhusak/vim-dadbod-ui',
  cmd = { 'DBUIToggle', 'DBUIAddConnection', 'DBUI', 'DBUIFindBuffer', 'DBUIRenameBuffer' },
  config = conf.vim_dadbod_ui,
  requires = {'tpope/vim-dadbod', opt = true },
})

```

### Actual behaviour

`vim_dadobd` installed in start_dir

### Expected behaviour

install in opt_dir

wbthomason:
Should fix #1038 by adding a missing check for a possible `requires` case.

@glepnir, does this work for you?

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -270,7 +270,7 @@ manage = function(plugin_data)
     -- Handle single plugins given as strings or single plugin specs given as tables
     if
       type(plugin_spec.requires) == 'string'
-      or (type(plugin_spec.requires) == 'table' and #plugin_spec.requires == 1)
+      or (type(plugin_spec.requires) == 'table' and not vim.tbl_islist(plugin_spec.requires) and #plugin_spec.requires == 1)
     then
       plugin_spec.requires = { plugin_spec.requires }
     end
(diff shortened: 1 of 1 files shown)


lua/packer.lua
```diff
@@ -267,7 +267,11 @@ manage = function(plugin_data)
   plugins[plugin_spec.short_name].url = util.remove_ending_git_url(plugin_spec.url)
 
   if plugin_spec.requires and config.ensure_dependencies then
-    if type(plugin_spec.requires) == 'string' then
+    -- Handle single plugins given as strings or single plugin specs given as tables
+    if
+      type(plugin_spec.requires) == 'string'
+      or (type(plugin_spec.requires) == 'table' and #plugin_spec.requires == 1)
```
akinsho, line 273: @wbthomason how would this behave if `requires` is `requires = {{'name/thing', config = blah}}`, seems like it would wrap the plugin in another list?. Should it be if 
```suggestion
      or (type(plugin_spec.requires) == 'table' and not vim.tbl_islist(plugin_spec.requires) and #plugin_spec.requires == 1)
```

wbthomason: You're right, that would be better. Thanks!

wbthomason:
@glepnir please let us know if this does not fix your problem, but I'm going to go ahead and merge as this is a simple enough change.

PR #1044, opened 2022-09-04
Fix: use vim.notify in packer.log

wbthomason:
`vim.notify` exists now, and is preferable to the old `echom` method of logging. Some parts of `packer` used `vim.notify` directly, while some used `log`. This led to inconsistent output.

This PR makes the `log` API use `vim.notify` internally, and moves manual uses of `vim.notify` to use the `log` API.

- Use vim.notify instead of echom to log to console
- Move manual uses of vim.notify to use the log API

### DIFF
--- a/lua/packer/log.lua
+++ b/lua/packer/log.lua
@@ -99,8 +99,13 @@ log.new = function(config, standalone)
   local console_output = vim.schedule_wrap(function(level_config, info, nameupper, msg)
     local console_lineinfo = vim.fn.fnamemodify(info.short_src, ':t') .. ':' .. info.currentline
     local console_string = string.format('[%-6s%s] %s: %s', nameupper, os.date '%H:%M:%S', console_lineinfo, msg)
-
-    vim.notify(string.format([[[%s] %s]], config.plugin, console_string), vim.log.levels[level_config.name:upper()])
+    -- Heuristic to check for nvim-notify
+    local is_fancy_notify = type(vim.notify) == 'table'
+    vim.notify(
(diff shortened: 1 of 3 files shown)


wbthomason:
Note that there's still a manual use of `vim.notify` in the compiled code. I left this because (1) we don't want to have to pull in `packer.log` in the compiled output and (2) all the compiled stuff will go away soon.

lua/packer/log.lua
```diff
@@ -100,18 +100,7 @@ log.new = function(config, standalone)
...
-
-    local split_console = vim.split(console_string, '\n')
-    for _, v in ipairs(split_console) do
-      vim.cmd(string.format([[echom "[%s] %s"]], config.plugin, vim.fn.escape(v, '"')))
-    end
-
-    if config.highlights and level_config.hl then
-      vim.cmd 'echohl NONE'
-    end
+    vim.notify(string.format([[[%s] %s]], config.plugin, console_string), vim.log.levels[level_config.name:upper()])
```
akinsho, line 103: For users of things like `nvim-notify` I think adding a title via the third options is a nice to have i.e. looks much better, but this is a minor aesthetic point.

akinsho (approved):
LGTM 👍🏿 nice to have everything consolidated in one place. Would be nice to also bring in the use of `vim.ui.input` for things like the prompts to remove plugins etc.

lua/packer/log.lua
wbthomason: Oh, good point. The only annoying challenge is figuring out when we can remove the leading plugin name in the message, to prevent redundancy...

wbthomason: Added a heuristic check for whether `vim.notify` is the fancy version or the built-in.

PR #1047, opened 2022-09-04
fix(update): error message when unknown plugin is provided

AckslD:
Currently if you do eg `PackerUpdate foobar` and there is no plugin `foobar` you get a bit cryptic error about accessing a `nil` value. This just adds a more clear error message.

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -521,6 +521,9 @@ packer.update = function(...)
     log.debug 'Gathering update tasks'
     await(a.main)
     update_tasks, display_win = update(plugins, installed_plugins, display_win, results, opts)
+    if update_tasks == nil then
+      return
+    end
     vim.list_extend(tasks, update_tasks)
     log.debug 'Gathering luarocks tasks'
     local luarocks_ensure_task = luarocks.ensure(rocks, results, display_win)
@@ -602,6 +605,9 @@ packer.sync = function(...)
     log.debug 'Gathering update tasks'
     await(a.main)
     update_tasks, display_win = update(plugins, installed_plugins, display_win, results, opts)
+    if update_tasks == nil then
+      return
+    end
     vim.list_extend(tasks, update_tasks)
     log.debug 'Gathering luarocks tasks'
     local luarocks_clean_task = luarocks.clean(rocks, results, display_win)
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -527,13 +527,16 @@ packer.update = function(...)
     if luarocks_ensure_task ~= nil then
       table.insert(tasks, luarocks_ensure_task)
     end
+
+    if #tasks == 0 then
+      return
+    end
+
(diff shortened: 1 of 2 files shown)


wbthomason (approved):
Thanks!

lua/packer.lua
```diff
@@ -602,6 +605,9 @@ packer.sync = function(...)
     log.debug 'Gathering update tasks'
     await(a.main)
     update_tasks, display_win = update(plugins, installed_plugins, display_win, results, opts)
+    if update_tasks == nil then
+      return
+    end
```
wbthomason, line 610: I was a bit too fast with the review... I think we do not want to return early here, but rather set `update_tasks = {}`, because of the possibility of adding Luarocks tasks:
```suggestion
if update_tasks == nil then
  update_tasks = {}
end
```

lua/packer.lua
```diff
@@ -521,6 +521,9 @@ packer.update = function(...)
...
```
wbthomason, line 526: Same comment and suggestion: to account for cases with *only* Luarocks tasks, I think we cannot exit early here.

wbthomason (changes requested):

lua/packer.lua
AckslD: Do we really want to continue at all if a user specifies an unknown plugin to update? Seems we could treat it as a critical error and abort everything?

If we would not want to abort, the creation of `display_win` would have to be changed since it will also be `nil` in this case (and used later).

wbthomason: I'm inclined to say that we should continue if there are valid tasks to complete. If anything, maybe we don't want to return early from `update(...)` - instead, we could send the error message and keep checking the remaining plugins. This would also simplify the creation of `display_win` (though, in another PR, I should really un-nest that logic; it's poorly designed right now).

My argument for continuing if there are valid tasks at all is that `packer` generally tries to make errors minimally-breaking, e.g. using `pcall` in `load` and the compiled file. This is perhaps less likely here - the only case that brings a user to this logic is passing a bad argument to a manual call to `packer.update` or to `PackerUpdate` (resp. `sync/Sync`) - but it still seems more philosophically  consistent.

I could be convinced otherwise if you disagree, though.

AckslD: Makes sense @wbthomason :+1: I changed it to not abort, but instead just skip that plugin (with a warning). Also added check if number of tasks is 0 before calling `display_win` in `update` and `sync` (`update_headline_message` and `final_results`).

wbthomason:
Thanks! I made a small tweak to check `#tasks` before we add the task metadata (the max job count and termination check function), since otherwise I think this check will always be `true`. If you don't see any obvious issues, I think we're good to merge.

AckslD:
Awesome, looks good :+1:

wbthomason (approved):

PR #1049, opened 2022-09-05
Always fully resolve symlinks

### LINKED ISSUE #1048: If the location of a local plugin is a symlink, `PackerSync` generates an error
- `nvim --version`: 0.7.2
- `git --version`: 2.37.3
- Operating system/version: macOS Monterey 12.3.1
- Terminal name/version: kitty 0.26.1

### Steps to reproduce

Starting at 2b0af3b20f3ddf18ee2d5022dd80635dbc35eb1c, if the path given to a local plugin is a symlink, the following error occurs if `PackerSync` is run and the symlink in the packer directory exists:

```
[packer.nvim] [ERROR 18:15:15] async.lua:20: Error in coroutine: ...packer/opt/packer.nvim/lua/packer/plugin_types/local.lua:56: attempt to index a nil value
```

In my setup, I'm using `fzf` installed with Homebrew, so the main path is a symlink to the "cellar" binary:

```
  use({
    '/opt/homebrew/opt/fzf',
    cond = function()
      return vim.fn.executable('fzf') == 1
    end,
  })
```

`/opt/homebrew/opt/fzf` is a symlink pointing (currently) to `/opt/homebrew/Cellar/fzf/0.33.0`.

I added some `print()` statements after [line 53 of local.lua](https://github.com/wbthomason/packer.nvim/blob/master/lua/packer/plugin_types/local.lua#L53):

```lua
      local resolved_path = vim.fn.resolve(to)
      print('From: ' .. from)
      print('To: ' .. to)
      print('Resolved: ' .. resolved_path)
```

And when the updater code is run, it gives the following output

```
From: /opt/homebrew/opt/fzf
To: /Users/bryan/.local/share/nvim/site/pack/packer/opt/fzf
Resolved: /opt/homebrew/Cellar/fzf/0.33.0
```

### Actual behaviour

An error occurs if the symlink in the packer directory exists:

```
[packer.nvim] [ERROR 18:15:15] async.lua:20: Error in coroutine: ...packer/opt/packer.nvim/lua/packer/plugin_types/local.lua:56: attempt to index a nil value
```

### Expected behaviour

If the symlink exists and points to the correct directory, nothing happens. If the symlink exists and points to the wrong directory, it is updated. If the symlink does not exist, it is created.

wbthomason:
This PR (hopefully - needs testing) fixes #1048 by consistently using `fs_realpath` to resolve local plugin targets at both install and update time.

### DIFF
--- a/lua/packer/plugin_types/local.lua
+++ b/lua/packer/plugin_types/local.lua
@@ -50,7 +50,7 @@ local function setup_local(plugin)
     return async(function()
       local r = result.ok()
       disp:task_update(plugin_name, 'checking symlink...')
-      local resolved_path = vim.fn.resolve(to)
+      local resolved_path = vim.loop.fs_realpath(to)
       if resolved_path ~= from then
         disp:task_update(plugin_name, 'updating symlink...')
         r = await(unlink(to)):and_then(symlink(from, to, { dir = true }))
--- a/lua/packer/plugin_types/local.lua
+++ b/lua/packer/plugin_types/local.lua
@@ -30,7 +30,7 @@ local symlink = a.wrap(symlink_fn)
 local unlink = a.wrap(vim.loop.fs_unlink)
 
 local function setup_local(plugin)
-  local from = util.strip_trailing_sep(plugin.path)
+  local from = vim.loop.fs_realpath(util.strip_trailing_sep(plugin.path))
   local to = util.strip_trailing_sep(plugin.install_path)
 
   local plugin_name = util.get_plugin_full_name(plugin)


wbthomason:
@bryanforbes: when you have a chance, could you please confirm that this fixes #1048 for you? This is quickly hacked out; if it doesn't fix the issue then I'll spend more time in a day or two to properly test a fix.

wbthomason:
Thanks @bryanforbes for the report and testing! Bypassing review to merge because this is a very simple change. That said, if anything breaks on Windows with local plugins soon, this PR should be a first suspect.

bryanforbes:
Thanks for the quick fix!!

PR #1053, opened 2022-09-06
Pass hook name into plugin run functions

itmecho:
This PR adds a third argument the plugin run function, allowing the user to easily run code only on install or update of a plugin.

It should be backwards compatible as existing functions that only accept 2 arguments will still work as before.

### DIFF
--- a/lua/packer/plugin_utils.lua
+++ b/lua/packer/plugin_utils.lua
@@ -208,7 +208,7 @@ plugin_utils.load_plugin = function(plugin)
   end
 end
 
-plugin_utils.post_update_hook = function(plugin, disp)
+local run_hook = function(plugin, disp, hook_name)
   local plugin_name = util.get_plugin_full_name(plugin)
   return a.sync(function()
     if plugin.run or not plugin.opt then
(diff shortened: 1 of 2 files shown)


lua/packer/plugin_utils.lua
```diff
@@ -284,4 +284,12 @@ plugin_utils.post_update_hook = function(plugin, disp)
   end)
 end
 
+plugin_utils.post_install_hook = function(plugin, disp)
+  return run_hook(plugin, disp, 'post_install')
```
akinsho, line 288: Rather than an ever expanding list of arguments, I think when you start to pass more and more args into a function it might be time to pass in a table instead, so you scale out the inputs more easily without worrying about naked args and positioning and nil-ness i.e.
I think this should be
```lua
  return run_hook(plugin, disp, { hook_name = 'post_install' })
```

itmecho: Not sure if it needs it as it's an internal function but I've left the default `hook_name` as an empty string. Not sure what the best thing to set it to would be?

PR #1055, opened 2022-09-07
feat(packer): query what plugin to load if none if specified

AckslD:
This allows `PackerLoad` to take no arguments and in that case query the user for the plugin to load using `vim.ui.select`.

To do this a bit more cleanly I changed the command to be created with the lua api instead of `vim.cmd`. Maybe all commands should be changed to that but I think that's anyway the plan for #1042.

Also, is `packer.loader` considered part of the public api? Since this version changes it's expected argument. Could keep the same behaviour of `packer.loader` though if needed and have another internal function which is called by `PackerLoad`.


AckslD:
I guess we need to check if `nvim_create_user_command` actually exists, since 0.5 and 0.6 are still supported.

PR #1063, opened 2022-09-10
Add support for grouped keys

musjj:
If you have a lot of keys that you want to use for lazy loading, your config can become a bit hairy:

```lua
use {
  "FooBar/Baz",
  keys = {
    { "n", "a" },
    { "n", "b" },
    { "n", "c" },
    { "n", "d" },
    { "v", "e" },
    { "v", "f" },
    { "v", "g" },
    { "v", "h" },
  },
}
```

With this PR, you can now instead group your keys like this:

```lua
use {
  "FooBar/Baz",
  keys = {
    n = { "a", "b", "c", "d" },
    v = { "e", "f", "g", "h" },
  },
}
```


musjj:
Anything blocking this from getting merged?
To clarify, this PR is backwards-compatible with the original way of specifying keys.

README.md
```diff
@@ -523,7 +523,11 @@ with `bax` filetype.
 
 #### Keybindings
 
-Plugins may be lazy-loaded on the use of keybindings/maps. Individual keybindings are specified either as a string (in which case they are treated as normal mode maps) or a table in the format `{mode, map}`.
+Plugins may be lazy-loaded on the use of keybindings/maps. Individual keybindings are specified either as a string (in which case they are treated as normal mode maps) or a table in one of the following formats:
```
wbthomason, line 526: Here and in the `doc/` file, we need to change the wording a bit. It currently only talks about **individual** keybindings, but we want to say that you can specify *the whole set* of keybindings in one of these two table formats, not individual keybindings.

wbthomason (changes requested):
Thanks! Sorry for not seeing this before now. I think this is mostly good to go, but the updated language in the docs is technically incorrect, so we need to fix that before merging.

I'll have limited internet for the next week, but please @ me if you get to this and I haven't responded by Dec 21.

README.md
musjj: Sorry for the late reply! Do you think they're clearer now?

PR #1082, opened 2022-10-02
README: improve `autocmd` instructions

amarakon:
First, you can use `source` without arguments to source the current buffer. As for Lua, it is better to use actual Lua code than calling Vim from Lua.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,7 @@ You can configure Neovim to automatically run `:PackerCompile` whenever `plugins
 ```
 augroup packer_user_config
   autocmd!
-  autocmd BufWritePost plugins.lua source <afile> | PackerCompile
+  autocmd BufWritePost plugins.lua source | PackerCompile
 augroup end
 ```
 
@@ -213,12 +213,15 @@ This autocommand can be placed in your `init.vim`, or any other startup file as 
 Placing this in `plugins.lua` could look like this:
 
 ```lua
-vim.cmd([[
-  augroup packer_user_config
-    autocmd!
-    autocmd BufWritePost plugins.lua source <afile> | PackerCompile
-  augroup end
-]])
+vim.api.nvim_create_augroup('packer_user_config', {})
+vim.api.nvim_create_autocmd(
+  'BufWritePost',
+  {
+    group = 'packer_user_config',
+    pattern = 'plugins.lua',
+    command = 'source | PackerCompile'
+  }
+)
 ```
 
 ## Bootstrapping


AzP:
I was just about to supply a PR with this change, good I found that you'd already done it :)

AzP (approved):
I've tested this and it is the more modern way of doing it in nvim.

dsully:
Even more lua:

```language:lua
vim.api.nvim_create_autocmd("BufWritePost", {
  pattern = "lua/plugins.lua",
  callback = function(args)
    vim.cmd.source(args.file)
    vim.cmd.PackerCompile()
  end,
  desc = "Compile Packer plugins source.",
})
```

AzP:
Nice, looks even better.

Susensio:
> Even more lua:
> 
> ```
> vim.api.nvim_create_autocmd("BufWritePost", {
>   pattern = "lua/plugins.lua",
>   callback = function(args)
>     vim.cmd.source(args.file)
>     vim.cmd.PackerCompile()
>   end,
>   desc = "Compile Packer plugins source.",
> })
> ```

It's not working for me:
```
Error detected while processing BufWritePost Autocommands for "plugins.lua":
Error executing lua callback: /home/susensio/.config/nvim/lua/user/plugins.lua:22: attempt to
 index field 'cmd' (a function value)
stack traceback:
        /home/susensio/.config/nvim/lua/user/plugins.lua:22: in function </home/susensio/.con
fig/nvim/lua/user/plugins.lua:21>

```

dave-kennedy:
> vim.api.nvim_create_autocmd("BufWritePost", {
>   pattern = "lua/plugins.lua",
>   callback = function(args)
>     vim.cmd.source(args.file)
>     vim.cmd.PackerCompile()
>   end,
>   desc = "Compile Packer plugins source.",
> })

1. I think we want to use an autocommand group here, in case the file is sourced more than once.
2. I don't think that pattern will ever match because it isn't absolute and doesn't contain a wildcard (see `:help autocmd-pattern`).
3. Per the OP's original suggestion, we can omit the argument to `vim.cmd.source()`.

So something like this:

```
vim.api.nvim_create_autocmd('BufWritePost', {
  group = vim.api.nvim_create_augroup('packer_auto_compile', {}),
  pattern = '*/lua/plugins.lua',
  callback = 'source | PackerCompile'
})
```

I actually prefer to see a message when the command is executed. Also, `*/lua/plugins.lua` might match a lot of files unintentionally, so I use the full path in the pattern:

```
vim.api.nvim_create_autocmd('BufWritePost', {
  group = vim.api.nvim_create_augroup('packer_auto_compile', {}),
  pattern = vim.fn.expand('$HOME') .. '/dotfiles/vim/lua/plugins.lua',
  callback = function (ctx)
    vim.cmd.source()
    vim.cmd.PackerCompile()
    print('Packer compiled!')
  end
})
```

PR #1091, opened 2022-10-10
Return snapshot async fn so it can be awaited

sheharyarn:
Exactly as it says on the can. I'd like to do something like:

```lua
await(packer.snapshot(path, plugins))
-- then do some other stuff with the generated snapshot
```


sheharyarn:
Alternative to #898.

Issue #1086, opened 2022-10-06
Question


notwidow:
Hi, I have 15 plugins installed and listed by packer also. But when doing packersync or packer update it sometimes shows 17 plugins and sometimes 16 while plugins specified are 15. why is this so?

nat-418:
Please provide steps to reproduce this behavior including a minimal Neovim configuration.

notwidow:
> Please provide steps to reproduce this behavior including a minimal Neovim configuration.

well, first count how many plugins you used and specified in packer.nvim and then run a packersync.. while syncing look at how many plugins it is downloading...it must be 1 or 2 higher than the specified plugins...so why is it showing 1 or 2 more...

nat-418:
I tested with an empty init.lua except for this packer code:

```
return require('packer').startup(function(use)
  use('https://github.com/wbthomason/packer.nvim')
end)
```

snyc says 3 plugins, status says 1

notwidow:
Hey when will it be fixed?

nat-418:
whenever they accept my PR. I don't know how actively packer is maintained

notwidow:
> whenever they accept my PR. I don't know how actively packer is maintained

yeah. i see my fav plugin manager's developers are not that active...

wbthomason:
Yes, my time is very limited. Most work right now is focused on a rewrite of `packer` to fix some long-standing issues and make maintenance easier, which will mean that I can get to more issues, etc. I'm trying to make incremental quality-of-life fixes where I can in the meanwhile. Sorry that this issue has taken so long.

@nat-418: Thanks for the PR! I've played around with it, and I think that changing the constant offset isn't quite right - the extra "plugins" are displayed because we're actually printing the length of the task list, which isn't 1-to-1 with the number of plugins. I think the better fix might be to change the headline text to be "completing N / M {task type: install, update, sync} tasks". Would that be a satisfactory fix?

wbthomason:
(also @notwidow for your opinion on my proposed fix)

nat-418:
Thanks for making Packer and for the explanation. Your solution sounds good to me.

wbthomason:
Yup, understood! Sorry if I came across as angry. Check #1190 out if you get a chance; I'll merge shortly if there are no complaints, as it's a pretty trivial change.

PR #1097, opened 2022-10-16
Fix incorrect numbering of plugins in dialogs

nat-418:
Packer is currently showing the wrong number of plugins during sync, install, and update dialogs. X plugins are shown on `PackerStatus` but `PackerSync` et al. show the wrong number. This resolves issue #1086 .


wbthomason:
Closing in favor of #1190, as discussed on #1086

PR #1113, opened 2022-10-29
fix(interactive): let interactive mode work for remote clients attached to headless instances

### LINKED ISSUE #1112: Interactive mode doesn't work when using remote client attached to headless neovim instance
#507 broke interactive mode when using a remote neovim client that is attached to a headless neovim instance after it has been started.

### Steps to reproduce

```bash
$ nvim --headless --listen localhost:7777
$ neovide --remote-tcp localhost:7777 # or another client that can connect to remote headless instances
$ # run :PackerSync in neovide instance
```

### Actual behaviour

No `Packer` windows appear, even though an interactive client is attached.

### Expected behaviour

Because the `non_interactive` config is set to `false`, the `Packer` windows should appear since there is actually an interactive client attached. `api.nvim_list_uis()` returns an entry after the remote client is attached.

jdrouhard:
Fixes #1112

Instead of calculating whether the nvim instance is headless at startup and caching that (both in the config.non_interactive flag as well as a local var in the display module), just recalculate whether nvim is interactive whenever it needs to make the decision (when displaying windows or prompting the user).

### DIFF
--- a/lua/packer.lua
+++ b/lua/packer.lua
@@ -130,9 +130,6 @@ packer.init = function(user_config)
   config.pack_dir = join_paths(config.package_root, config.plugin_package)
   config.opt_dir = join_paths(config.pack_dir, 'opt')
   config.start_dir = join_paths(config.pack_dir, 'start')
-  if #vim.api.nvim_list_uis() == 0 then
-    config.display.non_interactive = true
-  end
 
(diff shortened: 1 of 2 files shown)


jdrouhard:
ping @wbthomason

wbthomason (approved):
LGTM, thanks! Sorry for the delay in reviewing.

PR #1149, opened 2022-12-03
feat(compile): use nvim_create_user_command to make lazy-load commands

wbthomason:
This is a slight performance improvement over the old style of command creation, and moves `packer` toward the modern Neovim APIs

Not immediately merging because I need to track down the Neovim version when `nvim_create_user_command` was released - if it's v0.8.0, I'll want to hold off on this change since the release was quite recent.

I could add logic to test for the function's existence and fall back to the old method if it's not present, and might. But for now, I'd prefer to avoid adding that complexity.

### DIFF
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -571,7 +571,9 @@ local function make_loaders(_, plugins, output_lua, should_profile)
     local command_line
     if string.match(command, '^%w+$') then
       command_line = fmt(
-        'pcall(vim.cmd, [[command -nargs=* -range -bang -complete=file %s lua require("packer.load")({%s}, { cmd = "%s", l1 = <line1>, l2 = <line2>, bang = <q-bang>, args = <q-args>, mods = "<mods>" }, _G.packer_plugins)]])',
+        [[pcall(vim.api.nvim_create_user_command, '%s', function(cmdargs)
+          require('packer.load')({%s}, { cmd = '%s', l1 = cmdargs.line1, l2 = cmdargs.line2, bang = cmdargs.bang, args = cmdargs.args, mods = cmdargs.mods }, _G.packer_plugins)
+        end, packer_cmd_opts)]],
         command,
         table.concat(names, ', '),
         command
@@ -749,6 +751,7 @@ local function make_loaders(_, plugins, output_lua, should_profile)
   -- The command and keymap definitions
   if next(command_defs) then
     table.insert(result, '\n-- Command lazy-loads')
+    table.insert(result, "local packer_cmd_opts = {nargs='*', range=true, bang=true, complete='file'}")
     timed_chunk(command_defs, 'Defining lazy-load commands', result)
     table.insert(result, '')
   end
--- a/lua/packer/compile.lua
+++ b/lua/packer/compile.lua
@@ -570,10 +570,17 @@ local function make_loaders(_, plugins, output_lua, should_profile)
   for command, names in pairs(commands) do
     local command_line
     if string.match(command, '^%w+$') then
+      -- Better command completions here are due to @folke
       command_line = fmt(
         [[pcall(vim.api.nvim_create_user_command, '%s', function(cmdargs)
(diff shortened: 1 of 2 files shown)


lewis6991:
I think it was added in v0.6

folke:
@wbthomason you can get completion to work for lazy loaded commands as well. Check https://github.com/folke/lazy.nvim/blob/6affae64545da6efd2940589bff8253c08bc7252/lua/lazy/core/handler.lua#L169

wbthomason:
@folke Thanks, that's a neat trick!

wbthomason:
> I think it was added in v0.6: [neovim/neovim#16752](https://github.com/neovim/neovim/pull/16752)

Yeah, I think you're right. Thanks for tracking down the relevant PR; I was too lazy to try before sleeping yesterday.

kevinhwang91:
Error in packer_compiled: ...vim/site/pack/packer/opt/packer.nvim/lua/packer/load.lua:84: Vim:E477: No ! allowed
Please check your config for correctness

wbthomason:
@kevinhwang91 Did this happen on Neovim start, or when using a particular command?

wbthomason:
Pushed a temporary reversion in 1d8babf

wbthomason:
Did it happen when you just started neovim, though, or when you tried to use `Flog`? Or tab-complete `Flog`, maybe?

kevinhwang91:
```lua
        use {
            'rbong/vim-flog', branch = 'v2',
            cmd = {'Flog', 'Flogsplit'},
            requires = 'tpope/vim-fugitive',
            config = conf('plugs.flog')
        }
```
Type `:Flog` in the command will reproduce it.

kevinhwang91:
I found that it only throws E477 without calling `PackerCompile` after this PR. After calling `PackerCompile`, no errors anymore.

wbthomason:
Oh, I think I know what the problem is then (because I also cannot reproduce with a recompile). I'll add a patch to `load.lua`, test that I can't reproduce after that, and re-merge. Thanks!

wbthomason:
@kevinhwang91 Note also that `:Flog` does not support `bang`, so `:Flog!` *should* throw an error (per https://github.com/rbong/vim-flog/blob/master/plugin/flog.vim#L299)

This was a legitimate bug though, since `:Flog` alone also throws the error and should not.

kevinhwang91:
Yes, look like the issue is caused by the synchronization between `packer_compiled.lua` and `load.lua`

kevinhwang91:
It's a bit off-topic, looks like it's a core's bug.....

```lua
vim.api.nvim_create_user_command('SayHello', 'echo "Hello world!"', {bang = true})
vim.cmd('verb com SayHello')
vim.api.nvim_parse_cmd('SayHello', {})
```
image
I guess the `bang` in core is `false`, that's why `PackerCompile` can ignore E477.

wbthomason:
No, I believe that what happened was this:
- This PR changed the logic in `load` from providing a format string parameter as `cause.bang` to `cause.bang and '!' or ''` because `cause.bang` changed type from `string` to `boolean` with the switch to the new API
- If you had updated to this PR but not recompiled, `load` would still get passed the `string` version of `cause.bang` - this is the key to the bug
- Lua coerces the empty string `''` to `true` in the expression `'' and '!'`, so `'!'` was always returned
- This appended the `!` to commands that do not support it, like `Flog`, causing the E477 to be raised when those commands were used

That's why my patch fixes the issue by explicitly checking and handling the type of `cause.bang`.

kevinhwang91:
I know your explanation.

The compiled file after this PR:
```lua
pcall(vim.api.nvim_create_user_command, 'Flog', function(cmdargs)
    -- print(vim.inspect(cmdargs)) -- comment me

          require('packer.load')({'vim-flog'}, { cmd = 'Flog', l1 = cmdargs.line1, l2 = cmdargs.line2, bang = cmdargs.bang, args = cmdargs.args, mods = cmdargs.mods }, _G.packer_plugins)
        end,
        {nargs = '*', range = true, bang = true, complete = function()
          require('packer.load')({'vim-flog'}, { cmd = 'Flog' }, _G.packer_plugins)
          vim.api.nvim_input('<space><bs><tab>')
      end})
```
`bang = true` is passed to `nvim_create_user_command`, but you can comment `print(vim.inspect(cmdargs))` to inspect the result.

PR #1157, opened 2022-12-06
fix(compile): missing name and plugin_spec when loading config for plugins

younger-1:
In `PackerCompile`, packer will rerun config for these loaded plugins after source `packer_compiled.lua`.

But somehow it missed name and plugin_spec, which is not the case of other place using `loadstring` to run config.


wbthomason (approved):
Nice catch, thanks!

PR #1171, opened 2022-12-15
Pull upstream force-pushed Git tags

### LINKED ISSUE #1100: Sync fails when upstream force pushes Git tags
- `nvim --version`: v0.9.0-dev-1285-g760b399f6
- `git --version`: 2.37.0
- Operating system/version: macOS 12.6
- Terminal name/version: kitty 0.26.4
- 
### Problem description

Some projects update certain Git tags by force pushing them on a regular basis. Typical example: a `nighly` tag may change every day.

When this occurs anytime after packer has already been synced at least once, subsequent updates (sync) fail consistently with the message `[rejected] ... would clobber existing tag`. (see details below)

### Steps to reproduce

1. ```lua
    return require'packer'.startup(function(use)
      use 'kyazdani42/nvim-tree.lua'
    end)
    ```

1. ```vim
    :PackerSync
    ```

1. Force push any existing tag / wait until upstream does.

1. ```vim
    :PackerSync
    ```

### Actual behaviour

```
 ✗ Failed to update kyazdani42/nvim-tree.lua
  Errors:
    remote: Enumerating objects: 455, done.
...
    Resolving deltas: 100% (95/95), completed with 29 local objects.
    From https://github.com/kyazdani42/nvim-tree.lua
       875d38e..2b97063  master                   -> origin/*
     ! [rejected]        nightly                  -> *  (would clobber existing tag)
    875d38e
```

### Expected behaviour

```
 ✓ Updated kyazdani42/nvim-tree.lua
```

### Additional notes

Fetching from Git with the `--force` flag is the easiest solution that comes to mind: https://stackoverflow.com/q/58031165/4716370

My current workaround is to add this flag to [config.git.subcommands.update](https://github.com/wbthomason/packer.nvim/blob/6afb67460283f0e990d35d229fd38fdc04063e0a/doc/packer.txt#L218) while calling `init()`:

```lua
require'packer'.init{
  git = {
    subcommands = {
      update = 'pull --ff-only --progress --rebase=false --force'
    }
  }
}
```

### packer files

_Not relevant or already provided above._

overhacked:
Resolves #1100. Pull upstream tags even if they were force-pushed. Git version 2.20 changed the behavior of fetching tags so that non-fast-forward updates would fail. Some plugin authors, like `nvim-tree/nvim-tree.lua` use a tag named `nightly` that is force-pushed with latest updates to provide bleeding-edge changes.

By adding an explicit refspec to the `git fetch` command line with a leading plus-sign ('+') Git will always update the tag to the new commit, regardless of whether it was force-pushed upstream.

See https://www.git-scm.com/docs/git-fetch#Documentation/git-fetch.txt-ltrefspecgt for details (paragraph beginning "Until Git version 2.20...").

### DIFF
--- a/lua/packer/plugin_types/git.lua
+++ b/lua/packer/plugin_types/git.lua
@@ -380,8 +380,18 @@ git.setup = function(plugin)
         options = { env = git.job_env },
       }
 
+      -- If a tag is specified, name its refspec on the update
+      -- or fetch command line with a leading plus-sign ('+') to ensure that
+      -- upstream commits are pull/fetch-ed even if upstream force-pushed.
+      -- Not needed for branches, because in that case pulling force-pushed
+      -- commits is the default.
+      local explicit_refspec = ''
+      if plugin.tag then
+        explicit_refspec = fmt(' origin +refs/tags/%s', plugin.tag)
+      end
+
       if needs_checkout then
-        r:and_then(await, jobs.run(config.exec_cmd .. config.subcommands.fetch, update_opts))
+        r:and_then(await, jobs.run(fetch_cmd .. explicit_refspec, update_opts))
         r:and_then(await, handle_checkouts(plugin, install_to, disp, opts))
         local function merge_output(res)
           if res.output ~= nil then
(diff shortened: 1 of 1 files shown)


antoineco:
As mentioned in the linked issue, most people will run into the issue _without_ having an explicit tag set in their config.

For example, if upstream force pushes the `foobar` tag and users have a plain default Packer config (default branch), they'll run into the issue regardless of this PR, even if the `foobar` tag isn't relevant to them.

overhacked:
This is an over-complicated solution to a pretty simple problem. I posted my detailed comments on the issue: https://github.com/wbthomason/packer.nvim/issues/1100#issuecomment-1657036413.

Issue #886, opened 2022-04-24
Hererocks installation silently hangs if the default python command doesn't exist


babelfish:
- `nvim --version`: 0.7.0
- `git --version`: 2.25.1
- Operating system/version: Pengwin/Debian 10
- Terminal name/version: Windows Terminal 1.12.10983.0

### Steps to reproduce
In an installation that does not already have hererocks installed, attempt to install any rocks package without the `python` command available. I have both python 2 and 3 installed, but only available as `python2` and `python3`.

### Actual behaviour

Packer hangs indefinitely on "installing hererocks" with no log messages. I was only able to figure out what was going on by reading the packer source and manually trying to run the hererocks command it was trying to run.

### Expected behaviour

The installation should fail and provide an error message.

### packer files

<details>
<summary>Plugin specification file(s)</summary>

Simplified version with only the relevant info:

```lua
local packer = require('packer')
local use, use_rocks = packer.use, packer.use_rocks

packer.reset()
packer.init{
  log = { level = 'trace' },
  --luarocks = { python_cmd = 'python3' }
}


-- Lua packages
use_rocks 'luafilesystem'
```

</details>

<details>
<summary>packer log file</summary>

```
[DEBUG Sun 24 Apr 2022 07:04:22 PM EDT 4.4963620372526e+14] ...config/nvim/pack/packer/start/packer.nvim/lua/packer.lua:528: packer.sync: requiring modules
[DEBUG Sun 24 Apr 2022 07:04:22 PM EDT 4.4963620561182e+14] ...config/nvim/pack/packer/start/packer.nvim/lua/packer.lua:324: Processing plugin specs
[DEBUG Sun 24 Apr 2022 07:04:22 PM EDT 4.496362106793e+14] ...ack/packer/start/packer.nvim/lua/packer/plugin_utils.lua:176: Updating FS state
[DEBUG Sun 24 Apr 2022 07:04:22 PM EDT 4.4963637904433e+14] ...nvim/pack/packer/start/packer.nvim/lua/packer/update.lua:56: Fixing plugin types
[DEBUG Sun 24 Apr 2022 07:04:22 PM EDT 4.4963637911972e+14] ...nvim/pack/packer/start/packer.nvim/lua/packer/update.lua:67: Done fixing plugin types
[...]

PR #1187, opened 2023-01-08
Fix bug where hererocks installation hangs

hiimog:
Not all systems have a `python` executable in their path. In this situation, sync will never complete with the message: "installing hererocks...". This fix will log the situation and callback with an error code.

Issue #886

### DIFF
--- a/lua/packer/jobs.lua
+++ b/lua/packer/jobs.lua
@@ -50,7 +50,7 @@ end
 local spawn = a.wrap(function(cmd, options, callback)
   local handle = nil
   local timer = nil
-  handle = loop.spawn(cmd, options, function(exit_code, signal)
+  handle, pid = loop.spawn(cmd, options, function(exit_code, signal)
     handle:close()
     if timer ~= nil then
       timer:stop()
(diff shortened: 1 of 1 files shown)


wbthomason (commented):
Thanks for the PR and issue identification! I think the scope is actually broader than you've identified - this change will help protect against hangs due to missing `git` and `run` executables as well.

One question before merging: can you point me to docs that say `pid` is an error string if  `handle` is `nil`? The Luv docs seem to indicate that `pid` is always an integer, and if it might be an integer, then I'd like to add some logic to format the error message here.

wbthomason:
Also @hiimog please @ me; that helps me see/reply to things faster.

hiimog:
Happy to help, @wbthomason . [This](https://github.com/luvit/luv/blob/master/docs.md#error-handling) should provide some more clarity on the way errors are handled with luv. I suppose it is more accurate to say that `handle` and `pid` are `nil` and `error` in this case. This is also what I experienced when debugging: `nil, "ENOENT: no such file or directory"` as a result of [this](https://github.com/libuv/libuv/blob/v1.x/include/uv.h#L117).

wbthomason (approved):
LGTM, thanks!

Issue #1179, opened 2022-12-24
Autocompleting a lazy-loader dummy command calls the real command with args="nil"


mcwhitfield:
- `nvim --version`: v0.8.1
- `git --version`: 2.38.1
- Operating system/version: Arch rolling (6.0.11-arch1-1)
- Terminal name/version: gnome-terminal

### Steps to reproduce
Using Telescope as an arbitrary example:
```lua
use {
  'nvim-telescope/telescope.nvim',
  cmd = 'Telescope',
  config = function() require('telescope').setup() end,
  requires = 'nvim-lua/plenary.nvim',
}
```
Sync, restart, then
```lua
vim.fn.getcompletion('Telescope ', 'cmdline')  -- note space
-- or just type ":Telescope " with commandline autocomplete
```

### Actual behaviour
```
Error in packer_compiled: <...>packer/load.lua:91: Vim: [telescope.run_command]: Unknown command
Please check your config for correctness
```
is echoed, then valid completions for `:Telescope` are displayed

The completion function for dummy commands [leaves most elements of `cause` unset](https://github.com/wbthomason/packer.nvim/blob/master/lua/packer/compile.lua#L579), then `apply_cause_side_effects` sends `cause.args` to `fmt` [without checking](https://github.com/wbthomason/packer.nvim/blob/master/lua/packer/load.lua#L91) for `nil`. This results in `:Telescope nil` being invoked by the completion call (after the real command is loaded), resulting in the above error.

### Expected behaviour
Valid completions displayed with no error

### packer files

<details>
<summary>Plugin specification file(s)</summary>

```lua
return require('packer').startup(function(use)
    use 'wbthomason/packer.nvim'

    use {
        'nvim-telescope/telescope.nvim',
        cmd = "Telescope",
        config = function() require('telescope').setup() end,
        requires = 'nvim-lua/plenary.nvim',
    }
end)
```

</details>

<details>
<summary>packer log file</summary>

</details>

<details>
<summary>packer compiled file</summary>

```lua
-- Automatically generated packer.nvim plugin loader code

if vim.api.nvim_call_function('has', {'nvim-0.5'}) ~= 1 then
  vim.api.nvim_command('echohl WarningMsg | echom "Invalid Neovim version for packer.nvim! | echohl None"')
  return
end

vim.api.nvim_command('packadd packer.nvim')

local no_errors, error_msg = pcall(function()

_G._packer = _G._packer or {}
_G._packer.inside_compile = true

local time
[...]

Issue #1166, opened 2022-12-11
Recent update (on command-lazyloading) screws up vim-floaterm


Bekaboo:
- `nvim --version`: 0.8.1
- `git --version`: 2.38.1
- Operating system/version: Linux, kernel 6.0.11-arch1-1
- Terminal name/version: kitty 0.26.5

### Steps to reproduce

1. Use packer to install vim-floaterm with lazyloading:

```
use {
  'voldikss/vim-floaterm',
  cmd = {
    'FloatermNew',
    'FloatermToggle'
  },
}
```

2. Enter nvim
3. Type in commandline: `FloatermNew`
4. Hit space
5. Floaterm opens a new terminal with error right after we hit space

### Actual behaviour

See above.

### Expected behaviour

Floaterm should not open a new terminal before we hit <Enter>. The problem is command-lazyloading: if I do not lazy load Floaterm, or use key-lazyloading, the problem will not appear.

wbthomason:
I can't reproduce this. Is it still an issue? What `packer` commit are you on?

Bekaboo:
> I can't reproduce this. Is it still an issue? What `packer` commit are you on?

Hi! Thank you for your reply, I also find that with minimal config i.e. single `init.lua` that only calls `packer.startup` and registers `packer` and `vim-floaterm` does not have this issue. I'm trying to figuring out the real reason. I'll close this if the issue is related to packer. Sorry for the inconvenience.

Bekaboo:
It turns out `nvim-cmp` is the cause of the problem (don't know why). Close this issue.

Bekaboo:
Reopen this issue because I find starting from commit `ed2d5c9c17f4df2eeaca4878145fecc9669e0138` the problem arises. Before this commit packer use `file` for `complete` option; after this commit packer use a custom function to get the completion list. This brings a better completion experience but also introduces the bug.
@wbthomason

Bekaboo:
A short video demonstrating the bug:

https://user-images.githubusercontent.com/76579810/207523205-bd5db373-0769-4d8a-a33c-4f5016822f58.mp4

PR #1188, opened 2023-01-09
fix(compile): do not call real commands in lazyloading dummy commands…

Bekaboo:
… (#1179, #1166)

Current version of packer will make unexpected call to lazy-loaded commands when packer tries to get completion of that command.

This commit fixes the issue by passing empty table as `cause` to function `packer_load` to avoid calling the actual command in this case.


Bekaboo:
Bug demo:

https://user-images.githubusercontent.com/76579810/211249641-269cc714-74e1-48b4-a528-64e41451e7b2.mp4

After fixing:

https://user-images.githubusercontent.com/76579810/211249681-e838b061-411a-44c8-a5bb-20759c99a344.mp4

wbthomason (approved):
LGTM; thank you for catching and fixing this!

PR #1190, opened 2023-01-11
Fix: Change display headline text to not indicate that all tasks are plugins

### LINKED ISSUE #1086: Question

wbthomason:
Fixes #1086

@nat-418, @notwidow


nat-418:
Looks good to me.

PR #1193, opened 2023-01-12
Multiple modifications to improve snapshot functionality and user autocmds

Bekaboo:
These commits mainly aims to:

1. Improve the functionality of snapshots
2. Add more user autocmds
3. Echo informative messages when a `Packer*` command is finished

For the first point, now packer are capable to:
- automatically generate snapshot files on sync/update/remote/install
- allow using functions to specify the name of snapshot files and snapshot path
- generate human-readable `.json` with each package in a seperate line and is always sorted in ascending order, so that it is easier to track with version control tools like git
- `PackerSnapshot` can take any number of arguments (`nargs=*`). When no argument is given, packer will automatically fall back to the default snapshot name
- add `.json` appendix for the default snapshot name
- use `config.snapshot.name`, `config.snapshot.path`, and `config.snapshot.auto`, and `config.snapshot.silent_overwrite` to configure snapshot behavior

image
More human-redable

image
Clearer version control

The the second point:
- new user autocmds `PackerInstallComplete`, `PackerUpdateComplete`, `PackerCleanComplete`, `PackerSnapshotComplete`, `PackerSnapshotRollbackComplete`, `PackerSnapshotDeleteComplete`, and `PackerCompileComplete` are added

For the third point:
- in the function `packer.on_complete` a notification is added to inform user of the completion of a task (i.e. sync, install, update, snapshot ...)

Some side notes: thank you for this awesome plugin, I've used it for over a year and love it. For me it is the fastest package manager since I can disable packer itself and only load the compiled file if it is there. Taking advantage of this feature packer is noticeably faster than lazy.nvim (15 ~ 25ms for packer and 25~35ms for lazy.nvim, though both is fast enough). Use packer as a 'compiler' also means that the startup time will not increase when the config file is broken into smaller files, as long as the specs are the same. Hoping packer keep thriving and keep fast in the future :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,6 +284,7 @@ default configuration values (and structure of the configuration table) are:
 ```lua
 {
   ensure_dependencies   = true, -- Should packer install plugin dependencies?
+  auto_snapshot = false,  -- Automatically updates snapshot file when package changes
   snapshot = nil, -- Name of the snapshot you would like to load at startup
   snapshot_path = join_paths(stdpath 'cache', 'packer.nvim'), -- Default save directory for snapshots
   package_root   = util.join_paths(vim.fn.stdpath('data'), 'site', 'pack'),
--- a/README.md
+++ b/README.md
@@ -566,7 +566,14 @@ name and information table as arguments.
 require knowing when the operations are complete, you can use the following `User` autocmds (see
 `:help User` for more info on how to use):
 
-- `PackerComplete`: Fires after install, update, clean, and sync asynchronous operations finish.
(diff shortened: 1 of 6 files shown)


Bekaboo:
Sorry, forget to mention that commit [2090053](https://github.com/wbthomason/packer.nvim/pull/1193/commits/2090053b4d62be7a62a9c5a6d5aa59d477cc4161) has breaking change (remove `PackerComplete`). We can easily add it back if necessary, though.

PR #1200, opened 2023-01-22
add plugin overriding feature

prevostcorentin:
PR related to issue #1201.

### DIFF
--- a/tests/plugin_utils_spec.lua
+++ b/tests/plugin_utils_spec.lua
@@ -1,39 +1,42 @@
-local a = require('plenary.async_lib.tests')
+local a = require 'plenary.async_lib.tests'
 local await = require('packer.async').wait
 local async = require('packer.async').sync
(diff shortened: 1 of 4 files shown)


prevostcorentin:
It can be improved:
If two plug-ins defines `override=true`, the second one in the read order will replace the first order. Because the plug-in specification table is filled asynchronously, two overrides will bring hazardous situations.

Two solutions comes in mind:
* Explicitly **forbid** more than one overrides for a given plug-in
* Specifiy an override priority explicitly for a given plug-in

tests/plugin_utils_spec.lua
```diff
@@ -1,39 +1,42 @@
-local a = require('plenary.async_lib.tests')
```
prevostcorentin, line 1: Many undesirable changes are brought by an auto format feature.
It can be remove if needed.

